PR #6, opened 2019-11-18
Implements trust-region solver

TomRiem:


kellertuer:
I think we can merge this, since it fulfils the code coverage requirements. The difference in the return value is something we should maybe generalize anyways (which is the one thing left in code coverage).

PR #13, opened 2020-02-23
[WIP] Refactor to use ManifoldsBase and Manifolds

kellertuer:
A separate package for manifolds was introduced in 
https://github.com/JuliaNLSolvers/ManifoldsBase.jl
and
https://github.com/JuliaNLSolvers/Manifolds.jl
which this rework is now based on.

The only features that are not yet available are a
* Independent `random_point` and `random_tangent`
* a `typical_distance`

and the latter is not yet registered. This will only be merged, when these remaining points are finished and all tests are back up running again.


kellertuer:
i finsihes most functions, there is just

- [ ] a bug somewhere in Douglas Rachford that yields a non-convergent test
- [ ] the trust regions example uses a product manifold, but extracting things from product points and returning product points correctly is a little tricky still...

kellertuer:
All algorithms run, but the DR does not yet converge.
The remaining stuff is merely renaming functions to fit the new code style as well as adapt tests to work again with CI.

PR #29, opened 2020-05-27
Introduces Bezier curves and gradient&cost to minimise their acceleration

kellertuer:
This PR introduces the gradient and cost from [here](https://www.frontiersin.org/articles/10.3389/fams.2018.00059/full)
- [x] A tutorial how to use the Bezier curves and its accelerations gradient.
- [x] Test coverage checks (hence this PR)
- [x] maybe a speedup by using more in-place operations during the chain rule for the gradient.


src/functions/gradients.jl
```diff
@@ -1,3 +1,180 @@
...
+function ∇acceleration_bezier(
+    M::Manifold,
+    B::AbstractVector{P},
+    degrees::AbstractVector{<:Integer},
+    T::AbstractVector{<:AbstractFloat},
+) where {P}
+    gradB = _∇acceleration_bezier(M, B, degrees, T)
+    Bt = get_bezier_segments(M, B, degrees, :differentiable)
+    for k=1:length(Bt) # we interpolate so we do not move end points
+        gradB[k].pts[end] .= zero_tangent_vector(M, Bt[k].pts[end])
```
mateuszbaran, line 39: I think you could use `zero_tangent_vector!` instead of broadcasting.

src/functions/gradients.jl
```diff
@@ -1,3 +1,180 @@
...
+        # update p - effect from left and right segment as well as from c1 cond
+        X3 = ∇B[k].pts[1] .+ adjoint_differential_geodesic_endpoint(
+            M,
+            Bt[k-1].pts[m-1],
+            Bt[k].pts[1],
+            2.,
+            ∇B[k].pts[2]
+        )
+        # store
+        ∇B[k-1].pts[end-1] .= X1
```
mateuszbaran, line 170: You could probably just do `∇B[k-1].pts[end-1] .+= adjoint_differential_geodesic_startpoint(`?

src/functions/gradients.jl
kellertuer: Good idea, thanks!

src/functions/gradients.jl
kellertuer: no since the three terms are mixed, so I need the `∇B[k-1].pts[end-1]` in the three computations before (as well as others), i.e. for `X3`, so I think here I need the three allocations for X1, X2 & X3.

mateuszbaran: Right, it's not that easy to improve.

PR #30, opened 2020-06-22
Fix Get Started link

freemin7:


kellertuer:
Cool, thanks for noticing; I might have changed the URL style somewhen, so that's where this appeared.

PR #33, opened 2020-09-03
Type stabilization

mateuszbaran:
The current approach of being extremely dynamic during optimization comes at a very big cost -- optimization could be much faster. This PR aims to slightly reduce what can be changed during optimization to make the optimization type-stable in typical scenarios and thus significantly faster.

It's very WIP as I need your feedback about which parts need to stay dynamic and which do not. I would generally assume that `Dict`s can be relatively safely changed to `NamedTuple`s and most (all?) (mutable) structs can be changed to be concretely typed. For example, in `RecordOptions`, if you want to remove or add a recorded thing you could setup everything you want to record at the beginning and then enable or disable recording things instead of modifying `RecordOptions`.


kellertuer:
Thanks, these are already some nice improvements, especially  for the type stability; some of that code even stems from quite early times when I was working with Julia, so thanks for revising those.

I don't completely follow your RecordOptions explanations, currently there is a dictionary of fields to record from the options that are decorated.

mateuszbaran:
> I don't completely follow your RecordOptions explanations, currently there is a dictionary of fields to record from the options that are decorated.

Maybe I didn't explain it clearly. Do you ever modify that dictionary during a single optimization run?

kellertuer:
No, and concerning, that its storing an array every iteration, it should even not change.

Different maybe for the DebugOptions (which are similar but print something instead of record), for them I would be fine with a change.

mateuszbaran:
OK, that's good to know. You can disregard my comments about `RecordOptions` then.

kellertuer:
Can you explain what the advantage of `Callable` is instead of (just) `Function`?

mateuszbaran:
Actually using `Any` is even better. The thing is that any object can be callable and it's useful to allow users to pass anything callable, not just functions. This is useful when writing type-stable code because you can refer to fields in a `struct` instead of global state (or local variables) when implementing a function.

kellertuer:
Right, functors are actually quite nice; I just wasn'T aware of the callable-type.

mateuszbaran:
Can I change `StoreOptionsAction` https://github.com/JuliaManifolds/Manopt.jl/blob/c6327cc7d03883623e33e60398c198af3563803b/src/plans/options.jl#L174-L180 to make it have an immutable set of keys?

src/functions/proximalMaps.jl
```diff
@@ -267,12 +265,12 @@ function prox_TV2(M::PowerManifold{N,T}, λ, x, p::Int=1) where {N,T}
...
   for k in 1:d # for all directions
+    ek = CartesianIndex(ntuple(i  ->  (i==k) ? 1 : 0, d) ) #k th unit vector
     for l in 0:1
       for i in R # iterate over all pixel
         if (i[k] % 3) == l
-          I = [i.I...] # array of index
-          JForward = I .+ 1 .* (1:d .== k) #i + e_k
-          JBackward = I .+ 1 .* (1:d .== k) # i - e_k
+          JForward = i.I .+ ek.I #i + e_k
+          JBackward = i.I .- ek.I # i - e_k
```
mateuszbaran, line 273: I guess this was a bug?

mateuszbaran: Both `JForward` and `JBackward` had the same code (with addition)..

kellertuer: Oh, yes, thanks for noticing; I actually have not yet used this function in Julia (just transcribed it from my Matlab code) – so yes this is a copy paste error.

mateuszbaran:
OK, I'll probably stop at this place and continue in another PR. Right now the test example of `cyclic_proximal_point` is about 4x faster then before. For gradient descent it's about 9x faster (if you replace functions referencing global state with callable struct, impossible on Manopt master). Without that it's about 2x slower for some reason.

kellertuer:
Tghanks a lot for carefully checking and refactoring the code; I will review that soon. 

Can you show the replacement in an example? I am not exactely sure what you mean by „replace functions referencing global state with callable struct“?

I am not sure what codecov does here, but tests are running, that's good :)

mateuszbaran:
> Can you show the replacement in an example? I am not exactely sure what you mean by „replace functions referencing global state with callable struct“?

When you reference an external symbol in a function, for example `F` here:
```
    M = Circle()
    r = [-π/2,π/4,0.0,π/4];
    f = r;
    F(x) = 1/10*sum(distance.(Ref(M),f,Ref(x)).^2)
```
the function `F` holds references to `M` and `f`, so that when you change these variables, `F` uses the new values. This way the compiler often can't even assume what type `M` or `f` are. A simple solution is to make a struct (or a mutable struct) like this:
```
struct F2sT{TM, TF}
    M::TM
    f::TF
end
(F2::F2sT)(x) = 1/10*sum(distance.(Ref(F2.M),F2.f,Ref(x)).^2)
```
Now `M` and `f` are local to an object of type `F2sT` and their types can be known during compilation.

kellertuer:
Ah, that's good to know, especially since I like those functors a lot actually :)

kellertuer (approved):
Looks very good, thanks for optimising the code.

Issue #20, opened 2020-04-19
Fix Douglas Rachford with the new product manifold


kellertuer:
For more that 2 proves, i.e. for the parallel Douglas Rachford get/set_component seem to not work, as can be seen in the DouglasRachford SPD example https://github.com/JuliaManifolds/Manopt.jl/blob/master/examples/Total_Variation/SPDImage_DR_vs_CPPA.jl

PR #35, opened 2020-10-10
Fixes Douglas Rachford Tests

kellertuer:
* [x] single (2prox) tests
* [x] SPD example for more than 2 proxes (see #20)


kellertuer:
It seems running https://github.com/JuliaManifolds/Manopt.jl/blob/6cce00bfa5e998af029aaea1f211b8606a84b2a1/examples/Total_Variation/SPDImage_DR_vs_CPPA.jl combines the power of power manifold into a single index array, which then fails.

kellertuer:
For the SPD example to work, this PR will have to wait for JuliaManifolds/Manifolds.jl#235

PR #36, opened 2020-10-13
Barzilai-Borwein step size

r-dornig:
Adds a non-monotone line-search which uses the Barzilai-Borwein step size.


kellertuer:
This already looks quite nice!
Can you run formatter once and commit the changes?
The command simplified to 
````
julia  -e 'using JuliaFormatter; format(".")'
````
recently.

PR #51, opened 2021-01-20
Introduce modifying high level interfaces

kellertuer:
If you do not want to allocate new memory, this small PR introduces `!`-variants of the solvers that work in place.


mateuszbaran:
Nice! One thing that would be a nice addition to this are in-place proximal maps.

src/solvers/ChambollePock.jl
```diff
@@ -46,8 +46,51 @@ For more details on the algorithm, see[^BergmannHerzogSilvaLouzeiroTenbrinckVida
...
+    ξ::T,
+    m::P,
+    n::Q,
+    prox_F::Function,
+    prox_G_dual::Function,
+    forward_operator::Function,
+    adjoint_linear_operator::Function;
+    kwargs...,
+) where {P,T,Q}
+    x_res = allocate(x)
```
mateuszbaran, line 62: These allocated variables are never used later.

mateuszbaran:
Also, without in-place `step_solver!` there is little benefit in doing this.

mateuszbaran:
So for example this method: https://github.com/JuliaManifolds/Manopt.jl/blob/master/src/solvers/conjugate_gradient_descent.jl#L82-L88 should use in-place versions of `retract`, `get_gradient` and `vector_transport_to`. BTW, in-place generic accumulation like in `o.δ` is a bit problematic now (broadcasting is not extensible enough to make it nice).

src/solvers/ChambollePock.jl
kellertuer: Ah, thanks for noticing, maybe my concentration was already a little off after changing so many interfaces. I corrected this.

kellertuer:
Thanks for checking this PR.
* `step_solver!` (as the `!` indicates) already works in place of the `options` modifying especially it modifies `o.x`. So this is also a Bugfix basically, since the old functions „claimed“ to be non modifying, while they actually were modifying x0.
* Thanks for noticing to use `retract!`, the problem here, is that my tests on `Circle` fail if I do that. 
* where is there a `vector_transport_to`? Within the direction?

Concerning the `get_gradient` and the note on proximal maps; this would be quite a change to the interface, so that's why I did not yet do this. Basically, there is two problems here
1. For the circle or the positive numbers one has to find a way to still actually not have modifying functions for those, since thereon it is not possible; this also holds for the algorithms actually
2. for beginners it's harder to get started if they have to implement a `∇f!(y,x)` instead of just `∇f(x)` (similarly for proximal maps. And (similarly to 1) I am not sure how to easily support both.

Having that would be great. I am sure that would not be backwards compatible, but a Manopt 0.3 is anyways a good idea, since with the forthcoming quasi Newton (LBFGS especially) we have a major feature coming along. I am just not sure how to do both modifying variants and ease of use.

mateuszbaran:
> * `step_solver!` (as the `!` indicates) already works in place of the `options` modifying especially it modifies `o.x`. So this is also a Bugfix basically, since the old functions „claimed“ to be non modifying, while they actually were modifying x0.

Yes, it just doesn't use in-place functions.

> * Thanks for noticing to use `retract!`, the problem here, is that my tests on `Circle` fail if I do that.

That's a bit problematic. Essentially your circle test assumes support for immutable point representations and that requires either not using in-place variants at all or modifying options through references (something like `Ref(o, x)` instead of `o.x` but weirdly `Ref` can't do that; it should be easily doable with a custom reference type though).

> * where is there a `vector_transport_to`? Within the direction?

It's here: https://github.com/JuliaManifolds/Manopt.jl/blob/816cdc7dc275b49f3f4dccc733a94e5e7eb22493/src/solvers/conjugate_gradient_descent.jl#L88

> * For the circle or the positive numbers one has to find a way to still actually not have modifying functions for those, since thereon it is not possible; this also holds for the algorithms actually
> * for beginners it's harder to get started if they have to implement a `∇f!(y,x)` instead of just `∇f(x)` (similarly for proximal maps. And (similarly to 1) I am not sure how to easily support both.

I think the easiest solution would be having two variants of `step_solver!` for each solver: a mutating one and a nonmutating one. The current implementations would be exactly the nonmutating ones and wouldn't use in-place retract, etc. The mutating ones wouldn't work on `Circle`. I think it would make sense to make the new variants you add here use the mutating `step_solver!` (`step_solver!!` ?) methods. https://github.com/JuliaNLSolvers/NLSolvers.jl does the same thing with having separate in-place and out-of-place implementations.

kellertuer:
> Yes, it just doesn't use in-place functions.

Exactly. 

> 
> > * Thanks for noticing to use `retract!`, the problem here, is that my tests on `Circle` fail if I do that.
> 
> That's a bit problematic. Essentially your circle test assumes support for immutable point representations and that requires either not using in-place variants at all or modifying options through references (something like `Ref(o, x)` instead of `o.x` but weirdly `Ref` can't do that; it should be easily doable with a custom reference type though).

Yes, I am not yet sure what to do with that.
And thanks for pointing out the vector transport, I looked at gradient not conjugate gradient, sorry. That could maybe be made inlace, yes.

> I think the easiest solution would be having two variants of `step_solver!` for each solver: a mutating one and a nonmutating one. The current implementations would be exactly the nonmutating ones and wouldn't use in-place retract, etc. The mutating ones wouldn't work on `Circle`. I think it would make sense to make the new variants you add here use the mutating `step_solver!` (`step_solver!!` ?) methods. https://github.com/JuliaNLSolvers/NLSolvers.jl does the same thing with having separate in-place and out-of-place implementations.

I have to think about that, and yes it would most probably be a `step_solver!!`, since it anyways always modifies `o` in place , but the question then is whether or not to modify the fields of `o` in place.
My problem with that idea is, that I would like to avoid implementing things twice. For gradient, sure, that's easy, but for ChambollePock or QuasiNewton, that would require quite some work and reorganising (to check which parts can maybe be still used by both). And, also, a proper idea how to handle gradient evaluations (in place vs out of place) provided by the user.

I have to think about that.

kellertuer:
I thought about this for a while. I think I have two things I am not sure how to solve

1. we have to find a way to distinguish Problems having `∇F` and those having a `∇F!`. The first would reallocate per iteration, the second not. I currently fear this would mean to reimplement `step_sover!` twice (for each variant) but I hope that on more complex methods this can be done on “lower level” functions, since usually each part of an iteration is a seperate sub function, see for example https://github.com/JuliaManifolds/Manopt.jl/blob/816cdc7dc275b49f3f4dccc733a94e5e7eb22493/src/solvers/ChambollePock.jl#L118-L125
I think this could be done introducing new Problem types; for example introducing an `AbstractGradientProblem`and a `MutatingGradientProvlem` that would have `∇F!` instead of a `∇F`.
While it might mean reimplementing a few things and having a few duplicate implementations, doing it “as low as possible” in the levels might make the reimplementations as less as possible.

2. I am not yet sure what to do with the high Level interfaces then. Do we do a `gradient_descent` that goes “full allocating”, i.e. does not work on its input and have a `∇F`? Do we then do a `gradient_descent` that both works on the input `x` as well as with `∇F!`? Is there a reason to have a mix (for example mutating the input but having `∇F`)? What would we call that?

mateuszbaran:
Ad. 1. Yes, it would certainly make sense to information about in-place/out-of-place in the problem, but maybe adding a type parameter to `Problem` would be more suitable? You would have a simpler type hierarchy and this option is applicable to all types of problems.

Ad 2. I think there is no point in doing a mixed approach. Mutating `gradient_descent` can mutate both the argument and intermediate values.

kellertuer:
A type parameter sounds really nice, yes. That's what we do. Thanks for the discussion and the tipp.

One could do a mixed variant that uses non-mutating gradient functions but works mutatingly on the option fields, but maybe with above parameters one could introduce this if there is a need for this.

I think I would still like to finish this small PR (it mainly makes the interfaces more rigorous about whether they are mutating the index or not)? I will start the rework to what we discussed then next week maybe.

mateuszbaran:
> One could do a mixed variant that uses non-mutating gradient functions but works mutatingly on the option fields, but maybe with above parameters one could introduce this if there is a need for this.

Yes, of course, there could always be more than two values for that parameter.

> I think I would still like to finish this small PR (it mainly makes the interfaces more rigorous about whether they are mutating the index or not)? I will start the rework to what we discussed then next week maybe.

Sure, that's a good plan.

kellertuer:
Cool. And many thanks for this discussion and its nice plan to a solution. I had for some time in mind to find a way to introduce mutating gradient functions. Now I know how.

Issue #52, opened 2021-01-22
Introduce mutating gradient function containing problems.


kellertuer:
as discussed in #51.

PR #53, opened 2021-01-24
Introduce mutating functions for Problem member functions

kellertuer:
This is a major rework of the inner structure of `Manopt.jl` to support mutating problem functions, most prominently `gradient!(X,x)`.

* [x] introduce a parameter for the problem to indicate whether its member functions are mutating or not.
* [x] introduce a unified `gradient!!` function and encapsulate all of this in `get_gradient`/`get_gradient!`
* [x] introduce fallbacks for `Circle`/`PositiveNumbers` since then `gradient!!` never mutates.
* [x] check where `get_gradient!` should be introduced
* [x] refactor `hessian` to `hessian!!`
* [x] update `ApproximateHessian` to act as a normal hessian (mutating or allocating) function
* [x] refactor `subgradient` to `subgradient!!`
* [x] refactor `proxes` to `proximal_maps!!`
* [x] refactor `∇` to `stochastic_gradient!!`
* [x] refactor all 4 functions from `PrimalDualProbem` to `!!`.
* [x] update gradient and prox function library to provide efficient mutating and non mutating variants. Check that all grad/prox/cost hast `M` as their first parameter.
    * [x] adjoint differentials
    * [x] bezier curves
    * [x] costs
    * [x] differentials
    * [x] Jacobi fields
    * [x] proximal maps
* [x]  improve test coverage
* [x] carefully only load random functions and non-mutating-specials if manifolds is loaded.
* [x] check/update documentation
* [x] check tutorials and examples.

This then solves #52.
  
The problem parameter might even later be used to indicate whether the gradient/hessian is Riemannian or Euclidean, though their conversions have to first be more established.

Maybe a small benchmark of the two new methods (providing either a allocation or a mutating gradient) would be nice, too.


kellertuer:
This PR is finished, we just need JuliaManifolds/Manifolds.jl#334 to be merged.

PR #55, opened 2021-02-05
Fix latex typo in JacobiFields tutorial

fkastner:
I just deleted a spurious dollar sign which prevented the equation from rendering.


kellertuer:
Cool, thanks for noticing! While you're at it, can you fix the two lines above, too? There's two formulae in `\(...\)` which should ne in `$...$`.

fkastner:
The push preview doesn't work for PRs submitted from a fork but I think it should be ok now.

kellertuer:
Yes, that's a security feature, I think – thanks for the quick fix, I fear the `~` might be rendered still, could you also remove those? And thanks again for seeing this typo :)

fkastner:
I thought they were latex style non-breaking spaces but sure, I can remove them.

kellertuer:
Cool! They _are_ indeed LaTeX nonbreaking spaces, _but_ at the point where they are it's markdown.
Just for safety I'll let the test run through before I merge. Thanks for the fast fix.

PR #62, opened 2021-03-19
loosen type constraints on gradient to be different from point type.

### LINKED ISSUE #61: FixedRankMatrices representation of manifold points and tangent vectors
Hello,

Thanks for this great Toolbox!
I have been using PyManopt so far but I wanted to give a try to Julia.
The API is quite different so it raised some questions for me.

I am trying to optimize over the manifold of fixed rank matrices with a simple gradient descent algorithm (and trust regions later). From what I understand in the gradient plan `GradientDescentOptions` the point on the manifold and the gradient need to have the same type (annotated with `::P`). But for fixed rank matrices the points are represented with the class `SVDMPoint` and the tangent vectors with the class `UMVTVector`.
So I end up getting errors when I try to run `gradient_descent` with a simple cost function and I don't really get what I am doing wrong here.
I am pretty new with Julia so I might be missing something obvious.

More generally, I was wondering if you had examples with other manifolds than the Sphere, and in particular with the FixedRankMatrices manifold?

Two unrelated questions (I can open other issues if needed):
- Is there some automatic differentiation available? Is it possible to simply feed the Euclidean gradient and Hessian and get them converted?
- The recording tool that you have looks extremely useful. I was wondering if it's possible to record quantities like the number of evaluations of the cost / gradient / Hessian?

Thank you very much!

kellertuer:
resolves #61.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ examples/**/*.jld2
 docs/src/tutorials/*.md
 .vscode
 Manifest.toml
+*.DS_Store
--- a/src/plans/gradient_plan.jl
+++ b/src/plans/gradient_plan.jl
@@ -110,30 +110,37 @@ a default value is given in brackets if a parameter can be left out in initializ
 # Constructor
 
     GradientDescentOptions(x, stop, s [, retr=ExponentialRetraction()])
+    GradientDescentOptions(M, x, stop, s [, retr=ExponentialRetraction()])
+    GradientDescentOptions(x, X, stop, s [, retr=ExponentialRetraction()])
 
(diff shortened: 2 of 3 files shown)


mateuszbaran (approved):
Maybe it would be a good idea to check all solvers whether they work for fixed rank matrices or not?

kellertuer:
In general yes, the one problem we should solve before is the gradient and Hessian conversion functions. For now, I do not want to put them in every example bz hand.

And also, one could include a few examples maybe also in the tests, but some are also running too long. I have to think about that.

PR #66, opened 2021-03-20
Sample random points and tangent vectors from FixedRankMatrices manifold

qrebjock:
Adding simple functions to sample from the FixedRankMatrices manifold.


qrebjock:
Manopt doesn't have Manifolds in its dependency. But I think we need to do `import Manifolds: SVDMPoint, UMVTVector` at the beginning of `random.jl` in order to have the types for `random_point` and `random_tangent`.
Or is there another way to do it?

kellertuer:
Exactly, `Mantopt` only loads the `random`s if `Manifold` is already loaded. The idea is, that `Manopt.jl` stays lighter in its dependencies by only depending on `ManifoldsBase`.

In line 105 following pf `Manopt.jl` you see that the random parts are loaded as soon as Manifolds are available. You can add the elements you need there.

kellertuer:
It would be nice if you could also add test cases, too. Maybe like generate a random point/tangent and verify that it is a point/tangent, see `test/functions/test_manifold.jl` to see how it is done for the other random functions (maybe I should move that stuff to a `test_random.jl` ;) ).

kellertuer (approved):
All tests pass, all new lines are covered, so this looks good to me. Thanks for contributing those and writing the tests.

* Would you like to finish by raising the version number in the main `Project.toml` to `0.3.2`? (0.3.1 was released today).
* If you want you can add yourself to the `zenodo.json`, where currently I put contributors into two categories: project member for people that contributed code and other for people that reported typos. Just make sure to not break the json file, since then zenodo would not generate a doi ;)

qrebjock:
Done :)

Thanks a lot for the help.
I understand the project structure much better now.

kellertuer:
Well, thank you for your contribution! Looking forward to seeing more from your work :)

PR #65, opened 2021-03-20
use-case, trying out SE(2) with ProductRepr

dehann:
Hi this is an attempt to help.  Think I'm coming up against infrequent dispatches needed for Group Manifolds perhaps.

Following from #64 , this is an attempt to structure example, tests, and best usage.  Note that this test file does not run due to multiple errors.  Please move or delete if this is not the desired location or otherwise.  Note the `ProductRepr` approach follows earlier discussion at https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/issues/41#issuecomment-733636425

This test tries to run Manopt with either `NelderMead` or `gradient_descent` on a `SpecialEuclidean(2)` example.

Current known issues include the `broadcast` used inside NelderMead, or possible immutability when using gradient descent on StaticArrays.  Third, when dropping StaticArrays for conventional Arrays leads to yet more errors. 

cc @mateuszbaran , @kellertuer 

Please say how best I can help?

PR #68, opened 2021-03-22
Create test_NelderMead_Sphere2_consistency.jl

dehann:
Prior discussion at #65 

Hi @kellertuer ,

I noticed that NelderMead produces changing results for the `Sphere(2)` example from the docs.  Thought I'd PR this short test file as MWE (currently failing and not in runtests.jl yet).  Perhaps I'm misunderstanding something about how to use NelderMead in this case, either way hopefully this helps document or debug.

Best,
Dehann

EDIT, PS, although Nelder-Mead mean optimization is contradictory, I think the result should at least be consistent.  So perhaps good as a test, even though computations are being wasted?  Ref: https://github.com/JuliaManifolds/Manopt.jl/pull/65#issuecomment-803257250

### DIFF
--- a/test/solvers/test_NelderMead_Sphere2_consistency.jl
+++ b/test/solvers/test_NelderMead_Sphere2_consistency.jl
@@ -1,0 +1,54 @@
+# test NelderMead consistency
+# see lead-up discussions at Manopt #65
+
+using Test
+using Manopt, Manifolds
+using Random
+
+
+##
+
+@testset "check NelderMead consistency on Sphere(2)" begin
+
+## build problem (from Manopt docs)
+n = 100
+σ = π / 8
+M = Sphere(2)
+dim = manifold_dimension(M)
+x = 1 / sqrt(2) * [1.0, 0.0, 1.0]
+Random.seed!(42)
+data = [exp(M, x, random_tangent(M, x, Val(:Gaussian), σ)) for i in 1:n]
+
+F(M, y) = sum(1 / (2 * n) * distance.(Ref(M), Ref(y), data) .^ 2)
+
+
+## solve with defined initialization
+xMean1 = NelderMead(M, F, data[1:(dim+1)])
+xMean2 = NelderMead(M, F, data[(dim+2):(2*dim+1)])
+
+@test isapprox(xMean1, xMean2, atol = 1e-2)
+
+## solve with random initialization
+xMean3 = NelderMead(M, F, data[1:(dim+1)])
+xMean4 = NelderMead(M, F, data[1:(dim+1)])
+xMean5 = NelderMead(M, F, data[1:(dim+1)])
+
+@test isapprox(xMean1, xMean3, atol = 1e-2)
+@test isapprox(xMean1, xMean4, atol = 1e-2)
+@test isapprox(xMean1, xMean5, atol = 1e-2)
+
+@test isapprox(xMean2, xMean3, atol = 1e-2)
+@test isapprox(xMean2, xMean4, atol = 1e-2)
+@test isapprox(xMean2, xMean5, atol = 1e-2)
+
+@test isapprox(xMean3, xMean4, atol = 1e-2)
+@test isapprox(xMean3, xMean5, atol = 1e-2)
+
+@test isapprox(xMean4, xMean5, atol = 1e-2)
+
+##
+
+end
+
+
+#
--- a/test/solvers/test_NelderMead_Sphere2_consistency.jl
+++ b/test/solvers/test_NelderMead_Sphere2_consistency.jl
@@ -29,9 +29,9 @@ xMean2 = NelderMead(M, F, data[(dim+2):(2*dim+1)])
 @test isapprox(xMean1, xMean2, atol = 1e-2)
 
 ## solve with random initialization
-xMean3 = NelderMead(M, F, data[1:(dim+1)])
-xMean4 = NelderMead(M, F, data[1:(dim+1)])
-xMean5 = NelderMead(M, F, data[1:(dim+1)])
+xMean3 = NelderMead(M, F)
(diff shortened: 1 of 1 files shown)


kellertuer:
Note that `NelderMead` always uses an initial population to work with, which is initialised by random points, see https://github.com/JuliaManifolds/Manopt.jl/blob/c632de3ac451081db8840570491abc2c5e98b8ce/src/solvers/NelderMead.jl#L42 , which in your case are 3 random points. Sure with different random points the result is different. If you start with the same population, (or the same seed before calling `NelderMead`, the result is of course consistent.

dehann:
I think there might be a problem with the implementation, since the data used for the cost function is only created once.  I'd expect solving multiple times on exactly the same data/cost-function should produce the same results (even if different starting locations/populations are used)?

Cost function is created once here:
https://github.com/JuliaManifolds/Manopt.jl/pull/68/files#diff-b700790a86d914b61de70ad8c373c7f24a132243ad0dff46578f42556cdd03c4R20

Two prescribed starting populations (cost function is unchanged):
https://github.com/JuliaManifolds/Manopt.jl/pull/68/files#diff-b700790a86d914b61de70ad8c373c7f24a132243ad0dff46578f42556cdd03c4R26-R27

As well as the random initial population here (cost function is unchanged):
https://github.com/JuliaManifolds/Manopt.jl/pull/68/files#diff-b700790a86d914b61de70ad8c373c7f24a132243ad0dff46578f42556cdd03c4R32-R34

But in both cases the cost function (and its internal random data) should be exactly the same...  Perhaps I am again missing something.

> Do you think the docs should be changed?

If this comment above is right then I don't think the docs need to be updated.  Separately, I did send a small spelling fix PR and think I understand the current docs well enough (hope so).

dehann:
Ah, only reading other remark now:
https://github.com/JuliaManifolds/Manopt.jl/pull/65#issuecomment-804041255

kellertuer:
The problem is not the data that you create. That is unique and fine the way it is (that is the data you compute the mean on). Sure.

But Nelder Mead uses a population (on S2 3 points) to “move around and look for a minimum”, see the `population` parameter. If you use the same start population (remember to use a copy since the population is changed during the iterations), then you get a consistent result. The population is not your data.
Keep also in mind that the Riemannian Center of mass is not necessarily unique, ie. the cost is especially not globally convex (globally convex functions on the sphere are only the constant functions), so there are local minimisers and depending on your initial population you will end up near different local minimisers.

dehann:
What I didn't check was whether these different final solutions are somehow equivalent.  Maybe I incorrectly assumed that a random `Sphere(2)` population only has one globally unique mean (the cost function I used in this example).  As in exactly this example:
https://manoptjl.org/stable/tutorials/MeanAndMedian.html

That said, the mean of (x,y,z) points on a random `Sphere(2)` population should be just one point though?  So is the problem in the case of this PR then that there is no (NelderMead) workable path to the global minimum from random starting locations, but rather many "local minima" along the way?

kellertuer:
Well despite NelderMead being quite inefficient for this case (i.e. a long computation/iteration might be needed to get close), let*s look at one example of options (x,y,z):

Take 3 points on the equatior equally distributed round the circle (i.e. 120 degrees apart from each other). You can easily compute that all 3 points, x, y, and z are local minimisers for the mean of these three points (actually this holds for any odd number of points equally spaced on the equator).

And sure if you points happen to be local (say all meeting around a North Pole camp fire) that it there is a unique global minimiser near the North Pole, still there are a few local minimisers around the globe, still.

dehann:
Correction, you are right sorry there is not a globally unique minimizer on Sphere(2), there are at least two.  I said that too broadly.  However, I did expect this example to have a broader "region of attraction" since the points are quite concentrated around the true mean point.  I thought NelderMead would be much less likely to find the opposite pole as a solution in this case:
https://manoptjl.org/stable/assets/images/tutorials/startDataAndCenter.png

I fully agree with points around the equator example.

Here are 5 solutions I got when running this test file (notice how these seem to be 5 different solutions, so the same minima are never found):
```julia
julia> hcat(xMean1, xMean2, xMean3, xMean4, xMean5)
3×5 Matrix{Float64}:
  0.850536   0.78079    0.732007   0.46042       0.502207
 -0.0500822  0.408801   0.528803  -0.000597486  -0.682392
  0.523526   0.472491  -0.429573   0.887701      0.531159
```

Their cost values are:
```julia
julia> F(M, xMean1)
0.16400002590282048
julia> F(M, xMean2)
0.2038102464219593
julia> F(M, xMean3)
0.8130026522335593
julia> F(M, xMean4)
0.24542194638355772
julia> F(M, xMean5)
0.502675574646437
```

If all is good then we can close this PR too.  I'm just a bit surprised that the answers are so different for a seemingly simple case.  

Thanks for the feedback!

kellertuer:
I can confirm your results and think, you can not do much better with `NelderMead` actually. Keep In mind that NelderMead never uses even first order (i.e. gradient) information and mainly just compares function values of `F` to decide which point from the population to move into which direction, so at first glance I think all your results look fine. Not that not only your data has to be quite local (which it is, for sure), but also your population (which it isn't for the means 3-5).

I did the following short check by computing 
```
Julia> meanC = mean(M,data)
```
this might also not be that precise, since I think we are on the “maybe a little faster than more precise” area of parameters...
Then
```
julia> F(M, meanC)
0.14903020156241287
```
which is not that far away at least from your `xMean1` in cost (but it is in distance a little bit away).

So I think this is all correct, mainly due to NelderMead not being that fast converging (I am not sure it is converging to a minimiser theoretically at all) and that efficient, especially for a problem where you are low-dimensional and have a gradient at hand anyways. If it's ok with you, let's close this PR.

PR #71, opened 2021-03-24
Fix trust region

qrebjock:
Hello,

I propose these changes to fix a few bugs in the trust region algorithm and to improve efficiency.

Bugs that it should fix:
- The evaluation for the `StopWhenAny` stopping condition is lazy. So if the first condition is satisfied then the second one won't be evaluated and checking if it was satisfied with the `.reason` field won't work. This causes trust region not to update the trust radius correctly if another stopping condition is satisfied for the tCG sub-routine.
- After the tCG algorithm is done the field `o.Heta` is computed before updating `o.eta`. This causes the Hessian to be incorrect.
- In the stopping conditions of tCG there are two conditions on the residual norm. However, the initial residual norm should change after each iteration of the trust region algorithm. But this is not the case in the current implementation: it remains to its very first value because of the way the stopping conditions are built.
- The stopping conditions `StopIfResidualIsReducedByPower` and `StopIfResidualIsReducedByFactor` should be both satisfied to exit tCG. In the current implementation it is stopping when any of them is satisfied.
- The term `beta` in the tCG algorithm should actually be the inverse of the expression in the current code.

These bugs prevent the algorithm from converging quickly. With these changes it is more or less equivalent to the matlab version (I made a few tests to see if the outputs are the same). It's much faster and can reach super-linear convergence.

I tried to reduce the number of calls to the gradient and Hessian too:
- In the trust region algorithm we can reuse the gradient and the Hessian from the tCG algorithm.
- In tCG we need only 1 gradient call at initialization and 1 Hessian call per inner loop

However I am not sure exactly when memory is allocated. I think things can be improved further. The matlab version is a bit faster. That's probably because of too many memory allocations but I couldn't identify all of them.


kellertuer:
Oh, thanks! This algorithm was a Bachelor thesis recently, and I when handed in, the algorithm was _much_ slower, but we improved it afterwards, sorry that we might have missed a few points.

In the end, it should be allocations that make us only as fast as Matlab, but it might take some time until I have time to look deeper into that.

At first glance, you are now missing two lines of code coverage. For the rest – I need a few days maybe to check your changes.

mateuszbaran:
I will look into performance and memory allocations (but it may take a day or two). Could you share the example you are benchmarking? Many performance problems don't show up for objective functions that take a long time to compute, or there may be some problems in the objective function itself.

qrebjock:
That would be great!

I am experimenting with the following code:
```julia
using LinearAlgebra
using Manopt, Manifolds
using BenchmarkTools


m = 300
n = 300
k = 30

M = FixedRankMatrices(m, n, k)
A = randn(m, n)

function F(M, p)
    return norm(p.S)^2 / 2 - dot(p.S, diag(p.U' * A * p.Vt'))
end

function gradF(M, p::SVDMPoint)
    dS = diagm(p.S)
    return UMVTVector(
        -(I - p.U * p.U') * A * p.Vt',
        dS - p.U' * A * p.Vt',
        -p.U' * A * (I - p.Vt' * p.Vt)
    )
end

function hessF(M, X, H)
    dSinv = diagm(1 ./ X.S)
    return UMVTVector(
        (I - X.U * X.U') * (X.U * H.M + H.U - A * H.Vt' * dSinv),
        H.M,
        (H.M * X.Vt + H.Vt - dSinv * H.U' * A) * (I - X.Vt' * X.Vt)
    )
end

X = random_point(M)

@btime trust_regions(
    M,
    F,
    gradF,
    hessF,
    X,
    retraction_method=PolarRetraction(),
    stopping_criterion=StopWhenAny(StopAfterIteration(50), StopWhenGradientNormLess(1e-6)),
    # debug=[:Iteration, " | ", :Cost, "\n", :Stop],
);
```
This requires the retraction defined in this PR: https://github.com/JuliaManifolds/Manifolds.jl/pull/340.

kellertuer:
Thanks for the example. I will also take a look for the mutating variants later, i.e. if you define a `gradF!` and a `HessF!` and use the `evaluation=MutatingEvaluation()` option – that should also be checked to work fast/fine.

kellertuer:
Just a first short remark, I introduced `|` for stopping criteria, so you can also just write
```
stopping_criterion= StopAfterIteration(50) | StopWhenGradientNormLess(1e-6)
```
(similarly if you want a `StopWhenAll`, use `&`).

And if you change code make sure to run
```
using JuliaFormatter; format(".")
```
to unify the code style (indent, spacing etc).

qrebjock:
Ah good to know, the `|` operator is very convenient.

Something unrelated: I noticed that `StopAfterIteration(0)` performs one iteration. And `StopAfterIteration(1)` performs 2. And more generally `StopAfterIteration(n)` performs `n + 1`. Is this intended?

Yes I will format the code. I always forget, I should make it automatic when I save files.

kellertuer:
> Something unrelated: I noticed that `StopAfterIteration(0)` performs one iteration. And `StopAfterIteration(1)` performs 2. And more generally `StopAfterIteration(n)` performs `n + 1`. Is this intended?

Maybe an off-by-one error – I never really used the iteration stopping criterion until now. So no – that is not intended.

> Yes I will format the code. I always forget, I should make it automatic when I save files.

If we would not squash PRs when merging, you would notice quite a few “runs formatter” commits, so you are not the only one, we all do. Still, I like having that check to have a consistent code style. Until now I did not yet looking into automatically doing that locally on my machine.

kellertuer:
Thanks for noticing the changes and reducing the gradient evaluations, I think so far all you did is correct, I am just not yer sure, where we are currently using code coverage; but well.

mateuszbaran:
I've investigated that example a bit. The most important points are:
1. You should interpolate at least some of the arguments in benchmarking:
```julia
@btime trust_regions(
           $M,
           $F,
           $gradForig,
           $hessF,
           $X,
           retraction_method=PolarRetraction(),
           stopping_criterion=StopWhenAny(StopAfterIteration(50), StopWhenGradientNormLess(1e-6)),
           # debug=[:Iteration, " | ", :Cost, "\n", :Stop],
       );
```
2. I've managed to speed gradient calculation about 3x:
```julia
function gradFnew(M, p::SVDMPoint)
    dS = Diagonal(p.S)
    AVtp = A * p.Vt'
    pUpAVtp = p.U' * AVtp
    return UMVTVector(
        - AVtp + p.U * pUpAVtp,
        dS - pUpAVtp,
        -p.U' * (A - AVtp * p.Vt)
    )
end
```
3. It looks like time in this problem is mostly spent on doing linear algebra so micro-optimizations won't help much. You should re-use computed values of gradients and Hessians where possible. This solver is quite convoluted so I'm not sure if there are any remaining places where it could be done.
4. Manopt (the Matlab one) likely uses MKL instead of OpenBLAS used in Julia by default. On Intel processors MKL can be significantly faster so at least part of the difference may come from this.
5. Problem size is on the edge of what could be beneficial to speed up on GPU. If `m`, `n` or `k` were at least 4-5 times bigger, you may consider using CUDA.jl and `CuArray`.

TomRiem (approved):
Thank you for fixing and revising the code.

kellertuer:
Hi @qrebjock, could you just run once ``using JuliaFormatter; format(".")` on this branch? There seems to be a few lines that are not properly formatted. The code coverage is ok, then I would merge this and release a new minor version.

qrebjock:
Hello,

Sorry again for the late answer.
Thanks all for your comments!

Thank you for your benchmark and advice mateuszbaran, it's very instructive. Using MKL and interpolating arguments improve performance.
It doesn't match the MATLAB version yet, but I think that using mutating variants could help here too. I will try it.

I formatted the code so it should be good. I also changed the initial iteration counter to 1 in `solver.jl`. So that `StopAfterIteration(n)` stops after `n` iterations and not `n + 1`.

I was also working on numerical stability issues for trust regions. I observed for the fixed-rank matrices manifold some instabilities that are not in the MATLAB version. That's because a few additional steps are performed. In tCG at some point a tangent vector is re-projected on the tangent space, just to be sure that it's on it. This fix part of the issues. I am still experimenting to see where the rest comes from.

Having this re-projection step would require the manifold to have a function that takes a point in the ambient space and projects it on a tangent space. I guess that not all manifolds would have it, so it's maybe better to make this step optional.

kellertuer:
Thanks for carefully checking and further testing!

Yes, not all manifolds have project functions (for points/tangents), so it would be best to have an option for that to activate/deactivate that.
to be precise we only expect `AbstractEmbeddedManifolds` (where you know the embedding implicitly like for the sphere) and `EmbeddedManifolds` (where you couple two manifolds explicitly, this can for example also be used to have _different_ embeddings for a manifold) to have project (and embed) functions.
Feel free to propose an option for this.

Thanks for also checking the stopping criterion, the good old off-by-one errors...

kellertuer:
I checked the error in the test more closely and noticed that you fixed the bug, but missed one point (that also might not be that well documented): A lot of things are initialised at iteration 0 (i.e. before the first iteration).
Before an iteration starts, the `iter` is increased, that is, during _and_ after that iteration the iteration counter is at the value of the current iteration and only at the start of the iteration the counter is increased to said iteration.

With that the proxTV2 (which is computed by a sub gradient descent) was doing something wrong. So I propose to use https://github.com/JuliaManifolds/Manopt.jl/tree/kellertuer/fixPR71 this branch (merge it into here) to fix this issue. I started from your last commit here and made the tests work again.

What I did was
* set `iter` back to `0` in the beginning such that during and after the first iteration the `iter=1` holds but more importantly _before_ that iteration it is zero
* since we want to stop _after_ the `k` th iteration, the proper fix was to change the check in the stopping criterion from ` i > c.maxIter` to ` i >= c.maxIter` (stopping_criterion.jl line 42). Now it should be consistent, since after the `maxiter`th iteration the value is still `maxiter` and hence the criterion fires _after_ said iteration.

Since a few files were involved not yet touched here, I felt like it was easiest to to the new branch `fixPR71` that you can just merge here into (checked tests and format already).

Thanks for noticing, sometimes these off-by-one errors slip through.

mateuszbaran:
One more tip: most of the time here is actually spent doing SVD factorization in the retraction, so maybe using something like a truncated SVD from https://github.com/JuliaMatrices/LowRankApprox.jl could help?

kellertuer:
...and after checking my branch and the last note, please also bump the version to `0.3.8` (unless before you do that I find other things to update/publish).

qrebjock:
Done! Thanks for the help.

@mateuszbaran Yes I agree that it could help when the matrices are very big. That's definitely something to try.

Numerical issues are pretty different from the purpose of this PR so I guess it would be easier to open another one. Especially as I am still not sure where they all come from and it would be nice to merge these fixes.

kellertuer:
Thanks for the update – I also think the performance issue is something for a next issue, but thanks for bringing that up, @mateuszbaran !

I just fixed two things that were still off compared to master (the version had to be resolved and that we now have a proper `copyto!`). Let's see whether tests now run fine :)

kellertuer:
Looks great! Thanks for fixing this!

PR #73, opened 2021-03-25
Switch from deepcopy to allocate and copyto.

### LINKED ISSUE #72: Allocation of temporaries
When allocating temporaries, like here: https://github.com/JuliaManifolds/Manopt.jl/blob/45f587c988bad525ca23d19110f9af38baf240ae/src/solvers/conjugate_gradient_descent.jl#L57 , `deepcopy` is not a good solution because for static arrays it will return a new static arrays, while if you want to override that temporary in-place (for example using mutating functions like `retract!`) you need a mutable value. To solve this you need to call `allocate` instead and then `copyto!` if you need the temporary to be initialized.

kellertuer:
should resolve #72.

Currently the remaining problem is, that the new variant used here does not yet recursively copy and hence does not yet work as meant to.


kellertuer:
So here is a first approach using `recursive_copyto!`

Though with `ProductRepr` (see `manifold_functions.jl`) it seems to not work yet, or did I miss something on Static Arrays even? I don't see it directly.

Also currently JuliaFormatter seems to be broken on interpolation.

kellertuer:
Ha. Found the bug and JuliaFormatter should work again, too.

src/helpers/initialize_data.jl
```diff
@@ -0,0 +1,13 @@
...
+    recursive_copyto!(a,b)
+
+copy the values from `b` to `a` by recursively copying the contents.
+On the lowest level of `AbstractArray`s this falls back to `copyto!`.
+"""
+function recursive_copyto!(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:AbstractArray}
+    foreach(recursive_copyto!, a, b)
+    return a
+end
+function recursive_copyto!(a::AbstractArray{T}, b::AbstractArray{T}) where {T}
```
mateuszbaran, line 11:
```suggestion
function recursive_copyto!(a::AbstractArray, b::AbstractArray)
```
This won't force both element types to be the same.

src/helpers/initialize_data.jl
```diff
@@ -0,0 +1,13 @@
+"""
+    recursive_copyto!(a,b)
+
+copy the values from `b` to `a` by recursively copying the contents.
+On the lowest level of `AbstractArray`s this falls back to `copyto!`.
+"""
+function recursive_copyto!(a::AbstractArray{T}, b::AbstractArray{T}) where {T<:AbstractArray}
```
mateuszbaran, line 7:
```suggestion
function recursive_copyto!(a::AbstractArray{T}, b::AbstractArray{U}) where {T<:AbstractArray,U<:AbstractArray}
```
This won't force both element types to be the same.

mateuszbaran (commented):
This should work I think. Anyway, I still need to figure out what to do with nested power manifold.

kellertuer:
Where and what to figure out about `PowerManifold`? This function here works for this case (one of the tests is running on nested power).

mateuszbaran:
The remaining problem is what should mutating variants do for the nested power manifold (mutate the inner array like they do now or replace it). If we leave nested power manifold, we most likely need to update mean and median calculation in Manifolds.jl (to `recursive_copyto!` instead of `copyto!`).

mateuszbaran (approved):

kellertuer:
Ah, yes, I think it would be reasonable to even move `recursive_copyto!` to `ManifoldsBase` when we do this update.
And yes, I would like to keep it, since it is simpler to use (in my opinion).

mateuszbaran:
Sure, `recursive_copyto!` is necessary in any case and should be in ManifoldsBase.jl eventually.

Issue #70, opened 2021-03-22
[Records] get_record for specific symbol


qrebjock:
Hello, it's me again!

From what I understand in the documentation, the following code

```julia
using LinearAlgebra
using Manopt, Manifolds
using Random

m = 5
n = 5
k = 2

function F(M, p)
    return norm(p.S)^2 / 2 - dot(p.S, diag(p.U' * A * p.Vt'))
end

function gradF(M, p::SVDMPoint)
    dS = diagm(p.S)
    return UMVTVector(
        -(I - p.U * p.U') * A * p.Vt',
        dS - p.U' * A * p.Vt',
        -p.U' * A * (I - p.Vt' * p.Vt)
    )
end

A = randn(m, n);
M = FixedRankMatrices(m, n, k)
X0 = random_point(M);

record = gradient_descent(
    M,
    F,
    gradF,
    X0,
    retraction_method=PolarRetraction(),
    record=[:Iteration, :Cost],
    return_options=true
)

get_record(record, :Cost)
```
should return an array containing the sequence of costs.
But it returns an array of tuples containing all the records (:Iteration and :Cost).
Is it the intended behavior?

qrebjock:
Recording gradient norms is pretty common to evaluate the convergence rate of an algorithm.
Right now it is only possible for the gradient descent scheme I think. With trust regions there is no such option.
What do you think about moving the lines
```julia
@doc raw"""
    RecordGradientNorm <: RecordAction

record the current gradient norm value.
"""
mutable struct RecordGradientNorm <: RecordAction
    recorded_values::Array{Float64,1}
    RecordGradientNorm() = new(Array{Float64,1}())
end
function (r::RecordGradientNorm)(p::P, o::O, i::Int) where {P<:Problem,O<:Options}
    return record_or_reset!(r, norm(p.M, o.x, o.gradient), i)
end
```
to `record_options.jl`?
I guess you haven't done it for a good reason but maybe there is a way to fix it.

qrebjock:
Recording gradient norms is pretty common to evaluate the convergence rate of an algorithm.
Right now it is only possible for the gradient descent scheme I think. With trust regions there is no such option.
What do you think about moving the lines
```julia
@doc raw"""
    RecordGradientNorm <: RecordAction

record the current gradient norm value.
"""
mutable struct RecordGradientNorm <: RecordAction
    recorded_values::Array{Float64,1}
    RecordGradientNorm() = new(Array{Float64,1}())
[...]

PR #74, opened 2021-03-31
Refactor record=

kellertuer:
This PR takes up #70 and introduces a machinery to (mostly automatically) assign symbols to `RecordActions` and simplify access to the recorded values.

If one takes a too complicated scheme (for example a `RecordEvery` for a single action within an `RecordGroup`) this might still fail, but this might also be a non-recommended corner case for easy access.

@qrebjock let me know what you think.

_edit_: The tutorial I developed during extending the interface is https://manoptjl.org/previews/PR74/tutorials/HowToRecord.html

Still to do
* [x] improve test coverage and 
* [x] check that the changes are nonbreaking – they are nearly up to the change from `:All` to `:Iteration`.
* [x] check rendered documentation.


kellertuer:
@qrebjock would you maybe take a look if this is what you had in mind? Especially the tutorial here https://manoptjl.org/previews/PR74/tutorials/HowToRecord.html illustrates to what extend the interface was extended to hopefully match your request.

qrebjock:
Hello, yes I replied in the other thread :).
Thanks again for doing this! I am not very familiar with this part of the code so I am very happy that you implemented it.

PR #75, opened 2021-04-07
Typo 'Manidold' on some places of documentation.

fabiom:


kellertuer:
Thanks for noticing and directly fixing it!

PR #78, opened 2021-04-16
Remove recursive_copyto! in favour of copyto!(M,y,x)

kellertuer:
This extends `recursive_copyto!` (the copyto analogue as allocate is for similar) to more types.

@mateuszbaran should the function maybe be moved to base and the implementations provided in manifolds?


mateuszbaran:
Hm... would we actually need `recursive_copyto!` if nested power manifold used non-mutating variants of functions? I thought we did but now I don't really see a good justification for that.

BTW, a nested power manifold that uses non-mutating variants is actually something useful, for example that's the only way to be able to do power manifold for non-mutable point representations.

kellertuer:
Well for me `recursive_copyto!` is the analogue to `copyto!` as `allocate` it the analogue of `similar`, isn‘t it? 

And sure, Power with non-mutating like Oblique (nested power of Circle) uses this already.

mateuszbaran:
At the moment the only reason I can see to have `recursive_copyto!` is to support recursively copying nested power manifold, which could be easily replaced by a non-recursively copying one, which would simplify the interface. `allocate` has many other reasons to exist, for example element type promotion or handling cases where result type depends on the function being used.

kellertuer:
you're right allocate does a few things more. Can you help me see how this can be done “easily”? I don't see how that would work (otherwise I would've chosen the simpler one, especially if it avoids the cases provided in this PR for each and every subtype of `MPoint`and `TVector`).

mateuszbaran:
I think all that needs to be done is adding specialized methods of all mutating functions for the nested power manifold that call non-mutating functions instead of mutating ones. They would look more or less like that:
```julia
function log!(M::NestedPowerManifold, X, p, q)
    rep_size = representation_size(M.manifold)
    for i in get_iterator(M)
        X[M, i] = log(
            M.manifold,
            _read(M, rep_size, p, i),
            _read(M, rep_size, q, i),
        )
    end
    return X
end
```

kellertuer:
do you mean a `log`? calling `log` within `log!` looks a little strange.

but what I also need – at least here is something like deep copy for nested types, that's the main point of `recursive_copyto!`.

kellertuer:
by the way the current log works fine for nested stuff :)

mateuszbaran:
> do you mean a `log`? calling `log` within `log!` looks a little strange.

Yes, I do actually mean that. You the mutate array here but not its elements.

> but what I also need – at least here is something like deep copy for nested types, that's the main point of `recursive_copyto!`.

I'm currently trying to see whether there are any other use cases for `recursive_copyto!` than recursively mutating nested power manifold.

> by the way the current log works fine for nested stuff :)

Yes, apart from breaking `copyto!` and being incompatible with immutable things stored inside (`Oblique` is a good example here because it needed special-casing, it's not a nested power manifold).

kellertuer:
I don't understand your last part, where is this breaking something? I just don't see it.

 Don't get me wrong, but all I want is to have a proper, recursively copying (i.e. not leaving references in the array of arrays) values. Most prominently for initialising data/the first iterate. I don't see another possibility then the current way.

The log – as far as I see works as intended (but also does not need such a copy), so this is maybe another issue?

mateuszbaran:
> Don't get me wrong, but all I want is to have a proper, recursively copying (i.e. not leaving references in the array of arrays) values. Most prominently for initialising data/the first iterate. I don't see another possibility then the current way.

OK, but do you need it for any other manifold than the nested power manifold? Which, as I wrote here, could be modified to not need it (which would also be a good idea for other reasons?)

> The log – as far as I see works as intended (but also does not need such a copy), so this is maybe another issue?

Yes, that's a separate issue, but if the nested power manifold is the only motivation for recursive copyto, then solving that other issue would remove the need to have recursive copyto. And if it's not the only motivation, then knowing the other issues it solves would help designing and documenting `recursive_copyto!` properly.

kellertuer:
> > Don't get me wrong, but all I want is to have a proper, recursively copying (i.e. not leaving references in the array of arrays) values. Most prominently for initialising data/the first iterate. I don't see another possibility then the current way.
> 
> OK, but do you need it for any other manifold than the nested power manifold? Which, as I wrote here, could be modified to not need it (which would also be a good idea for other reasons?)

So for all that are _just_ arrays, I don't need it (so most are just passing down to `copyto!`)
For all subtypes of `MPoint` and `TVector` we also need something like that, since we have to do a `copyto!`(sometimes maybe `reursive_copyto!`? I am not yet sure, might depend on the fields) per field.

> > The log – as far as I see works as intended (but also does not need such a copy), so this is maybe another issue?
> 
> Yes, that's a separate issue, but if the nested power manifold is the only motivation for recursive copyto, then solving that other issue would remove the need to have recursive copyto. And if it's not the only motivation, then knowing the other issues it solves would help designing and documenting `recursive_copyto!` properly.

I don't see how solving the other issue would do that? Solving the log does not provide me with a way of copying values (not references) over...

mateuszbaran:
> For all subtypes of `MPoint` and `TVector` we also need something like that, since we have to do a `copyto!`(sometimes maybe `reursive_copyto!`? I am not yet sure, might depend on the fields) per field.
> 
> > >

For these types you can just add methods to the normal `copyto!` and it will work.

> I don't see how solving the other issue would do that? Solving the log does not provide me with a way of copying values (not references) over...

You just wouldn't have to copy these values recursively.

Anyway, I think I can also imaging this being useful for things like checking whether one point can actually be copied over to another one. So, what about doing: `copyto!(M, q, p)` for copying points on manifold `M` and `copyto!(M, Y, p, X)` for copying tangent vectors? That could work recursively on (recursively copying) nested power manifold.

kellertuer:
you're right, for the types one could just use the classical `copyto!`, I will extend `Manifolds.jl` accordingly the next days.

Oh, I think a `copyto!(M, q, p)`is a very neat idea and sounds less clumsy. It's default would just be to pass to `copyto!` and for nested structures would do the nested (recursive) stuff. I will also address that in a PR for Manifolds.jl – maybe introducing the interface already in ManifoldsBase?

mateuszbaran:
Great! I think that `copyto!(M, q, p)` could go directly to `ManifoldsBase.jl`.

kellertuer:
I will start in ManifoldsBase then.

kellertuer:
Looking at the code, I think I started using that for ProductRepr, too?
For Nested Arrays it acts recursively, but there its mainly a deep copy.
but for ProductRepr just an element wise `copyto!.(M.manifold, y.parts, x.parts)` would be fine, too, right?

kellertuer:
So this might be superseeded by the better `copyto!` but I'll leave it open until I have
* finished the ManifoldsBase.jl PR
* adapted all manifolds in Manifolds.jl
* opened the PR here to adapt to the new and nicer `copyto!`. 

Thanks for your patience and explanations Mateusz!

mateuszbaran:
> Looking at the code, I think I started using that for ProductRepr, too?
> For Nested Arrays it acts recursively, but there its mainly a deep copy.
> but for ProductRepr just an element wise `copyto!.(M.manifold, y.parts, x.parts)` would be fine, too, right?

Yes, for `ProductRepr` is should be just elementwise copy (just please use map instead of broadcasting, it sometimes generates faster code).



> So this might be superseeded by the better `copyto!` but I'll leave it open until I have
> 
> * finished the ManifoldsBase.jl PR
> * adapted all manifolds in Manifolds.jl
> * opened the PR here to adapt to the new and nicer `copyto!`.
> 
> Thanks for your patience and explanations Mateusz!

Good plan :slightly_smiling_face: . Also thank you for this discussion, it was quite useful.

kellertuer:
I already used map here and adapted it for Manifolds.jl already (the PR there is of course still WIP).

kellertuer:
This should work once JuliaManifolds/Manifolds.jl#350 is merged

PR #79, opened 2021-04-20
An alternating gradient descent and a tutorial on Geodesic regression

kellertuer:
This PR is work in progress to illustrate a little example on geodesic regression on manifolds, though again (as so often) on the sphere.
For the second part a simple alternating gradient descent scheme on product manifolds is defined.

* [x] Test
* [x] Thoroughly test documentation that it is clear how the component gradients have to be provided.


kellertuer:
For the now new alternating gradient descent (despite some debugging for the concrete example) a possibility for a vector of step sizes is still missing.

src/plans/alternatig_gradient_plan.jl
```diff
@@ -0,0 +1,287 @@
+@doc raw"""
```
mateuszbaran, line 1: There is a typo in the name of this file (alternatig instead of alternating).

kellertuer: Oh, thanks, currently I seem to have even introduced a bug, the third example currently does not converge anymore.

Issue #86, opened 2021-06-18
Basic `gradient_descent` optimization fails  


Affie:
Hi I'm new to Manopt and I'm trying to figure out how to integrate Manopt into IncrementalInference (also see https://github.com/JuliaRobotics/IncrementalInference.jl/pull/1276 for more complex tries). 

The optimization is formulated as a factor graph and this example represents a simple factor graph with one variable and 2 priors (so basically the minimum of 2 points).  I don't get the results I expect and I made this MWE:

```julia
using Manopt
using Manifolds
using StaticArrays

# "factor"
function Prior_distance(M, meas, p)		
    return distance(M, meas, p)
end

## sanity check factors
M = TranslationGroup(2)

p = SA[1., 0]
q = SA[1., 1]
Prior_distance(M, p, q)

p = SA[0., 1]
q = SA[1., 1]
Prior_distance(M, p, q)

p = SA[0., -1]
q = SA[0., 0]
Prior_distance(M, p, q)

p = SA[0., -1]
q = SA[0., -1]
Prior_distance(M, p, q)

## Try  with Manopt

priorx1 = SA[1.1, 1.0]
priorx2 = SA[0.9, 1.0]

# Factor graph with 2 priors on variable x
function F(M,x) 
    return Prior_distance(M, priorx1, x) + Prior_distance(M, priorx2, x)
end

function gradF(M, x)
    Manifolds.gradient(M,(x)->F(M,x), x)
end

# start value
x0 = SA[0.,0]
# test cost function
F(M, x0)
gradF(M, x0)

# run
xMean = gradient_descent(M, F, gradF, x0)
F(M, xMean)
gradF(M, xMean)

#= results
# I expect [1,1]
julia> xMean = gradient_descent(M, F, gradF, x0)
2-element MVector{2, Float64} with indices SOneTo(2):
  0.9859635170635069
 -0.0031281100934246187

# the cost is still high
julia> F(M, xMean)
2.0163938998034947

# and the gradient also
julia> gradF(M, xMean)
2-element MVector{2, Float64} with indices SOneTo(2):
 -0.02757083871664094
 -1.9899485375183275
=#
```

cc @dehann 

EDIT: I'm on
  [1cead3c2] Manifolds v0.5.4
  [0fc0a36d] Manopt v0.3.9

kellertuer:
Interesting.
[...]

PR #87, opened 2021-06-18
note on step and stop for gradient descent

dehann:
xref https://github.com/JuliaManifolds/Manopt.jl/issues/86#issuecomment-864185971

### DIFF
--- a/src/tutorials/MeanAndMedian.jl
+++ b/src/tutorials/MeanAndMedian.jl
@@ -159,6 +159,19 @@ render_asymptote(export_folder * "/startDataCenterMean.asy"; render=2) #src
 #md #
 #md # ![The resulting mean (orange)](../assets/images/tutorials/startDataCenterMean.png)
 #
+# !!! note
+#
+#     Gradient descent `stepsize` might be constant and not guarenteed to converge in all cases, 
+#     but [can be changed](https://manoptjl.org/stable/solvers/index.html#StoppingCriteria-1) to fit a 
+#     particular case better.  [For example, see the Armijo step size rule](https://manoptjl.org/stable/plans/index.html#Stepsize-1):
+#     ```julia
+#     xMean = gradient_descent(M, F, gradF, x0; stepsize=ArmijoLinesearch(),  debug= [:Iteration, :Cost,"\n",10])
+#     ```
+#     The stopping criterion can also be controlled via keyword
+#     ```julia
+#     stopping_criterion=StopAfterIteration(200) | StopWhenGradientNormLess(10.0^-12)
+#     ```
+#
 # ## Computing the Median
 #
 # !!! note
--- a/src/tutorials/MeanAndMedian.jl
+++ b/src/tutorials/MeanAndMedian.jl
@@ -155,8 +155,7 @@ xMean2 = gradient_descent(
 )
 # which finishes in 5 steaps, just slightly better than the previous computation.
 F(M, xMean) - F(M, xMean2)
-# Note, other optimization cases may vary where adjusting default stepsize 
-# can have a more significant impact on performance.
+# Note that other optimization tasks may have other speedup opportunities.
 #
 # To even have more precision, we can further require a smaller gradient norm. 
 # For this we can change the [`StoppingCriterion`](@ref) used, where several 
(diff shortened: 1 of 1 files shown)


kellertuer:
I would prefer that both get demonstrated also in their effect then, not just as a note, i.e. 
* run the mean with Armijo rule and see that it is faster (less iterations) – show function value afterwards
* then increase tolerance in stopping criterion to illustrate that the cost can even be decreased if we “sharpen” the stopping criterion.

kellertuer:
And looking at the numbers it might be better to only do that with the median, since that runs quite some iterations ;)

kellertuer:
After line 142 in the MeanAndMedian tutorial add something like (with proper formatting Ian `#`s in front of text and such)

A way to get better performance (and convergence) is to switch the default [`ConstantStepsize`](@ref)`(1.0)` with a step size that performs better, for example the [`ArmijoLinesearch`](@ref)`()`.
' Since the mean using a constant step size is already quite fast, we even have to tweak the default values for the `contractionFactor` and the `sufficientDecrease`. We obtain

````
xMean2 = gradient_descent(
    M,
    F,
    gradF,
    data[1];
    stepsize = ArmijoLinesearch(1.0, ExponentialRetraction(), 0.99, 0.5),
    debug=[:Iteration, " | ", :x, " | ", :Change, " | ", :Cost, "\n", :Stop],
)
````

And while it finishes in already 5 steps, it is even (just so slightly) better than the previous computation

````
F(M, xMean) - F(M, xMean2)
````

To even have more precision, we can further require a smaller gradient norm. For this we can change the
[`StoppingCriterion`](@ref) used, where several criteria can be combined using `&` and/or `|`.
If we want to decrease the final gradient (from less that `1e-8`) norm but keep the maximal number of iterations to be 200, we can run

```julia
xMean3 = gradient_descent(
    M,
    F,
    gradF,
    data[1];
    stepsize = ArmijoLinesearch(1.0, ExponentialRetraction(), 0.99, 0.5),
    stopping_criterion = StopAfterIteration(200) | StopWhenGradientNormLess(1e-15),
    debug=[:Iteration, " | ", :x, " | ", :Change, " | ", :Cost, "\n", :Stop],
)
```
so this now takes 10 iterations, we get a very small gradient, but we do not gain that much anymore in the cost itself

```julia
F(M,xMean2)-F(M,xMean3)
```

dehann:
Hi @kellertuer , oh sure I will do -- suggestions sound good.  Sorry I missed these messages yesterday.

dehann:
I'm not 100% familiar with the Literacy.jl documentation support package but hopefully got the text added correctly.  

I copied over your text with a few small grammar updates and one additional sentence (to help avoid the problem both Affie and I stepped in):

> Note, other optimization cases may vary where adjusting default stepsize can have a more significant impact on performance.

dehann:
i see Formatter is giving 3 warnings, best I can make out is:
```
input data has violations
```

but not sure if that is just a bad parse on functions for docs, or if that is an actual warning of interest?

kellertuer:
Concerning the formatting I hope that the Formatter now reports missing format as changes you can just accept, if not, let me know.

src/tutorials/MeanAndMedian.jl
```diff
@@ -140,6 +140,42 @@ xMean = gradient_descent(
...
+    F,
+    gradF,
+    data[1];
+    stepsize=ArmijoLinesearch(1.0, ExponentialRetraction(), 0.99, 0.5),
+    debug=[:Iteration, " | ", :x, " | ", :Change, " | ", :Cost, "\n", :Stop],
+)
+# which finishes in 5 steaps, just slightly better than the previous computation.
+F(M, xMean) - F(M, xMean2)
+# Note, other optimization cases may vary where adjusting default stepsize 
+# can have a more significant impact on performance.
```
kellertuer, line 159:
```suggestion
# Note, other optimization tasks may have other speedup opportunities.
```

src/tutorials/MeanAndMedian.jl
```diff
@@ -140,6 +140,42 @@ xMean = gradient_descent(
 )
 nothing #hide
 #
+# A way to get better performance (and convergence) is to switch the default
```
kellertuer, line 143:
```suggestion
# A way to get better performance and for convex and coercive costs a guaranteed convergence is to switch the default 
```

kellertuer (approved):

PR #91, opened 2021-08-03
Use default retraction and vector transport method in solvers

Nikdwal:
Pretty much the obvious thing to do after the update to ManifoldsBase v0.12.3
https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/79

Summary:
- All methods in `src/solvers` use `default_retraction_method(M)` etc as a default keyword argument. This can't be done in the constructors of various `Options` objects, because they don't receive the manifold as an argument.
- Some automatic changes coming from `JuliaFormatter`
- Corrected typo


src/solvers/ChambollePock.jl
```diff
@@ -47,9 +47,9 @@ For more details on the algorithm, see[^BergmannHerzogSilvaLouzeiroTenbrinckVida
 * `stopping_criterion` – (`stopAtIteration(100)`) a [`StoppingCriterion`](@ref)
 * `update_primal_base` – (`missing`) function to update `m` (identity by default/missing)
 * `update_dual_base` – (`missing`) function to update `n` (identity by default/missing)
-* `retraction_method` – (`ExponentialRetraction()`) the rectraction to use
-* `inverse_retraction_method` - (`LogarithmicInverseRetraction()`) an inverse retraction to use.
-* `vector_transport_method` - (`ParallelTransport()`) a vector transport to use
+* `retraction_method` – (`default_retraction_method(M)`) the rectraction to use
+* `inverse_retraction_method` - (`default_inverse_retraction_method(N)`) an inverse retraction to use.
```
kellertuer, line 51:
```suggestion
* `inverse_retraction_method` - (`default_inverse_retraction_method(M)`) an inverse retraction to use.
```

Are you sure about the `N` here? If I remember correctly, it might even be that both the retraction method and the inverse retraction method are used on _both_ manifolds. So if, it should be consistent and both should use the default from one manifold.

kellertuer (commented):
Thanks, I though of doing that when you finished your last PR (since this is of course straight forward idea) but did not find the time to.

What do you think about the options that can't do that currently, since they do not have a manifold in their constructor? One could unify Options constructors to also always have a manifold as first argument, for example.
The reason that was not done until now is, that Options do not store the manifold (the Problem does), but one could unify that still, that the constructor requires that.

src/solvers/ChambollePock.jl
Nikdwal: I had never heard of Chambolle-Pock prior to this PR. The documentation suggests that the logarithmic map is evaluated at Λ
(p), which is a point on N and that the exponential map and parallel transport are evaluated on M.

kellertuer: Oh, never mind, the paper introducing this algorithm was published only this January; and usually the name Chambolle-Pock is mainly known to the imaging community (other names are along the lines of primal dual hybrid gradient or PDHG, but since we had eRCP and lRCP variants of CP, we took the shorter abbreviations to not write eRPDHG).

As you can see https://manoptjl.org/stable/solvers/ChambollePock.html there is a logarithmic map on N in step 1 and a logarithmic map on M in step 4. The exponential map is only used on M though.

I think the paper also only uses exp/log, but sure for implementation one can easily go for retractions.

Nikdwal:
For the point about who implements this functionality, I don't think I will touch the default retractions in `Manifolds.jl` because I don't know which default is appropriate for each manifold. The only thing I would do there is to retain the default `ExponentialRetraction` for the manifolds that support it, in order to maintain consistency with earlier versions, even if this is not the most efficient. 

I wouldn't change the constructors of Options just for this. If it ain't broke, don't fix it. I'm rather conservative about changing the interface. Besides, most users would just use the solvers, which pass the default retraction down to the Options object they construct.

The test seems to fail because an older version of ManifoldsBase is installed. Do you know why it doesn't use 0.12.3?

src/solvers/ChambollePock.jl
Nikdwal: In that case, it would make sense to modify both the solver and `ChambollePockOptions` to use a different inverse retraction for M and N. That would be for a different PR though.

kellertuer:
I think for most manifolds the implementation detail in Manifolds.jl can stay the same. We should just revise those that do not have an exp/log to use a reasonable default somewhen. That is the benefit we can provide now :)

I am not sure why the test suite errors, wait, I do :)
Manopt.jl carefully uses/imports functions from ManifoldsBase (one could also switch to a `using ManifoldsBase` and import only those that are really newly defined here and there), there your new terms are missing. I am not sure whether we also have to export them then

src/solvers/ChambollePock.jl
kellertuer: Sure, that sounds reasonable but you are right that is for another PR :)

PR #94, opened 2021-10-11
A tutorial on AD (and FD) on manifolds and in the embedding in Pluto

kellertuer:
This is a PR for a first tutorial in Pluto using the (revised) machinery from Manifolds to compute gradients.

Still To Do

* [x] Read the text carefully again
* [x] include the tutorial somehow in the documentation or at least get in rendered into the documentation folder and link to it.


kellertuer:
This one has no interactivity for now, right, so that should be fine. However,
* [ ] Resizing the iframe correctly
* [ ] finding out why Pluto seems unable to load the packages

are the two points missing.
The first does not work at all locally (and nearly not here), the second works locally but does not work here on the CI, which might make these two points hard to tackle.

PR #95, opened 2021-11-03
Pre-compute and reuse the inner-product between the search direction and the Gradient.

johannvk:
When performing the linesearch backtracking algorithm, the inner-product between the search direction `η` and the gradient `gradFx` does not change. 
As it can be expensive computationally to compute this scalar quantity, we can speed up execution by computing this quantity once and storing it in a variable for later use.

Another minor change removes the line
`s = s * contract # correct last` 
as this is redundant when we do not enter the first `while`-loop, and might lead to shorter step-sizes than possible if the initial step size tried out fulfills both requirements enforced by the `while`-loops.

I ran the tests by including `test/runtests.jl`, and they all passed with this modification.


kellertuer:
Thanks for your contribution.

* Can you please check the code formatting (install `JuliaFormatter.jl` and run `using JuliaFormatter; format(".")` to fix the formatting errors?
* Concerning the correction – I think _if_ we enter the first loop, we have to contract the last increase, so please add the line back in but with an if, such that it only corrects if there was an increase before.
* We should also carefully look why and where we use 10 lines of code coverage.

kellertuer:
The Code formatter here runs fine (see checks below), the one check that fails is that currently the code coverage goes down which codecov does not like.

kellertuer:
@johannvk could you maybe check whether we can fix the code coverage going down here? If those lines are not used by your new code, let's put this under #101. It would be great to keen coverage above 99.5 still.

johannvk:
Excuse the delay, but I've now merged the current master into this branch and re-ran the tests locally where they all passed.

It seems the code coverage has also been fixed by merging with the current master, although an Adjoint Differential test fails 
on the "Julia 1.7 - macOS-latest" CI.

kellertuer:
Mac OS latest seems to sometimes end up in `NaN`s, but restarting the test worked fine. Thanks for the update! As a small last step you could bump the version number (in the `Project.toml`) to `0.3.22`, then we can register this small update as a new patch version.

johannvk:
Yup will do with regards to bumping the version. Good to hear that restarting the test worked.

PR #98, opened 2021-12-01
Switch Tutorials to Pluto Notebooks and use PlutoStaticHTML for displaying them in the docs.

kellertuer:
This is just a TestPR to test how the rendering works with GH Actions.

_edit:_
With the newest version of PlutoStaticHTML this works fine, so we only need to transfer the literate files

* [x] Geodesic Regression
* [x] GradientofSecondOrderDifference
* [x] HowToRecord
* [x] JacobiFields
* [x] MeanAndMedian
* [x] StochasticGradientDescent

To pluto notebooks


docs/make.jl
```diff
@@ -52,14 +53,37 @@ for (i, f) in enumerate(pluto_files)
...
+                /* Apparently, Pluto wraps LaTeX inside a span class 'tex'. */
+                var equations = \$('p.tex');
+                for (var i = 0; i < equations.length; i++) {
+                    var text = equations[i].textContent;
+                    text = text.split('\$').join('');
+                    console.log(text);
+                    render(text, equations[i]);
+                }
+            })
+        })
```
rikhuijzer, line 85: I've now changed this to a more elegant KaTeX solution as can be found at <https://rikhuijzer.github.io/PlutoStaticHTML.jl/dev/#LaTeX-equations>. Renders inline and display math both fine: https://huijzer.xyz/posts/correlations/.

kellertuer: Ah but KaTeX is already loaded by documenter in the beginning anyways I think? Well, The options part looks much nicer, let's see how that works on CI :) thanks for providing this directly as a suggestion here, I did not see the LaTeX Equations sections since it is under Franklin. One could maybe restructure that to also have that for Documenter-stuff.

docs/make.jl
```diff
@@ -52,14 +53,49 @@ for (i, f) in enumerate(pluto_files)
...
+
+    <!-- To automatically render math in text elements, include the auto-render extension: -->
+    <script defer src="https://cdn.jsdelivr.net/npm/katex@0.15.1/dist/contrib/auto-render.min.js" integrity="sha384-+XBljXPPiv+OzfbB3cVmLHf4hdUFHlWNZN5spNQ7rmHTXpd7WvJum6fIACpNNfIR" crossorigin="anonymous"
+        onload="renderMathInElement(document.body);"></script>
+        
+  <script>
+  const options = {
+    delimiters: [
+      {left: "$$", right: "$$", display: true},
+      {left: "$", right: "$", display: false},
```
rikhuijzer, line 87:
```suggestion
      {left: "\$\$", right: "\$\$", display: true},
      {left: "\$", right: "\$", display: false},
```
Something like this might solve the issue, but I haven't tested it.

docs/make.jl
rikhuijzer: Oh yes that's very likely. My bad.

kellertuer:
Hm, while it seemed to work in between and it renders fine as HTML (though not rendering math locally) the current rendered version is somehow off since it seems to confuse markdown and html somewhere.

rikhuijzer:
> the current rendered version is somehow off since it seems to confuse markdown and html somewhere.

The developer tools console gives "renderMathInElement undefined":

image

kellertuer:
No it currently also does not work locally – but I am also not an HTML expert, especially not when it even comes to JS, then I am very soon very lost and out of ideas. So maybe your nicer fix and code is not a fix?

kellertuer:
Going back to my code block/Script (before your suggested change) the code works again (only that now also the link atop is correct with the new meta block in the beginning).

kellertuer:
The export works and looks good so far, just that I would like to color input and output cells differently - for best of cases remove the border of the `<pre>` of outputs (they just do not have classes) and maybe remove its upper margin so they are below their input.
I fear this has to be done with JS, since the `:has` property one could use in CSS for checking for children, is not supported by (nearly) any browser.

kellertuer:
I think I found a solution, after I saw how require.js works. Now I have a one-line-JS code that takes all `<code>`s of class `code-output` and adds `pre-output` to its parent `<pre>`.
This is then styled by CSS to have no border, be directly below its (input/code) previous pre (might look ugly if that is not present but currently I do not have those). And voilá
Bildschirmfoto 2021-12-01 um 22 34 54

that is at least my idea of grouping those two cells of in/output. Since I just remove the border and fiddle with spacing – It also works fine in both light and dark mode.

kellertuer:
It (and the tests on 1.7) seem to finally work. Maybe I will also port all other literate tutorials to pluto notebooks, since the new form is really quite nice now (even benchmarks look very fancy!).

docs/Project.toml
```diff
@@ -12,10 +11,11 @@ Manifolds = "1cead3c2-87b3-11e9-0ccd-23c62b72b94e"
...
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
-Documenter = "0.24"
-Manifolds = "0.5"
+Documenter = "0.27"
+Manifolds = "0.5,0.6,0.7"
 ManifoldsBase = "0.11,0.12"
+PlutoStaticHTML = "2.1"
```
rikhuijzer, line 21:
```suggestion
PlutoStaticHTML = "3.4"
```
CI is running v2.1.1. I have no idea why you were locally on a newer version

kellertuer: Ah, I even checked status locally but not the toml - thanks for noticing.

PR #100, opened 2022-01-25
Fix random rotation tangent vectors

kellertuer:
This PR fixes a bug (spotted by @const-ae) and fixed the dimensions of random tangent vectors.
Also updates them to promote element type, i.e. if p and sigma are Float32 – say – the result is as well.


kellertuer:
Let's keep track of the few wobbling test coverage lines in #101.

Issue #99, opened 2022-01-17
Trust-regions returns a point outside the manifold


IWalds:
Hello,

I have the following code, which defines an optimization problem on the oblique manifold, and calls the trust-regions solver on it. I expect the output to belong to the oblique manifold, but it is actually does not (as can be seen from computing the maximal column norm of the output matrix, which should be 1, but is around 1.4).

```julia
using Printf, Manifolds, ManifoldsBase, Manopt

n = 20
p = 2
M = Oblique(p,n)

# Cost matrix
C = zeros(n,n)
for k=1:n-1
    C[k,k+1] = 1 ;
    C[k+1,k] = 1 ;
end

Vinit = [ones(1,8) zeros(1,3) zeros(1,n-11) ;
         zeros(1,8) ones(1,3) -ones(1,n-11)] ;

function cost(M,V)
    # V -> Tr(C*V'*V)
    sum(vec((V*C) .* V))
end

function riem_grad(M,V)
    project(M,V,2*V*C')
end

function riem_hess(M,V,W)
    m = sum((V*C').*V,dims=1)
    WS = W * C - W .* m
    2. * project(M,V,WS)
end

@printf("Maximal column norm (before Trust Regions) : %.4f\n",
        maximum(sqrt.(sum(Vinit.^2,dims=1))))

Vrec = trust_regions(M,cost,riem_grad,riem_hess,Vinit;
                        stopping_criterion=StopAfterIteration(20))

@printf("Maximal column norm (after Trust Regions) : %.4f\n",
        maximum(sqrt.(sum(Vrec.^2,dims=1))))
```

I am unsure what the cause of the problem is. My impression is that there are numerical instabilities in the truncated conjugate gradient descent solver when the iterates come close to a degenerate critical point, which result in `β` becoming quite big, and in `o.δ` leaving the tangent space of the manifold.

If I add a projection step after the update of `o.δ` in the truncated conjugate gradient descent solver, the issue disappears, but I do not know if it is a sensible fix.
```diff
    o.residual = o.residual + α * o.Hδ
    # Precondition the residual.
    o.z = o.randomize ? o.residual : get_preconditioner(p, o.x, o.residual)
    zr = inner(p.M, o.x, o.z, o.residual)
    # Compute new search direction.
    β = zr / o.z_r
    o.z_r = zr
    o.δ = -o.z + β * o.δ
+    o.δ = project!(p.M,o.δ,o.x,o.δ)
```

[...]

PR #102, opened 2022-01-26
Projection to the tangent space in the truncated conjugate gradient solver

IWalds:
This projection can be activated/deactivated through an option
`projection_to_tangent`, in the truncated conjugate gradient
solver and in the trust region solver. By default, I have set
this option to true.

Which function should be used as projection depends on what is
implemented on the manifold. It can be
- the identity (if no projection is defined)
- `project!`
- the composition of `project!` and `embed` if `embed` is
  defined (because the vector to be projected must
  be given to `project!` in "ambient space representation"
  if applicable)

I have found no convenient way to test whether `project!` and
`embed` are defined for a given manifold. I used try/catch
tests, see file hessian_plan.jl.

### DIFF
--- a/src/plans/hessian_plan.jl
+++ b/src/plans/hessian_plan.jl
@@ -66,6 +66,14 @@ a default value is given in brackets if a parameter can be left out in initializ
         initiated with a random tangent vector. If set to true, no
         preconditioner will be used. This option is set to true in some
         scenarios to escape saddle points, but is otherwise seldom activated.
+* `projection_to_tangent` : projection operation to be applied to each
+     iterate; if it is the projection onto the tangent space, it
+     ensures all iterates stay in the tangent space (otherwise, there
+     can be numerical instabilities which make them leave the tangent
+     space). Default is "do nothing".
+     Format : a function of three parameters `(M,p,X)`, where `M` is
+     the manifold, `p` a point on `M` and `X` the tangent vector; the
+     function must rewrite `X` in place and produce no output.
 
 # Constructor
 
@@ -97,6 +105,7 @@ mutable struct TruncatedConjugateGradientOptions{P,T,R<:Real,SC<:StoppingCriteri
     new_model_value::R
     κ::R
     randomize::Bool
+    projection_to_tangent::Any
     initialResidualNorm::Float64
     function TruncatedConjugateGradientOptions(
         p::HessianProblem,
@@ -104,6 +113,7 @@ mutable struct TruncatedConjugateGradientOptions{P,T,R<:Real,SC<:StoppingCriteri
         η::T,
         trust_region_radius::R,
         randomize::Bool;
+        projection_to_tangent=(M,p,X) -> (),
         θ::Float64=1.0,
         κ::Float64=0.1,
         stop::StoppingCriterion=StopWhenAny(
@@ -122,6 +132,7 @@ mutable struct TruncatedConjugateGradientOptions{P,T,R<:Real,SC<:StoppingCriteri
         o.η = η
         o.trust_region_radius = trust_region_radius
         o.randomize = randomize
+        o.projection_to_tangent = projection_to_tangent
         o.model_value = zero(trust_region_radius)
         o.κ = zero(trust_region_radius)
         return o
@@ -145,6 +156,11 @@ a default value is given in brackets if a parameter can be left out in initializ
         random tangent vector. If set to true, no preconditioner will be
         used. This option is set to true in some scenarios to escape saddle
         points, but is otherwise seldom activated.
+* `projection_to_tangent` : projection operation to be applied to each
+     iterate of the inner solver. Default is "do nothing". The other
+     sensitive choice is the projection onto the tangent `(M,p,X) ->
+     project!(M,X,p,X)`, which enforces numerical stability and
+     ensures iterates remain in the tangent space
 * `ρ_prime` : a lower bound of the performance ratio for the iterate that
         decides if the iteration will be accepted or not. If not, the
         trust-region radius will have been decreased. To ensure this,
@@ -181,6 +197,7 @@ mutable struct TrustRegionsOptions{
     max_trust_region_radius::R
     retraction_method::RTR
     randomize::Bool
+    projection_to_tangent
     ρ_prime::R
     ρ_regularization::R
 
@@ -205,7 +222,8 @@ mutable struct TrustRegionsOptions{
         ρ_regularization::R,
         randomize::Bool,
         stopping_citerion::SC,
-        retraction_method::RTR,
+        retraction_method::RTR;
+        projection_to_tangent = (M, p, X) -> (),
     ) where {P,T,SC<:StoppingCriterion,RTR<:AbstractRetractionMethod,R<:Real}
         o = new{P,T,SC,RTR,R}()
         o.x = x
@@ -217,6 +235,7 @@ mutable struct TrustRegionsOptions{
         o.ρ_prime = ρ_prime
         o.ρ_regularization = ρ_regularization
         o.randomize = randomize
+        o.projection_to_tangent = projection_to_tangent
         return o
     end
 end
@@ -230,6 +249,7 @@ function TrustRegionsOptions(
     randomize::Bool,
     stopping_citerion::SC;
     retraction_method::RTR=ExponentialRetraction(),
+    projection_to_tangent = ((M, p, X) -> ()),
 ) where {P,T,R<:Real,SC<:StoppingCriterion,RTR<:AbstractRetractionMethod}
     return TrustRegionsOptions{P,T,SC,RTR,R}(
         x,
@@ -240,7 +260,8 @@ function TrustRegionsOptions(
         ρ_regularization,
         randomize,
         stopping_citerion,
-        retraction_method,
+        retraction_method;
+        projection_to_tangent = projection_to_tangent,
     )
 end
 
--- a/src/plans/hessian_plan.jl
+++ b/src/plans/hessian_plan.jl
@@ -66,14 +66,17 @@ a default value is given in brackets if a parameter can be left out in initializ
         initiated with a random tangent vector. If set to true, no
         preconditioner will be used. This option is set to true in some
         scenarios to escape saddle points, but is otherwise seldom activated.
-* `projection_to_tangent` : projection operation to be applied to each
+* `project_to_tangent!` : projection operation to be applied to each
      iterate; if it is the projection onto the tangent space, it
      ensures all iterates stay in the tangent space (otherwise, there
      can be numerical instabilities which make them leave the tangent
      space). Default is "do nothing".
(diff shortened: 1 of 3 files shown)


IWalds:
(This is a proposition of fix for Issue #99.)

kellertuer:
The usual approach (here and in several methods in Manifolds) is: Assume all required methods are there.
You get a MethodError (currently not always but as soon as ManifoldsBase 0.13.0 is finished) that a method is not defined. That way the user gets the direct feedback, which method is missing.
So you can omit the try catch.

Why the `embed`? Usually the implicit assumption is that the element you have _is_ in the embedding already so there is no need for `embed` I think – or do you have a concrete manifold in mind where this is necessary?

kellertuer:
Oh yeah, don't worry too much about the Julia 1.7 ubuntu/windows tests, one of my tests for an algorithm with a random init seems to run bogus.

IWalds:
Thank you very much for the feedback!

I set the default to true because I think it makes things a bit easier for non-specialists. When I run basic tests with a solver, I prefer to get correct results at first. Understanding that the issue I observe is not due to my code being wrong, but to the solver being numerically unstable, and having the idea that there is an option I can activate to solve it is not completely obvious. I am therefore in favor making it true by default, but I am fine with doing otherwise if you think it is best.

Regarding `embed`, the [documentation](https://juliamanifolds.github.io/Manifolds.jl/stable/interface.html#ManifoldsBase.project!-Tuple{AbstractManifold,%20Any,%20Any,%20Any}) says that `project!` changes the representation of the tangent vector if applicable: the input tangent vector is supposed to be in "ambient space representation", but the projected vector may use an internal representation (with the example of Lie groups, where the internal representation may be based on the Lie algebra structure). My understanding is thus that, if the iterate `δ` is in "internal" representation, it must be converted to "ambient" representation before `project!` is applied.

kellertuer:
Concerning the first, yes it is a numerical instability (but not a bug), so setting the default to something new/different than until now is not a bug fix but a breaking change (which means we have to release a new 0.x version formally) – so before we do that I'd like to be sure that this is really necessary.

Concerning the project and considering, that boolean parameters are in most cases anyways not recommended – what do you think about having a keyword argument which stores the projection function (similar to cost or gradient, properly parametrized in the struct), which 
either defaults to
* `(p,X) -> X` (if we agree on non-breaking fix)
* `(p, X) -> project!(M, X, p, X)`(I would prefer, if possible, the mutating variant)

and one can set it to `(p, X) -> project!(M, X, p, embed!(M, X, p, X))` for the case of Lie groups whose tangent vectors are stored in the Lie algebra.

This last thing can be automated (that is one can automatically check to some extend whether embed is needed or not), I think, once our [new decorator system](https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/91#issuecomment-1024192627) is in place (we might have a `IsLieGroup()` Trait and a `HasLieAlgebraRepresentation()` – or something like that – trait for manifold then, I hope.
I would definitely prefer a non-boolean option anyways

IWalds:
Hi,
I have turned the boolean option into a non-boolean one, which stores the projection function. The default function is to do nothing.
With this choice, as the function has to be supplied by the user, it is not necessary to test if the manifold is a Lie group or anything that would require an extra-embedding step; the user must take the embedding into account him/herself.

kellertuer:
I am not sure why the commit is not part of the PR here – ah you force pushed. Any reason to do so? Maybe - if possible avoid that, it makes the commenting on code a little difficult.

Further, for code style best run

```
using JuliaFormatter; format(".")
```
(after installing JuliaFormatter in Julia and running Julia from the folder of Manopt.jl). That fixes the formatter fail.

IWalds:
Regarding the `::Any` field, I will change it as soon as I can; thank you for pointing me to the `GradientProblem` example.

It seems necessary to me to store `projection_to_tangent` in the `TruncatedConjugateGradientOptions`. As to `TrustRegionsOptions`, it seemed simpler to me to also add a `projection_to_tangent` field there, so that it is available in `initialize_solver!(p::HessianProblem, o::TrustRegionsOptions)` and can be used to construct the truncated gradient options `o.tcg_options`. If you think it is a bad idea, I can try to do differently.

All functions I used are mutating: `o.projection_to_tangent(p.M, o.x, o.δ)` does not return anything, but modifies `o.δ` in place. This is why the default is not the identity, but the function which does nothing `(M, p, X) -> ()`.

I ran the formatter, and integrated most of the modifications it suggested. However, some suggestions would have modified lines of code I had not written myself (in addition to lines I had written). I was not too sure what I was supposed to do with these.

kellertuer:
All functions in Manopt.jl and Manifolds.jl – even _if_ they modify their input, also return the result; compare also `copyto!`, which does the same.

So the identity should actually be `(M, Y, p, X) -> copyto!(M, p, Y, X)` (we have a specific copyto for tangent vectors in ManifoldsBase.jl).

Also note that your current signature `(M, p, X)` is not the mutating signature of project (for tangent vectors). Note the difference of `project(M, p, X)` (non mutating, returns a new element) and `project!(M, Y, p, X)` mutates `Y` (i.e. in Y the projected X is returned, so Y is the corresponding tangent vector at p the projection yields).

IWalds:
I am a bit confused here. What the code should do is:
- in the default case : nothing more than it currently does ;
- in the non-default case : replace `o.δ` with its projection onto the tangent space.

My proposition is to store the desired behavior, under the form of a function, in the field `projection_to_tangent`. In the default case, this function should really be the function which does nothing. I do not see the point in defining it as `copyto!(M,o.δ,x,o.δ)`: it does not make the results of the algorithm different, and slightly increases the computational cost. In the non-default case, defining `projection_to_tangent = (M,p,X) -> project!(M,X,p,X)` and applying `projection_to_tangent` to `(M,x,o.δ)` does what I want.

kellertuer:
Ok, now I see.

I find the notation `(M,p,X) -> project!(M,X,p,X)` misleading or in other words:
Wouldn't it be nice to be able to state `projection_to_tangent = project!` ?
Also the default should be mutating and fit to that so be the copyto ! wrote above.

In your current form, you 
a) would always have to write `(M,p,X) -> project!(M, X, p, X)`
b) the actual call looks misleading, since it looks like the calls to `project(M,p,X)` (though called `project_to_tangent` in the options, for sure). _And_ even more the call looks like you might throw away the result.

I would prefer the call to look like
`projection_to_tangent(M,o.δ,x,o.δ)`or even `o.δ = project_to_tangent(M,o.δ,x,o.δ)` to emphasise that we are using this value further.

So I See that your definition should work, just that it is not so easy to use (as `projection_to_tangent = project!` ) and the call to this function itself is different from the conventions we have in the manifold conventions currently.

IWalds:
Hi,
I modified the pull request; sorry it took me a long time. I corrected the untyped field. Regarding the function which must be passed as a parameter, I changed the notation. In particular, we can now write `project_to_tangent! = project!`. I hope it makes things clearer.

IWalds:
I am sorry, but I still do not see why the function should return something. Indeed, if there is an output, it is ignored anyway.

The projection step is similar to the retraction step in `trust_regions.jl`.
https://github.com/JuliaManifolds/Manopt.jl/blob/5d0da4c0e803f315eedea6862dc178850cc32180/src/solvers/trust_regions.jl#L202-L203
Here, the target vector `o.x_proposal` is replaced by a new value, but the output of `retract!` is forgotten. Therefore, an output is not necessary.

kellertuer:
`retract!` not only modifies its second argument it also returns it. The same for `exp!` or `log!` or `project!` they all modify _and_ return the second argument because that is what Julia usually proposes to do.

IWalds:
Yes, I know that `retract!` returns something. However, here, its return value is ignored. So if `retract!` did not return anything, the code would behave exactly the same. Therefore, I do not see the point in adding an output to functions which do not naturally have one: from my understanding, the only effect it has on the behavior of the code is to make it slightly slower, by adding unnecessary computations.

IWalds:
Ok, I understand. I amended my previous commit to change the default value.

src/solvers/truncated_conjugate_gradient_descent.jl
```diff
@@ -121,12 +132,21 @@ function truncated_conjugate_gradient_descent!(
...
-        p, x, η, trust_region_radius, randomize; θ=θ, κ=κ, stop=stopping_criterion
+        p,
+        x,
+        η,
+        trust_region_radius,
+        randomize;
+        θ = θ,
+        κ = κ,
+        stop = stopping_criterion,
+        (project_to_tangent!)=project_to_tangent!,
```
kellertuer, line 149: Hm, I just noticed 
```suggestion
        project_to_tangent! = project_to_tangent!,
```
I think this should work as well and looks nicer. Or was this a format by the formatter?

src/plans/hessian_plan.jl
```diff
@@ -240,7 +275,8 @@ function TrustRegionsOptions(
         ρ_regularization,
         randomize,
         stopping_citerion,
-        retraction_method,
+        retraction_method;
+        (project_to_tangent!)=project_to_tangent!,
```
kellertuer, line 279:
```suggestion
        project_to_tangent! =project_to_tangent!,
```

src/plans/hessian_plan.jl
```diff
@@ -205,9 +237,10 @@ mutable struct TrustRegionsOptions{
         ρ_regularization::R,
         randomize::Bool,
         stopping_citerion::SC,
-        retraction_method::RTR,
-    ) where {P,T,SC<:StoppingCriterion,RTR<:AbstractRetractionMethod,R<:Real}
-        o = new{P,T,SC,RTR,R}()
+        retraction_method::RTR;
+        project_to_tangent!::Proj=copyto!,
```
kellertuer, line 241: uh, nice, the default even got short here! Nice observation.

kellertuer:
Great, we are getting there.

Can you check for code coverage, i.e. that you lines are covered by tests? And please run formatter

```
using Juliaformatter; format(".");
```
(when starting Julia from the folder of Manopt.jl and you have the JuliaFormatter package installed

src/plans/hessian_plan.jl
kellertuer: (no action necessary just noticed, this got really nice here).

kellertuer:
ah both the non-finishing tests and coverage are not your fault, those I have to fix soon; but please run formatter once, then :) You can also bump the minor version already (to 0.3.18 if I am not mistaken).

Oh is there a reason you force-push? That makes PRs sometimes a little strange, push should work as well.

kellertuer:
Oh, sorry, I just decided to rename the keyword, since `project_vector!`sounds a little nicer :) (and it will be included in the next release)

IWalds:
I am also fine with `project_vector!`. Thank you for merging!

kellertuer:
Thanks for your contribution :)

Issue #97, opened 2021-11-04
Extend Debug actions with a format field


kellertuer:
As proposed in #96 one could increase most Debug actions to use a format string and `printf` for example to have fixed length numbers.

To keep the current style of specifying debug just using symbols ( like `:Cost` or `:Change`), the following scheme could be adapted, that the following entries are allowed (in increasing complexity)
* `:Symbol` calls the debugger without a format and hence uses the default (which should be similar to the current format to be non-breaking)
* `(:Symbol, 3)` digits (space padded, not zero padded) reserved for an integer debug print, similar number of characters for string debugs.
* `(:Symbol, 4,2)` digits before and after the . in a float
* `(:Symbol, 3,2,1)` format a.bec where a has 3, b 2 and c 1 digit space.  
* `(:Symbol, "String")` provides a format 

These tuples should still be easy to read and use.

PR #107, opened 2022-02-12
Debug string format

kellertuer:
Start fixing #97 (proposed by @sweichwald .

# Approach

1. Every debug that prints a number gets a format string that is an `printf` style, this is still set to `"$prefix ..." with a reasonable default from before. This way it is backwards compatible.
2. We introduce a new shortcut for the symbol types in the debug array, for example for `:Cost` using `(:Cost, "| %3.4e |") introduces the format (see 1) in the second argument for the `:Cost` debug.
 
The main ones are changed, for the detailed and specific ones, this has still to be done, so this is a draft for now.


kellertuer:
The main problem seems to be that `@sprintf` can not be used well with non-concrete strings (i.e. if the string is in a variable) and the nicer way (`format`) is not yet available in Julia 1.5. Since I also struggle to get Manopt in Demoed in 1.4 or 1.5 – I might raise the min version – if I do not see good how to format a printf into a string.

mateuszbaran:
I think I'd just bump required version of Julia to 1.6. It's the new LTS so it should be OK. We could drop 1.5 in Manifolds.jl as well.

kellertuer:
Sounds fair with LTS; and this works in 1.6 I checked...

kellertuer:
The main design here is finished – the main (debug options.jl) parts are finished. Now just all dependent and special debugs have to be updated as well to the new signature using `format` instead of `prefix`.

kellertuer:
The special debug prints are still missing, especially the special prints in PrimalDualPlan, that are not just pass-forwards to a field debug print.

kellertuer:
Up to code coverage this should be finished.

PR #112, opened 2022-03-01
Implementation of `check_diff()` according to Boumal (2020) section 4.8

const-ae:
As first discussed in https://github.com/JuliaManifolds/Manifolds.jl/issues/453, I have now started to implement the `check_diff()` function. For me this function has been invaluable, to continuously make sure that my differential are working as expected. 

As discussed with @kellertuer, I made the plotting optional and implemented a method to detect the linear parts of the function and assess if the slope is roughly 2 for a large part of the domain.

There are probably several issues regarding the inclusion of dependencies (`Plots` and `DataFrames`) and other things, as I am still fairly new to using Julia. You have permission to edit the PR, which might be sometimes easier than always waiting for me to figure out how to do something with Julia :)

Best,
Constantin


kellertuer:
Great thanks for the start.
In Manifolds.jl we check whether Plots is loaded and only then add plotting functionality, so that there is no dependency but the package gains funcionality if that is available. I am sure this is possible for the plotting here as well – and I have to check how well that can be done with the DatFrames dependency (not sure why you need that), but this is already a good start.

const-ae:
> and I have to check how well that can be done with the DatFrames dependency (not sure why you need that)

For the linearization of the function part, I translated some R code. The advantage of using a data.frame is that I can reorder and subset multiple vectors at the same time, but maybe there is another elegant solution for this :)

kellertuer:
I am quite sure this is possible without data frames, but I have to take a closer look.

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+    M::AbstractManifold,
+    cost_fnc,
+    diff_fnc;
+    p=random_point(M),
+    X=random_tangent(M, p),
+    log_space_range=range(-8, 0; length=101),
+    plot=false,
+)
+    check_point(M, p)
+    check_vector(M, p, X)
```
kellertuer, line 87:
```suggestion
```
I would propose to leave out the checks. If so, the ValidationManifold would take care of that.

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+function solve_simple_lm(x, y)
+    # See https://en.wikipedia.org/wiki/Simple_linear_regression
+    correlation = cor(x, y)
+    beta = std(y) / std(x) * correlation
+    alpha = mean(y) - beta * mean(x)
+    r2 = size(x, 1) <= 2 ? 1 : correlation^2
+    return (alpha, beta, r2)
+end
+
+function linearize_function(x, y; r2_cutoff=0.99)
```
kellertuer, line 13: I am not sure what this function is for - it is undocumented and has very few comments – also I would prefer having this without the `DataFrame` dependency, what is it used fo? (If the function would be more documented I could propose an alternative...)

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+        b=zeros(n_elem),
+        r2=zeros(n_elem),
+        left=zeros(n_elem),
+        right=zeros(n_elem),
+        length=zeros(n_elem),
+        left_idx=zeros(Int32, n_elem),
+        right_idx=zeros(Int32, n_elem),
+    )
+    counter = 1
+    for i in 1:seq_len
```
kellertuer, line 30: ...this seems to build _all_ possible lines through all data? For two points your initial function to finde the regression line is a little too much (2 points uniquely define a line).

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+        r2=zeros(seq_len),
+        left=zeros(seq_len),
+        right=zeros(seq_len),
+        length=zeros(seq_len),
+        left_idx=zeros(Int32, seq_len),
+        right_idx=zeros(Int32, seq_len),
+    )
+    counter = 1
+    covered_set = collect(1:seq_len)
+    while size(covered_set, 1) > 0
```
kellertuer, line 60: For this while I am really not so sure what it does?

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+    lhs = [cost_fnc(M, geodesic(M, p, X, t)) for t in log_space]
+    rhs = [f0 + t * df0 for t in log_space]
+    err = abs.(rhs - lhs)
+
+    df = linearize_function(log10.(log_space), log10.(err))
+    total_quadratic_length = sum(df[1.8 .< df[:, :b], :length])
+    mostly_quadratic =
+        total_quadratic_length >
+        0.75 * (maximum(log_space_range) - minimum(log_space_range))
+    if plot
```
kellertuer, line 104: we could define a small `diff_check_plot` function that is empty by default but is this inner part if plots was loaded (see how I do that with random and Manifolds).

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+            Plots.plot!(
+                plt,
+                xs,
+                df[i, :a] .+ xs .* df[i, :b];
+                color=df[i, :b] > 1.8 ? "green" : "red",
+            )
+        end
+        return plt
+    else
+        if !mostly_quadratic
```
kellertuer, line 130: returning an exception is not that good here, we do that for our internal checks (but they are called by `is:_point`/`is_vector` which throw them.
We could print a warning or
we could do a similar p´boolean for prints or even to an `kwarg` of `print::Union{IO,Nothing}=nothing` (which one can set to `stdout`  to get a print detail.

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+        for i in 1:size(df, 1)
+            xs = [df[i, :left], df[i, :right]]
+            Plots.plot!(
+                plt,
+                xs,
+                df[i, :a] .+ xs .* df[i, :b];
+                color=df[i, :b] > 1.8 ? "green" : "red",
+            )
+        end
+        return plt
```
kellertuer, line 128: I think besides plotting and the text, the function should just return true/false

src/helpers/check_diff.jl
```diff
@@ -0,0 +1,136 @@
...
+    log_space = 10 .^ log_space_range
+
+    # Based on the Taylor approximation of development of the cost function around point x:
+    # f(R_x(t * X)) = f(x) + t * <grad f(x), X)> + O(t^2)
+    lhs = [cost_fnc(M, geodesic(M, p, X, t)) for t in log_space]
+    rhs = [f0 + t * df0 for t in log_space]
+    err = abs.(rhs - lhs)
+
+    df = linearize_function(log10.(log_space), log10.(err))
+    total_quadratic_length = sum(df[1.8 .< df[:, :b], :length])
```
kellertuer, line 100: Whats the magic number 1.8 here? Some tolerance we are allowed to deviate from 2.0? This also seems the only two fields you actually use from the large build data frame, `:b` and `:length`?
`:b` seems to be the slope we estimated, what‘s length? hEy does it read 75% not 80% below then? What's with slopes above 2?

kellertuer (commented):
Thanks for this nice idea, I wanted tp add checks (gradient / differential) for quite a while already) .
As a comparison, see https://github.com/NicolasBoumal/manopt/blob/master/manopt/tools/checkdiff.m – there one regression is done, not as many as you do.

# comments
*I carefully read your code – if you could provide more detail for one of your functions that would be great
* This PR misses documentation for the check function and testing.
* I wrote an idea how to omit the plots dependency.

kellertuer:
I tried to find a good interface at #117 by doing something write similar for gradients (actually that PR could just reuse functions from you in the end) - for now that documents my ideas for the generic interface - I did not yet have the time to check for the interval where the slope is correct (which I think your one complicated function aims to do).

const-ae:
Hey, thanks for the very detailed comments. I don't have time right now to go through each individual one right now. But I will try to get back to this hopefully next week.

Just to check, do you want to develop this and https://github.com/JuliaManifolds/Manopt.jl/pull/117 in parallel or should I focus on one of them?

kellertuer:
My idea in general would be as follows – I proposed more like the interface / how user should call it. If you check that and see how you could maybe use it (or adapt it and we agree on a common ground) that would be great.

I did not yet do the local/global slope-Check that you do (since I can reuse one if we fine a good way to do). So maybe we focus first on these two things:
* a good interface here (similar/like the one I worked out)
* find a bette way to do the slope check.

src/helpers/check_diff.jl
kellertuer: Ah now I see. We should try to do that without a Dataframe.

My idea (but I will only be able to program something like that maybe next weekend) would be to look for the longest interval (stored in log scale) that fulfils the slope (up to the precision/deviatin I tried to set up over on the other PR) and return that.

If that is the whole range the user reports - all is fine and we return true, if not we give the user the hint (as. `@warn`. maybe?) what the longest interval is and return false (or throw an error depending on what the user specified), so that an user (without the plot) still gets some idea of how good his differential is.

What do you think?

kellertuer:
I think I designed a first interface over on the other branch - I just have to debug the code itself, I do not yet get the right slopes out.

One thing I am unsure about is how to compare slopes of regression lines for different window widths (Manopt only compares over oner window size).

kellertuer:
HI,
just a small update – I think I finished my proposal of a check over in #117. One could now even adapt the function here and narrow `check_gradient` down to defining `df` and calling `check_differential` (I would make the name a little longer). But I also wrote the plot and the search for the best window as separate functions. The best window now looks for the best interval with respect to the slope (i.e. the interval where the slope is closest to 2 overall).

Let me know what you think :)

kellertuer:
After some discussion with @const-ae, this is now included in #117 – but thanks a lot for starting this feature, I was planning to do this for quite some time.

PR #114, opened 2022-03-11
Introduce differentials for geodesics, rename previous ones to shortest geodesics.

kellertuer:
We still have to
* [ ] adapt the examples accordingly
* [ ] adapt bezier differentials accordingly
* [ ] check how to handle non-shortest geodesics, since the current `exp` usage might not yield the right parallel transport then
* [ ] rename the adjoint differentials as well
* [ ] introduce the new geodesic adjoint differentials as well

Since this PR is breaking, we should wait for a few more breaking things before tagging this in a new version.


mateuszbaran:
I did this renaming in https://github.com/JuliaManifolds/ManifoldDiff.jl/commit/c38769dbef67a9fa3c2329f42781c1bc9a06f471 .

kellertuer:
Cool, thanks, I will just close this then (and at some point base this on Diff.

Issue #108, opened 2022-02-26
Allow Step sizes to stop with too small steps and an algorithm to have a corresponding Stopping Criterion.


kellertuer:
Currently step size functions, for example the ArmijoStepsize might run into a deadend when looking for a step size. It would be reasonable to allow for a maximal number of iterations or a minimal step size.

Similarly an algorithm should be able to stop in this case with a suitable stopping criterion.

PR #116, opened 2022-03-25
Introduce a stop_step fallback for backtracking-type line searches and a corresponding stopping criterion

kellertuer:
Based on an idea by @const-ae this PR introduces
* a `stop_step` fallback in the `linesearch_backtrack` to avoid too small steps; which is now an option in `Armijo` and `Nonmonotone` linesearches (as `linesearch_stopsize`). 
to also safely end an algorithm then, this PR also introduces `StopWhenStepsizeLess`.

* [x] check code coverage


kellertuer:
This resolves #108 .

PR #117, opened 2022-04-02
Check gradient and differential functions 

kellertuer:
Inspired by #112 this function starts a similar check for gradients (using the same approach like diff, just on the directional derivative).

- [x] Finish documentation
- [x] unify with `check_diff`
- [x] numerical analysis on the error to fine the interval where the slope matches
- [x] return value true/false
- [x] tests & code coverage.

### DIFF
--- a/src/helpers/check_gradient.jl
+++ b/src/helpers/check_gradient.jl
@@ -27,29 +27,31 @@ function check_gradient(
     X=random_tangent(M, p);
     plot=false,
     error=false,
-    io::Union{IO,Nothing}=nothing.limits = (-8.0, 1.0),
+    io::Union{IO,Nothing}=nothing,
+    limits = (-8.0, 1.0),
     N=101,
     log_range=range(limits[1], limits[2]; length=N),
     retraction_method=default_retraction_method(M),
 )
-    cost = F(M, p)
     gradient = gradF(M, p)
     Xn = X ./ norm(M, p, X) # normalize tangent direction
     is_vector(M, p, gradient, error)
     # function for the directional derivative
     df(M, p, Y) = inner(M, p, gradient, Y)
+    #
+    T = 10 .^(log_range)
     # points p_i to evaluate our error function at
-    points = map(t -> retract(M, p, X, t, retraction_method), log_range)
+    points = map(t -> retract(M, p, X, t, retraction_method), T)
     # F(p_i)
     costs = [F(M, pi) for pi in points]
     # linearized
-    linearized = map(F(M, p) - t * df(M, p, Xn))
+    linearized = map(t-> F(M, p) - t * df(M, p, Xn), T)
     lin_error = abs.(costs .- linearized)
-    plot && plot_slope(log_range, error, 2.0costs[1])
-    max_error = max.(lin_error)
+    max_error = maximum(lin_error)
+    plot && plot_slope(T, lin_error, 2.0, costs[1])
     if io !== nothing
         println(
-            io, "The maximal error in the gradient check is $max_error maximal error is "
+            io, "The maximal error in the gradient check is $max_error."
         )
     end
     # estmate (global) slope
--- a/src/helpers/check_gradient.jl
+++ b/src/helpers/check_gradient.jl
@@ -33,8 +33,8 @@ function check_gradient(
     N=101,
     log_range=range(limits[1], limits[2]; length=N),
     retraction_method=default_retraction_method(M),
-    slope_tol = 0.1,
-    window=nothing
+    slope_tol=0.1,
+    window=nothing,
 )
     gradient = gradF(M, p)
     Xn = X ./ norm(M, p, X) # normalize tangent direction
@@ -60,7 +60,7 @@ function check_gradient(
     y = log10.(lin_error)
     b = std(y) / std(x) * cor(x, y)
     a = mean(y) - b * mean(x)
-    if isapprox(b,2.0; atol=slope_tol)
+    if isapprox(b, 2.0; atol=slope_tol)
         plot && plot_slope(T, lin_error; line_base=costs[1], a=a, b=b, i=1, j=n)
         return true
     end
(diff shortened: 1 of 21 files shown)


kellertuer:
I think I like the interface so far, I must just have messed up the log scale somewhere I fear, so the function itself does not yet work.

And sure the evaluation and regression can be used for both gradient and differential later on (when it works).

src/helpers/check_gradient.jl
```diff
@@ -0,0 +1,119 @@
...
+    # linearized
+    linearized = map(t -> F(M, p) + t * df(M, p, Xn), T)
+    L = abs.(costs .- linearized)
+    # global fit a + bx
+    x = log_range[L .> 0]
+    y = log10.(L[L .> 0])
+    (a, b) = find_best_slope_window(x, y, length(x))[1:2]
+    if isapprox(b, 2.0; atol=slope_tol)
+        plot &&
+            plot_slope(T[L .> 0], L[L .> 0]; line_base=costs[1], a=a, b=b, i=1, j=length(y))
```
const-ae, line 59: `line_base` should be `L[1]`, as otherwise, the dashed line is on a completely different scale.

src/helpers/check_gradient.jl
```diff
@@ -0,0 +1,119 @@
...
+        (io !== nothing) && print(
+            io,
+            "You gradients slope is globally $(@sprintf("%.4f", b)), so within 2 ± $(slope_tol).\n",
+        )
+        return true
+    end
+    # otherwise
+    # find best contiguous window of length w
+    (ab, bb, ib, jb) = find_best_slope_window(log_range, log10.(L), window)
+    plot && plot_slope(T[L .> 0], L[L .> 0]; line_base=costs[1], a=ab, b=bb, i=ib, j=jb)
```
const-ae, line 69: See comment above.

src/helpers/check_gradient.jl
```diff
@@ -0,0 +1,119 @@
...
+            c = cor(x, y)
+            b = std(y) / std(x) * c
+            a = mean(y) - b * mean(x)
+            # look for the best error relative to log scale interval
+            r = (maximum(x) - minimum(x))
+            if abs(b - slope) < abs(b_best - slope)
+                a_best = a
+                b_best = b
+                i_best = j
+                j_best = j + w
```
const-ae, line 114: `j_best` can be outside the range of X. You are using `ib` and `jb` in line 72 to access element of the vector, which can throw an index out of bounds error.

src/helpers/check_gradient.jl
```diff
@@ -0,0 +1,119 @@
...
+            plot_slope(T[L .> 0], L[L .> 0]; line_base=costs[1], a=a, b=b, i=1, j=length(y))
+        (io !== nothing) && print(
+            io,
+            "You gradients slope is globally $(@sprintf("%.4f", b)), so within 2 ± $(slope_tol).\n",
+        )
+        return true
+    end
+    # otherwise
+    # find best contiguous window of length w
+    (ab, bb, ib, jb) = find_best_slope_window(log_range, log10.(L), window)
```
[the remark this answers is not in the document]
kellertuer: More like being lazy. For a zero in `L` the corresponding `y` is `-Inf`. This for sure screws with regression lines - on the global scale (56) this has an effect on the result, when looking for the best ones, they are filtered out again. But You are right, for completeness/correctness we could filter them out here as well.

src/helpers/check_gradient.jl
```diff
@@ -0,0 +1,119 @@
...
+        for j in 1:(n - w + 1)
+            x = X[j:(j + w - 1)]
+            y = Y[j:(j + w - 1)]
+            # fit a line a + bx
+            c = cor(x, y)
+            b = std(y) / std(x) * c
+            a = mean(y) - b * mean(x)
+            # look for the best error relative to log scale interval
+            r = (maximum(x) - minimum(x))
+            if abs(b - slope) < abs(b_best - slope)
```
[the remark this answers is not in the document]
kellertuer: Nice idea! Then we would look for a maximal interval (we luckily have `r` already, so we just look for max of `r`. I am just not 100% sure what to do if `b` is never within tolerance. I'll just check for the best then :)

src/helpers/check_gradient.jl
```diff
@@ -0,0 +1,119 @@
...
+
+From the best line the following data is returned
+
+* `a`, b` specifying the regression line `a + tb`
+* `i`, `j` determining the window, i.e the regression line stems from data `X[i], ..., X[j]`
+"""
+function find_best_slope_window(X, Y, window=nothing; slope=2.0)
+    n = length(X)
+    a_best = 0
+    b_best = 0
```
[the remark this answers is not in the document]
kellertuer: Thanks for spotting this.

src/helpers/check_gradient.jl
kellertuer: Ah, nice - this puts the dashed line atop the right slope if _all_ values fit, which is a nice feature as well.
I was going for the approach from the Matlab code where the dashed line is somewhere but one can still see the slope in comparison.

src/helpers/check_gradient.jl
kellertuer: It can only be, if the user specifies a too large window. So if it is `nothing` then all possible sizes are chosen, then nothing happens. Since j is chosen the right way for any window – a too large window might just yield a negative j (so i_best would cause an error).
I'll add a check for the windows.

src/helpers/check_gradient.jl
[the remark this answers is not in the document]
kellertuer: I now do both - yours as the primary goal, but I keep track of the best slope (as in this code) as long as no good enough slope is found. This way - if `b` is never good enough to be within the tolerance, at lest the best fitting b is returned.

src/helpers/check_gradient.jl
const-ae: I have code that throws an error in https://github.com/JuliaManifolds/Manopt.jl/pull/117/files#diff-83a07696aae40baf5007ef0f950a7bad4548d0180ae80ac2c4397cb75585830bR71

```
check_gradient(TR, cost_fnc, grad_fnc, plot = true)
ERROR: BoundsError: attempt to access 101-element Vector{Float64} at index [102]
Stacktrace:
 [1] getindex
   @ ./array.jl:861 [inlined]
 [2] check_gradient(M::TangentSpaceAtPoint{ℝ, Rotations{3}, Matrix{Float64}}, F::typeof(cost_fnc), gradF::typeof(grad_fnc), p::Matrix{Float64}, X::Matrix{Float64}; plot::Bool, error::Bool, io::Nothing, limits::Tuple{Float64, Float64}, N::Int64, log_range::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, slope_tol::Float64, window::Nothing)
   @ Main /private/tmp/check_grad_review.jl:56
 [3] top-level scope
   @ ~/prog/julia/manifold_derivations/2022-03-22_mnn_overlay.jl:119
```

Unfortunately, the code is not particularly reproducible, but I thought the root cause was this line, but I could also be wrong about this

kellertuer: hm, interesting, surely a one-off-error – for example in the very beginning there were a few ±1s missing in the idices but I had hoped to have resolved them (they do not a--read in my current first test cases).

But very cool that you are already using this :) that helps finding exactly those mistakes. I will have to take a closer look soon into the indices then.

kellertuer: Ha! But I took a first close look and missed to update `j_best` correctly. If you compare what we extract in `x` and `y` -- namely `j:(j+w-1)` (that is _w_ entries j, j+1, ..., j+w-1, since the first we could is basically entry zero), then `j_best` has to be `j+w-1` as well, and this only throws an error if your best window actually reaches into the last index. It's fixed in the newest commit :)

const-ae: Great :)

kellertuer:
This is now unified with the approach from #112 and includes a `check_differential` (which actually can just be used by `check_gradient`).

PR #118, opened 2022-04-02
Introduce a default to warn about ConstantStepsize in gradient_descent

kellertuer:
While ConstantStepsize is an easy start, it might not converge.
To avoid users being confused by increasing values, now a default debug is introduced warning about this.
As soon as a user specifies an own debug, this message vanishes, assuming, that the user knows the details by then.

The warning might also be nice in gerneal to check for increasing cost functions when using heuristic optimisers.

### DIFF
--- a/src/Manopt.jl
+++ b/src/Manopt.jl
@@ -375,7 +375,7 @@ export DebugGradient, DebugGradientNorm, DebugStepsize
 export DebugPrimalBaseChange, DebugPrimalBaseIterate, DebugPrimalChange, DebugPrimalIterate
 export DebugDualBaseChange, DebugDualBaseIterate, DebugDualChange, DebugDualIterate
 export DebugDualResidual, DebugPrimalDualResidual, DebugPrimalResidual
-export DebugProximalParameter
+export DebugProximalParameter, DebugWarnIfCostIncreases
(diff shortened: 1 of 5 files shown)


mateuszbaran:
Looks fine :+1: . My only worry is that performance may suffer on small problems, if I recall correctly debug actions were quite costly.

kellertuer:
That‘s a good point; I could do the warning about the default also hard-coded within the gradient descent.

mateuszbaran:
I don't really like hardcoding, I'll see how costly it is and what can be done to make it cheaper.

mateuszbaran:
Hm, so I've checked performance and I can tell that this debug action is not a big issue -- roughly 2-3 microseconds per iteration. Would be even less if debug options were type stable.

kellertuer:
Nice! I never even cared about performance of this nor about type stability, just because when I started this, I did not know about this and since then I never found the time.

mateuszbaran:
Yes, not much point in caring about it when the cost function almost always is at least an order of magnitude more expensive.

kellertuer:
So we could merge this? Together with the PR I just merged and the Compat bump, this would make a nice new patch release.

mateuszbaran (approved):
Yes, sure!

PR #120, opened 2022-04-05
Update to new ManifoldsBase.jl

mateuszbaran:
This works locally with the ManifoldsBase.jl/Manifolds.jl changes and should be nearly backward-compatible.


kellertuer:
Interesting and nice to know! I am not sure all examples would run, maybe I should have a script to check this.
I started the branch myself as well but got carried away in trying to adapt rand as well. Maybe its a good idea to first to this one and then do rand in a new one.

kellertuer:
Looks good to merge for me :)

kellertuer (approved):

PR #123, opened 2022-04-24
Introduce update functions for cost/gradient/hessian

kellertuer:
In order to do iterative algorithms with cost function updates, this PR aims to allow the problem to be updates (as well).
For the functions this might still be a little tricky, see https://discourse.julialang.org/t/updating-a-function-in-a-type-where-the-function-is-a-parameter/79933/1


kellertuer:
As the thread linked above, this might not be doable without a tremendous rewrite.

mateuszbaran:
What's the problem with just making a new `Problem` object or making a mutable struct for the function in problem if you need to modify the cost, gradient or hessian?

mateuszbaran:
At what points exactly do you need to update the problem?

kellertuer:
> What's the problem with just making a new `Problem` object or making a mutable struct for the function in problem if you need to modify the cost, gradient or hessian?

The problem of making a new `Problem` is that I would have to implement a `copy` / `update` function for every problem (and therein for everything you want to be able to update) even for the mutable case, since the type of the parameters change.
For the Hessian Problem that would be 3 copy functions for example.

A mutable struct for each function (since its parametrised) would be doable, but a little work on the user interface side. One could do `MutableCost` as a struct in `p.cost` sure, but then we would have to have a keyword to activate that? Or would we just tell the user: If you want to be able to update it wrap your cost in a `MutableCost` (which could be a functor?).

kellertuer:
> At what points exactly do you need to update the problem?

Let's say I have an option that stores a problem and options for a subsolver (so a user can use GradientDescent or quasi Newton as a sub solver for example), and in every iteration this sub solver (for example) solves a different Lagrangian function (in the Augmented Lagrangian Method). Then I want to be able to update the subproblem (but keep its type).

mateuszbaran:
OK, so what about something like
```julia
mutable struct MyMutableCostFunction{T}
    params::T
end

(f::MyMutableCostFunction)(M::AbstractManifold, x) = something(f, M, x)

function update_problem!(p::MyMutableCostFunction, update_data)
    p.params .= update_data
end

pp = CostProblem(DefaultManifold(2), MyMutableCostFunction(some_initial_params))
```
And then `update_problem!` could be called as a callback from the solver? Manopt.jl currently does not have callbacks but, for example, Optim.jl does.

I can help with some changes, I've also been thinking about the interface of Manopt.jl recently regarding the geodesic-loxodromes example.

kellertuer:
My first idea was to just make the problem mutable and replace the function, but since its parametrised that idea (which would have been an hour of implementation) does not work, so - without much more of an idea (even with a few ideas from the thread linked above) I do not yet have _any_ idea how to do this.

I (currently) do not even understand what you mean by callback-function?

Fo your code sketch
* what's the `something`?
* i do not want to update the params, I want to update the cost (or the gradient or the hessian)

My problem (mainly in my head maybe) is that introducing

```
struct MyMutableCost
   f::Function
end
````

would not work and parametrising it would make it again non-replaceable?

**Comment by mateuszbaran** [comment]
> I (currently) do not even understand what you mean by callback-function?

A function, specified as an option for the solver, called at a specific fixed point in the iteration of the solver, for example at the end or at the beginning of an iteration, or at any other point you need. Basically a standard way to inject additional operations into the solver.

> * what's the `something`?

It's the implementation of cost function.

> * i do not want to update the params, I want to update the cost (or the gradient or the hessian)

Well, I think you should try to express the update to cost, gradient or hessian as modification of some parameters. That's not always possible but it should be the first choice. Could you write some "old function vs new function" you have trouble turning into a parameter update?

> My problem (mainly in my head maybe) is that introducing
> 
> ```
> struct MyMutableCost
>    f::Function
> end
> ```
> 
> would not work and parametrising it would make it again non-replaceable?

You can always do
```julia
mutable struct MyMutableCost
    f::Function
end
```
but this comes at a significant performance cost.

> (edit: and maybe before we enter in this, we should finish the Manifolds.jl update – though today I fix a few small things for my students here)

These two should be relatively independent but I agree that doing too many things at the same time is not ideal.

kellertuer:
> > I (currently) do not even understand what you mean by callback-function?
> 
> A function, specified as an option for the solver, called at a specific fixed point in the iteration of the solver, for example at the end or at the beginning of an iteration, or at any other point you need. Basically a standard way to inject additional operations into the solver.

Ah, I currently realised that with the decoration of `Options`, that is I dispatch on a different solver step for the debug for example. The scheme is completely different but it is a way to modify the iteration as well.

> > * what's the `something`?
> 
> It's the implementation of cost function.

Then I do not know what `f` is :D 

> 
> > * i do not want to update the params, I want to update the cost (or the gradient or the hessian)
> 
> Well, I think you should try to express the update to cost, gradient or hessian as modification of some parameters. That's not always possible but it should be the first choice. Could you write some "old function vs new function" you have trouble turning into a parameter update?

Hm, I still have the feeling we are not on the same side here, or talking about different things. I have a cost (or gradient) that I want to replace by a new one, I do not have parameters for now. Sure for a Lagrange function one could come up with parameters (mainly `lambda` and `mu` in there are parameters), that is doable, but would solve the update for exactly that problem. Maybe for now, since my student wants to do that exactly for a Lagrange function, that is a good solution writing two structs `LagrangeCost` and `LagrangeGradient` that become cost and gradient and can be updated (in their parameters).

> [...] but this comes at a significant performance cost.

I learned from you to care about performance – at least sometimes :)

mateuszbaran:
> > > I (currently) do not even understand what you mean by callback-function?
> > 
> > 
> > A function, specified as an option for the solver, called at a specific fixed point in the iteration of the solver, for example at the end or at the beginning of an iteration, or at any other point you need. Basically a standard way to inject additional operations into the solver.
> 
> Ah, I currently realised that with the decoration of `Options`, that is I dispatch on a different solver step for the debug for example. The scheme is completely different but it is a way to modify the iteration as well.

Yes, debugging could also be done via a callback.

> Then I do not know what `f` is :D

Just some parameters for the cost function, in a very simple example you could have `(f::MyMutableCostFunction)(M::AbstractManifold, x) = distance(M, f.params, x)`.

> Hm, I still have the feeling we are not on the same side here, or talking about different things. I have a cost (or gradient) that I want to replace by a new one, I do not have parameters for now. Sure for a Lagrange function one could come up with parameters (mainly `lambda` and `mu` in there are parameters), that is doable, but would solve the update for exactly that problem. Maybe for now, since my student wants to do that exactly for a Lagrange function, that is a good solution writing two structs `LagrangeCost` and `LagrangeGradient` that become cost and gradient and can be updated (in their parameters).

Well, if you just want to replace the code of the cost function, you can do that currently in Manopt with no changes by adding a wrapper:
```julia
mutable struct Wrapper
    f::Any
end

(w::Wrapper)(args...; kwargs...) = w.f(args...; kwargs...)

p = CostProblem(DefaultManifold(3), Wrapper(your_cost_function))
# updating
p.cost.f = new_cost_function
```
That won't be fast, but it will work without modifying Manopt. I can later help making if fast, OK?

> > [...] but this comes at a significant performance cost.
> 
> I learned from you to care about performance – at least sometimes :)

Thanks :)

kellertuer:
> Well, if you just want to replace the code of the cost function, you can do that currently in Manopt with no changes by adding a wrapper:
> 
> ```julia
> mutable struct Wrapper
>     f::Any
> end
> 
> (w::Wrapper)(args...; kwargs...) = w.f(args...; kwargs...)
> 
> p = CostProblem(DefaultManifold(3), Wrapper(your_cost_function))
> # updating
> p.cost.f = new_cost_function
> ```
> 
> That won't be fast, but it will work without modifying Manopt. I can later help making if fast, OK?

I think then I would prefer to take directly
```
mutable struct LagrangeFunction
end
```
(storing the original cost f and the two constraint functions g and h and lambda and mu), since the functions f,g,h therein would not change and updating the cost would just be changing the values of lambda and mu and that should be possible since `LagrangeFunction` is mutable, right? 

Would not solve the generic case I wanted to go for here, but the use case I had in mind.

mateuszbaran:
> (storing the original cost f and the two constraint functions g and h and lambda and mu), since the functions f,g,h therein would not change and updating the cost would just be changing the values of lambda and mu and that should be possible since `LagrangeFunction` is mutable, right?

Yes, it should work fine.

> Would not solve the generic case I wanted to go for here, but the use case I had in mind.

I think declaring cost function (or gradient, or hessian) as a mutable struct is a perfectly fine solution in such cases.

kellertuer:
Ok, will ty that then with my student instead of rewriting like in this PR.

PR #128, opened 2022-06-02
Add debugs, records and timer

flgoyens:
Hello, 
I added the following features: 
- DebugGradientNorm() for HessianProblems
- RecordGradientNorm() for HessianProblems and for GradientProblems with QuasiNewtonOptions
- RecordTime() as a RecordAction() that returns the time elapsed at the end of each iteration
- StopWhenTimeElapsed() as a StoppingCriterion() based on the time elapsed.

This is my first contribution so I expect that some revisions and discussions might be needed.  I can say that the features are working fine on my machine. 
Best Florentin


kellertuer:
Thanks @flgoyens first of al for using the Package but also for contributing :)

I am sure we will get this fowrking. Two first things to check (since the tests here are failing) is

1) start Julia from the Manopt.jl folder and run `using JuliaFormatter; format(".")` in Julia REPL (of course also Install JuliaFormatter before :))
2) Currently here the tests are failing, to run them locally, make sure to have Manopt.jl in development mode (that is Manopt is loaded from the local folder you have the git repository and not from Julia Registry). For that again run Julia from the Manopt folder and then type `] dev .` (the `]` changes to package mode, the remainder tells Julia to set Manopt.jl into development mode and load the local version). Then run `] test Manopt` to run the tests. If you see what to fix, fix it, otherwise let me know :)

I am travelling tomorrow, but will try to find time Saturday morning to check the branch locally here as well.

From the current CI, I am wondering: Did you merge the current master and make that correctly? It complained `StopWhenStepsizeLess` would not exist, Ah I see, when merging Master you re-introduced that at one point, I will propose a change here where you can just click accept :)

3) Make sure you have good doc strings, for the timer I would prefer to either have that as a Debug action (or also a record action?) to be able to add that to arbitrary solvers and not just (hard encoded) to Hessian?

src/plans/stopping_criterion.jl
```diff
@@ -43,6 +43,24 @@ function update_stopping_criterion!(c::StopAfterIteration, ::Val{:MaxIteration},
     return c
 end
 
+"""
+    StopWhenTimeElapsed <: StoppingCriterion
+
+stores a threshold indicating when to stop after some elapsed time  [`GradientProblem`](@ref).
+"""
+mutable struct StopWhenTimeElapsed <: StoppingCriterion
```
kellertuer, line 51: Instead of this you can just use

https://manoptjl.org/stable/plans/stopping_criteria/#Manopt.StopAfter

which does the same (and stores the time internally instead of requiring _every_ option to have a `timer` field, which only just one option has. As mentioned in the comment above, I would prefer having a StepTimer debug/record as well similar to the timer stopping criterion (instead of the timer field in options - which would have to be reimplemented in every options).

src/Manopt.jl
```diff
@@ -300,7 +300,7 @@ export StopIfResidualIsReducedByFactor,
     StopWhenTrustRegionIsExceeded,
     StopWhenModelIncreased
 export StopAfterIteration, StopWhenChangeLess, StopWhenGradientNormLess, StopWhenCostLess
-export StopWhenStepsizeLess, StopAfter, StopWhenAll, StopWhenAny
+export StopWhenStepSizeLess, StopAfter, StopWhenAll, StopWhenAny, StopWhenTimeElapsed
```
kellertuer, line 303:
```suggestion
export StopWhenStepsizeLess, StopAfter, StopWhenAll, StopWhenAny, StopWhenTimeElapsed
```

kellertuer:
So I took a short look, Additionally to RecordTime (which is a very good idea!) We could also do `DebugTime` (the `format` thing might be little tricky but maybe you can see how this works by the other examples), but both should store the start time (or a generic timer) internally like `StopAfter` does (which is what you intended to do with `StopWhenTimeElapsed()` Let me know if the `StopAfter` could have an alias with a better name?

The reason to store the time internally is, that the Debug would work with any problem and any options, without them having to store anything (otherwise we would have to add a timer to _all_ options and to every new options we introduce).

kellertuer:
I just looked around a little how to best do the time thing. For debug maybe something like

```
@doc raw"""
    DebugTime()

Measure time and print the intervals. Using `start=true` you can start the timer on construction,
for example to measure the runtime of an algorithm overall (adding)

The measured time is rounded using the given `time_accuracy` and printed after [canonicalization]().

# Keyword Parameters
* `prefix` – (`"Last Change:"`) prefix of the debug output (ignored if you set `format`)
* `io` – (`stdout`) default steream to print the debug to.
* `format` - ( `"$prefix %s"`) format to print the output using an sprintf format, where `%s` is the canonicalized time`.
* `start` – (`false`) indicate whether to start the timer on creation or not.
* `time_accuracy` – (`Millisecond(1)`) round the time to this period before printing the canonicalized time
"""
mutable struct DebugTime <: DebugAction
    io::IO
    format::String
    last_time::Int64
    time_accuracy::Period
    function DebugTime(;
        start=false,
        io::IO=stdout,
        prefix::String="Last Change: ",
        format::String="$(prefix) %s",
        time_accuracy::Period=Dates.Millisecond(1),
    )
        return new(io, format, start ? time_ns() : 0)
    end
end
function (d::DebugTime)(::Problem, ::Options, i)
    if i==0 || d.last_time == 0 # init
        d.last_time = time_ns()
    else
        t = time_ns()
        p = Nanosecond(t-d.last_time)
        d.last_time = t
        Printf.format(d.io, Printf.Format(d.format),  canonicalize(round(p,d.time_accuracy)))
    end
    return nothing
end
```

For Record one could just store the `p` that I constructed here, i.e. `Period` objects?

Tehn in “magic symbols” one could do a 

```
    if :OverallTime in a
        dictionary[:Stop] = DebugTime(;start=true)
    end
```
Which starts a timer and records the time at the end (measuring hence overall runtime – in the DebugFactory (line 508) and a `:Time` (and maybe also the one with a Format) in the `DebugActionFactory`.

What do you think? Then the debug does not need an Option time field :)

I can help getting this started, I just have to check how to push here (or get access to your branch).

flgoyens:
Thanks for all the feedback @kellertuer 👍   It's indeed the fact that most options do not have a "timer" field which is causing the tests to fail. I did not know that StopAfter() existed. I am looking into your proposition above for DebugTime.

kellertuer:
Hi @flgoyens,
did you have time to work on this?
I would be really interested in the DebugTime, sp if you do not have time I could take that one over,
I would start from your branch such that your contribution is still in there.

PR #132, opened 2022-07-11
Add Primal dual riemannian semi smooth newton as a new optimisation method

wdiepeveen:


kellertuer:
Thanks for the PR concerning code overage this already looks quite good (130 of your 149 lines of code are covered)

a) There is 5 lines missing on
https://app.codecov.io/gh/JuliaManifolds/Manopt.jl/compare/132/tree/src/solvers/primal_dual_semismooth_Newton.jl

concerning the return of the options (maybe I should restructure code, since we do this test on every solver currently, but for now that is required)

b) In your plan the two nicely implemented access functions

https://app.codecov.io/gh/JuliaManifolds/Manopt.jl/compare/132/tree/src/plans/higher_order_primal_dual_plan.jl

are only called for the case where the pros is implemented as allocating for the normal, allocating case. The other three are missing.

I will try to find time to review your code here during the week, these show up as comments in this PR as well then to best keep track :)

src/solvers/primal_dual_semismooth_Newton.jl
```diff
@@ -0,0 +1,378 @@
...
+* `primal_stepsize` – (`1/sqrt(8)`) proximal parameter of the primal prox
+* `Λ` (`missing`) the exact operator, that is required if `Λ(m)=n` does not hold;
+`missing` indicates, that the forward operator is exact.
+* `dual_stepsize` – (`1/sqrt(8)`) proximal parameter of the dual prox
+* `reg_param` – (`1e-5`) regularisation parameter for the Newton matrix
+Note that this changes the arguments the `forward_operator` will be called.
+* `stopping_criterion` – (`stopAtIteration(50)`) a [`StoppingCriterion`](@ref)
+* `update_primal_base` – (`missing`) function to update `m` (identity by default/missing)
+* `update_dual_base` – (`missing`) function to update `n` (identity by default/missing)
+* `retraction_method` – (`ExponentialRetraction()`) the rectraction to use
```
kellertuer, line 35: this should be set to `default_retraction_method(M)`.

src/solvers/primal_dual_semismooth_Newton.jl
```diff
@@ -0,0 +1,378 @@
...
+* `inverse_retraction_method` - (`LogarithmicInverseRetraction()`) an inverse retraction to use.
```
kellertuer, line 36: ...you might have guessed by now, this needs to be `default_vector_transport_method(M)`.

src/solvers/primal_dual_semismooth_Newton.jl
```diff
@@ -0,0 +1,378 @@
...
+    # compute step
+    dx = get_vector(p.M, o.x, dx_coords, DefaultOrthonormalBasis())
+    dξ = get_vector(p.N, o.n, dξ_coords, DefaultOrthonormalBasis())
+
+    # do step
+    o.x = retract(p.M, o.x, dx, o.retraction_method)
+    return o.ξ = o.ξ + dξ
+end
+
+function construct_vector(
```
kellertuer, line 210: could you choose here (and for the next function) a little nicer function names? This function name tells us that the function constructs _soma_ vector but it is completely unclear which one and what for. Though this is an internal helper function, it should still get a doc-string explaining the function a little mode.

src/plans/higher_order_primal_dual_plan.jl
```diff
@@ -0,0 +1,234 @@
...
+    dual_stepsize::Float64
+    reg_param::Float64
+    stop::StoppingCriterion
+    update_primal_base::Union{Function,Missing}
+    update_dual_base::Union{Function,Missing}
+    retraction_method::RM
+    inverse_retraction_method::IRM
+    vector_transport_method::VTM
+
+    function PrimalDualSemismoothNewtonOptions(
```
kellertuer, line 124: Recently I switched the constructors for options to start with the manifold (similar to the problem). Though the manifold is not stored within the options, it enables to use the already mentioned `default_retraction_method(M)` and such for the three last warts you have here (and the docs should be adapted here as well for sure).

src/plans/plan.jl
```diff
@@ -11,8 +11,11 @@ include("quasi_newton_plan.jl")
 include("hessian_plan.jl")
 include("proximal_plan.jl")
 include("subgradient_plan.jl")
-include("particle_swarm_plan.jl")
+
 include("primal_dual_plan.jl")
+include("higher_order_primal_dual_plan.jl")
```
kellertuer, line 16: Why this order? Before they were ordered alphabetically. You do require the primal dual before right? Then why not leave the particle swarm up where it was at least?

src/plans/primal_dual_plan.jl
```diff
@@ -1,3 +1,5 @@
+abstract type AbstractPrimalDualProblem{T} <: Problem{T} end
```
kellertuer, line 1: Please add a doc string here as well and see where you best can add that in the documentation (`Problem.md`) as well.

src/plans/primal_dual_plan.jl
```diff
@@ -1,3 +1,5 @@
...
+
 @doc raw"""
     PrimalDualProblem {T, mT <: Manifold, nT <: Manifold} <: PrimalDualProblem} <: Problem{T}
```
kellertuer, line 4: This line is outdated and should be updated, since you introduces the abstract PD problem.

src/solvers/primal_dual_semismooth_Newton.jl
wdiepeveen: Changed

src/solvers/primal_dual_semismooth_Newton.jl
wdiepeveen: Changed

src/plans/higher_order_primal_dual_plan.jl
wdiepeveen: changed

src/solvers/primal_dual_semismooth_Newton.jl
```diff
@@ -0,0 +1,378 @@
+using SparseArrays
```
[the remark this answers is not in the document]
kellertuer: Then we also have to adapt the `Project.toml` and add it as a dependency including a compat entry. Otherwise the JuliaRegistry will deny the update. The easiest to add it to the Project.toml is by activating the environment (in the main folder this time) and add it there. But the compat-entry (line in the `[compat]` section of the file has to be done by hand, stating which version is required (of SparseArrays) to work here.

src/plans/higher_order_primal_dual_plan.jl
kellertuer: This I do not see in the PR, did you commit/push it ;) ?

wdiepeveen: I was too quick claiming this was fixed. Changed now :)

kellertuer: Mo problem, that‘s exactly why one can comment here and mark things as resolved when both have finished the thing discussed.

src/plans/primal_dual_plan.jl
wdiepeveen: I'm a bit confused here with the comment above. Would you please clarify what should happen? (what is outdated and what needs a docstring)

kellertuer: So the thing in the first line (comment above this one) is a new struct that needs a documentation (docstring) and an idea where to introduce that in the actual docs.

but also here in line 4 in the very end it still reads (in the doctoring (` <: Problem{T}`) which is not completely correct if one looks at the (new) actually definition of the problem.

src/solvers/primal_dual_semismooth_Newton.jl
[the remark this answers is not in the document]
wdiepeveen: I've added sparseArrays locally now and the project.toml has changed.

For compute, if I ask for ] status of Sparsearrays I don't get a version... Is this because it is a standard library? (https://docs.julialang.org/en/v1/stdlib/SparseArrays/) Don't we need a version then (like with Dates)?

Please share thoughts on how to proceed :)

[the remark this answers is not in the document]
kellertuer: Ah, I did not remember it was a standard Library. If the tests with Julia 1.6 run fine (we currently require Julia 1.6 at least), then the add should be all thats necessary (that is committing the new Project.toml as well, for sure).

src/plans/primal_dual_plan.jl
wdiepeveen: AbstractPrimalDualProblem:
- has docs now
- docs are called in problem.md
- Changed the docs of some functions that accept the Abstract problem whereas this was documented in the old way (just primalDualProblem)

src/solvers/primal_dual_semismooth_Newton.jl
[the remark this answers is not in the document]
wdiepeveen: I changed the toml, but GitHub might have missed it. Could you check if you see any changes

src/plans/primal_dual_plan.jl
```diff
@@ -1,7 +1,10 @@
+@doc raw"""
```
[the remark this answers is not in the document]
kellertuer: (make sure to locally pull when you accept this change here on the server)

src/solvers/primal_dual_semismooth_Newton.jl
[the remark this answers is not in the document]
wdiepeveen: It was already in the project.toml on master, so nothing changed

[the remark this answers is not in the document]
kellertuer: Oh, I overlooked that, sorry.

kellertuer (approved):

PR #135, opened 2022-07-19
Make types of retraction and VT in line searches constant

mateuszbaran:
As discussed on Slack.


kellertuer (approved):
Takk (thanks)!

We could just bump a version here to register this patch directly?

mateuszbaran:
Sure, I've bumped the version.

PR #137, opened 2022-07-22
Introduces the RTR-SR1 method. 

TomRiem:


kellertuer:
It seems you broke the type of your stopping criterion, see herre

https://github.com/JuliaManifolds/Manopt.jl/runs/7714291637?check_suite_focus=true#step:5:585

Do you switch between power and power_or_factor during the algorithm? I fear that might not be easily possible. If not maybe just the field is too strictly typed?

kellertuer:
I have two short remarks:

1. could you please check for code coverage a little bit? It seems mainly some new code lines in `hessian_plan.jl` are not covered by tests yet
2. The examples are meant to illustrate how to use certain algorithm on more practical examples (while the moreeducational ones are the tutorial ones) - yours seems to be mainly markdown for now, could you elaborate what the goal of the notebook is? If it is a full example we should add it to the list of things to render for the docs (see make.jl the example list/dictionary).

src/plans/hessian_plan.jl
```diff
@@ -405,60 +456,343 @@ function ApproxHessianFiniteDifference(
...
+        I, manifold_dimension(M), manifold_dimension(M)
+    ),
+    basis::B=DefaultOrthonormalBasis(),
+    nu::R=-1.0,
+    evaluation=AllocatingEvaluation(),
+    vector_transport_method::VTR=ParallelTransport(),
+) where {
+    mT<:AbstractManifold,P,G,B<:AbstractBasis{ℝ},R<:Real,VTR<:AbstractVectorTransportMethod
+}
+    if evaluation == AllocatingEvaluation()
```
kellertuer, line 546: What is the function`gradient` here? Also please do not do this with an `if` but with multiple dispatch i.e. by implementing two functions. We could do a function `get_gradient` (we have that for problems and options already but could extend that to a `(M,grad,eval,p)` signature as well.

test/solvers/test_trust_regions_hessian_update.jl
```diff
@@ -0,0 +1,181 @@
...
+    function hess(::Sphere, p::Array{Float64,1}, X::Array{Float64,1})
+        return 2 * (A * X - p * p' * A * X - X * p' * A * p - p * p' * X * p' * A * p)
+    end
+    grad!(::Sphere, X, p) = (X .= 2 * (A * p - p * p' * A * p))
+    function hess!(::Sphere, X, p, Y)
+        X .= 2 * (A * Y - p * p' * A * Y - Y * p' * A * p - p * p' * Y * p' * A * p)
+        return X
+    end
+    M = Sphere(n - 1)
+    x = random_point(M)
```
kellertuer, line 20: Please do not use random points in tests. If we have a bug that only sometimes happens, and this makes this test sometimes fail and not fail – we are in the same setting as for WolfePowell and Tests are super hard to debug if there is randomness in there.

Maybe just take the first unit vector – or even a smaller manifold, we do not have to beat benchmarks in tests, `n=7` is fine as well.

kellertuer:
...and you have to merge master into this branch to get the tests back up and running.

kellertuer:
Plase also locally run formatter (`using JuliaFormatter; format(".")`) to fix formatting :)

kellertuer:
There is one old ApproxBFGS (mutating case) function that is not covered – and the two old step sizes that you replaced by a new one. Are the two old step sizes still needed then? Otherwise (instead of adding tests) we could also remove them

kellertuer:
Running JuliaFormatter local on my machine, I was able to resolve the formatting issue (i.e. you just have to pull this version).

So we might just have to check for the last 8 lines of codecov, where that came from (or whether that's WolfePowell with its random calls that sometimes cover 10 lines and sometimes don't).

kellertuer:
In the code coverage we are now missing coverage for the two old step sizes. Are they still needed? Then they should be covered by tests – if not, they should be deprecated for now (and maybe called in the deprecated test section). Because removing them is a breaking change – which we only do very rare cases (but for example after the other, current, PR is finished.

kellertuer:
Just to finish this, I

* deprecated the two old ones, so they have warnings now what to replace them with
* introduced the update of stopping criteria values also for the new criterion
* made the constructor for the new stopping criterion a little nicer by introducing defaults
* adapted all tests to these changes
* bumped version.

If all tests pass an the coverage is good, this would be good to go if it is fine with you @TomRiem.

PR #138, opened 2022-07-25
Fix ChambollePock TangentBundle methods

mateuszbaran:
I'm fixing an issue introduced by https://github.com/JuliaManifolds/Manifolds.jl/pull/509 , that is TangentBundle no longer implementing `exp`, `log`, `distance` and `parallel_transpot`. Most issues are fixed by just having separate options for tangent bundle in options but one remaining problem is that `DebugEntryChange` requires `distance` to be implemented, which in turn requires `log`. My idea to solve that is to introduce inverse retraction-dependent `distance` approximations in ManifoldsBase.jl, that is something like `distance(M::AbstractManifold, p, q, m::AbstractInverseRetractionMethod) = norm(M, p, inverse_retract(M, p, q, m))`. What do you think about it?


kellertuer:
I think overall this looks fine and thanks for the fix, I am just a little insecure about the naming

`inverse_retraction_method_tb` but have not yet had a good idea how to name it instead. I feel `_tb` is not descriptive enough but writing `tangentbundle` is too long as an overall variable name.

mateuszbaran:
I was also thinking about calling it `inverse_retraction_method_m` and `inverse_retraction_method_n`, maybe that would be fine?

kellertuer:
What about `retraction_method_dual` and ìnvere_retraction_method_dual` since it refers to the n (N) one for the dual variable?

mateuszbaran:
That sounds fine :+1:

kellertuer (approved):
We could shortly check that the examples still run fine (but the tests still do, so probably they'll do as well), but otherwise this looks nice.

mateuszbaran:
I really doubt this change can break any examples.

PR #140, opened 2022-07-28
Really fix Chambolle-Pock

mateuszbaran:
I messed up #138 a bit and here is a fix. I forgot to add `dual` in one place (which made a different test fail with Sasaki) and didn't update all `tb` to `dual`.


kellertuer (approved):
Sorry due to Jetlag I might have also missed that ;)

PR #141, opened 2022-07-28
Fix three places that should use dual VT in Chambolle-Pock

mateuszbaran:
I have no idea how I managed to miss these. This PR will hopefully be the last one for this issue.


kellertuer (approved):
How many PRs do you expect this will take ;) ? Thats why I said: Are the tests and examples still running...

mateuszbaran:
My guess is that the tests are somewhat nondeterministic, I remember seeing the tests pass locally... I hope this is the last Chambolle-Pock fix.

kellertuer:
Hm, there is one vector check in the differentials that sometimes `NaN`s but the rest should be deterministic...

PR #142, opened 2022-07-29
Take adjoint of `Q` rather than transpose

dkarrasch:
This came up in a nanosoldier run in https://github.com/JuliaLang/julia/pull/46196. Transposes of Q's don't have specialized functions and fall back to `AbstractMatrix` methods, in particular in multiplication. Generic multiplication, however, is defined in terms of elementwise `getindex`, which is expensive and very slow in contrast to dense or structured matrices.


kellertuer (approved):
Cool, thanks for spotting this!

PR #143, opened 2022-07-31
Fix StopWhenChangeLess for Particle Swarm.

kellertuer:
Fixes two bugs:

As noticed in JuliaManifolds/Manifolds.jl#514 the particle swarm StopWhenChangeLess assumed implicitly that  M also works as its power manifold (which until that PR Euclidean did). This is now fixed

The sometimes randomly failing test on adjoint forward logarithm is fixed by initialising the in-place variable `Y` to zero since in these elements we add effects. Without that initialisation the computation fails if you pass NaN (in the test) or any nonzero vector (in general).


mateuszbaran (approved):
Thanks!

kellertuer:
The good thing is I finally took the time to also fix the random failing test as well

PR #144, opened 2022-08-04
Custom inverse retraction for `StopWhenChangeLess`

mateuszbaran:
One small change so that `StopWhenChangeLess` can be used on, for example, tangent bundles.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.3.36"
+version = "0.3.37"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"
--- a/src/plans/stopping_criterion.jl
+++ b/src/plans/stopping_criterion.jl
@@ -100,7 +100,7 @@ end
 function (c::StopWhenChangeLess)(P::Problem, O::Options, i)
     if has_storage(c.storage, :x)
         x_old = get_storage(c.storage, :x)
-        d = distance(P.M, O.x, x_old)
+        d = distance(P.M, O.x, x_old, default_inverse_retraction_method(P.M))
         if d < c.threshold && i > 0
             c.reason = "The algorithm performed a step with a change ($d) less than $(c.threshold).\n"
             c.storage(P, O, i)


kellertuer:
Does distance _always_ accept a retraction? I was not yet aware of that. For me that seems a little strange since I thought distance always refers to exp/log (as a fallback with norm). Anything else would be only an approximate distance?

mateuszbaran:
Distance accepting an inverse retraction is a new addition from when I was fixing Chambolle-Pock. For StopWhenChangeLess sometimes an approximation to the exact distance would be enough I think. Here is the relevant PR: https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/119 .

kellertuer:
Oh I am not complaining about the possibility to have an approximate distance with the help of a retraction, I am just confused about the naming (and must have missed that point in the linked PR, I should have asked it there).

kellertuer:
Maybe we could resolve this even more general, that one (alternatively) can specify the field (defaulting to `:x` for now, later maybe to `:Iterate`, see the Frank Wolfe branch for first ideas to decouple :x from the :Iterate) and a measure for distance (which default “just” to `(p,q) -> distance(M,p,q)` but that way you could pass in your favourite distance-like-function yourself?

mateuszbaran:
Sure, this stop condition could be improved further but maybe let's stop at having a better default in this PR?

kellertuer:
If you think this is too much work, sure; the only disadvantage I see is, that this introduces a generality that the future variant might not (again be able to) cover, so the more general change might be breaking

mateuszbaran:
I think whatever future design is going to be it should have the behavior from this PR as a default so the more general change should take it into account. This way users can just use this condition for tangent bundles without thinking that they don't have `log` implemented. Using custom distances should just be additional possibility on top of this, though maybe the default will be implemented slightly differently later.

kellertuer:
Hm,
my thinking is the following:
I would prefer to rename the retractoin-based-distance-approximation (calling it distance might be misleading) - then I would be fine using that name as a default here.
And the more general thing is just then not necessary here.

But you are right that might even be a breaking change in Manifolds and not necessarily here. So let‘s stay with your idea for now here.

kellertuer (approved):

mateuszbaran:
Ah, OK, so I think we can add that renaming issue to https://github.com/JuliaManifolds/Manifolds.jl/issues/438 to keep track of it?

kellertuer:
Yes that would be a good idea, a problem why I stay a little vague is, that I do not have a good idea for that thing yet (`retraction_distance` sounds a little strange to me for example).

PR #146, opened 2022-08-08
Introduce Debug and Record for time measurements

kellertuer:
Following the idea of @flgoyens, this PR takes over #128 and introduces both debug and record for time measurements; either in a cumulative way or an incremental way.

# Todo

* [x] increase test coverage
* [x] check / extend docs


kellertuer:
Oh with the current Pluto notebooks I have to move the tutorial extension to a next PR since this version first has to be registered before we can use it in a tutorial.

So @flgoyens does this do what you had in mind? We now cover cumulative and iterative times and for the recorder even one with just recording the final time.

kellertuer:
@flgoyens I would merge this if you do not mind (and do not protest within a few days ;) ) It should work fine now, is documented and tested.

PR #154, opened 2022-09-27
Use approximate distance for debugging and recording when the exact o…

mateuszbaran:
…ne is not available for a manifold

With Stiefel defaults (https://github.com/JuliaManifolds/Manifolds.jl/pull/538) this makes the original example in https://github.com/JuliaManifolds/Manopt.jl/discussions/153 work.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.3.40"
+version = "0.3.41"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"
--- a/src/plans/debug_options.jl
+++ b/src/plans/debug_options.jl
@@ -127,33 +127,34 @@ during the last iteration. See [`DebugEntryChange`](@ref) for the general case
 * `prefix` – (`"Last Change:"`) prefix of the debug output (ignored if you set `format`)
 * `io` – (`stdout`) default steream to print the debug to.
 * `format` - ( `"$prefix %f"`) format to print the output using an sprintf format.
+* `invretr` - (`nothing`) if not nothing, use the specify inverse retraction to be used
+    for appximating distance.
 """
-mutable struct DebugChange <: DebugAction
+mutable struct DebugChange{TInvRetr<:Union{Nothing,AbstractInverseRetractionMethod}} <:
+               DebugAction
     io::IO
     format::String
     storage::StoreOptionsAction
+    invretr::TInvRetr
     function DebugChange(;
         storage::StoreOptionsAction=StoreOptionsAction((:Iterate,)),
         io::IO=stdout,
         prefix::String="Last Change: ",
         format::String="$(prefix)%f",
+        invretr::Union{Nothing,AbstractInverseRetractionMethod}=nothing,
     )
-        return new(io, format, storage)
+        return new{typeof(invretr)}(io, format, storage, invretr)
     end
 end
 @deprecate DebugChange(a::StoreOptionsAction, pre::String="Last Change: ", io::IO=stdout) DebugChange(;
     storage=a, prefix=pre, io=io
 )
 function (d::DebugChange)(p::Problem, o::Options, i)
+    invretr = d.invretr === nothing ? default_inverse_retraction_method(p.M) : d.invretr
     (i > 0) && Printf.format(
         d.io,
         Printf.Format(d.format),
-        distance(
-            p.M,
-            get_iterate(o),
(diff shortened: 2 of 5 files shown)


mateuszbaran:
We could store the inverse retraction in `DebugChange` and `RecordChange` but that would complicate construction of objects of these types.

kellertuer:
Maybe a kwarg might actually be fine as well and then we store it within the action.

mateuszbaran:
> Maybe a kwarg might actually be fine as well and then we store it within the action.

Sure, done.

kellertuer:
Oh I see the problem: at creation time we do not Janie the manifold, that makes it a little complicated, that's right.

kellertuer:
So as a longer answer: I am not a fan of the `Nothing` possibility – the default should either be log or what ever gets back to normal distance (or is that Nothing? then this would be a little inconsistent to the interplay between exp/retract, which should be the role model for distance here?).

But I see the problem that we do not have the manifold at hand, so it is not so easy to call for the default inverse retraction.

For `Options` and also `Armijo` I made the manifold a mandatory argument to especially be able to call the defaults, but I am not yet sure what to do here, a breaking change seem to be a little too much for this?

mateuszbaran:
No, `nothing` can't be passed to `distance`, it only means (here) that we should take default inverse retraction, determined later when we have the manifold. It would be nice if we had the manifold at the time the Debug/Record object is created but we currently don't.

kellertuer:
The two options that I see here are

1) A breaking change – but then we should change all actions (Record & Debug) to also having the manifold as a mandatory parameter – this might be annoying when generating these yourself and might mean a lot of rework in the chain (or framework) they are embedded in

2) introduce a kwarg `manifold` that is then used for the default. That way the framework is not so much affected for now but one can introduce this later step by step.

I think I am in favour of 2.

Compared to Aptopns or Armijo, the manifold here is really not that important, so making it a kwarg is ok and it can default to the `DefaultManifold` from base of course.

mateuszbaran:
IMO the main problem with option 2 is that it doesn't fully solve the original issue, i.e. the user would have no clue that the solution is to provide the right manifold to the action. Definitely better than the current situation but less user friendly than the approach with `nothing` or mandatory manifold.

Anyway, feel free to pick whichever option you prefer :slightly_smiling_face: .

kellertuer:
I don't understand what you mean. Isn't that mainly a matter of documentation?
I mean without knowing the manifold, we can not magically guess something. And since until now the action is agnostic, which manifold will come along, we can just not guess something.
So when we change that – that is merely a matter of documentation.

An action can not just guess which distance to take. For the generic entry change case I even did that as “provide a function that computed the distance”, that would also be a way one could go for here as well.

mateuszbaran:
> I don't understand what you mean. Isn't that mainly a matter of documentation?

What I mean is: with all these recent fixes (including current state of this PR), the original code posted in that discussion thread just works and does a reasonable thing. With your option 2, it would still give the same vague error (as far as I understand it).

kellertuer:
Yes, with the `Nothing` trick (that we never used before) you postpone the decision which retraction to use to the actual moment where it is called. I am not much a fan of this since it is different from what we (or respectively I in this package) did until now. We never used a nothing.
So in the package design this is quite a break. And I find it non-intuitive to introduce such a second way.

Yes Option 2 would – in the code where this comes from – still yield an error – and one has to add the manifold somewhere. But that would be consistent with how _all other places_ of this package determine the retraction.

mateuszbaran:
It's a bit sad to not be able to defer determination of default value but I see your point -- I've checked and the only place where we use a similar trick is `rand`. All default values come from manifold and the problem here is that we don't have a manifold at the right place.

kellertuer:
But `rand` has the manifold as a second parameter if I see that correctly? Or do you mean for the tangent space?

mateuszbaran:
I mean the methods of `rand` that have `vector_at=nothing`, it's similar (we have `nothing` as a default) but also different because that `nothing` isn't saved anywhere, so it's not the same.

BTW, I'm currently removing the `nothing` option from this PR, I'm just not sure if I should add two keyword arguments (so that users can pass either manifold or inverse retraction) or only one.

kellertuer:
I think two is nice: Manifold means the default is used as the inverse retraction, or one can specify the inverse retraction directly.

mateuszbaran:
Then it gets a little tricky because I have to determine which one was specified. How would you like me to do that?

mateuszbaran:
OK, done :slightly_smiling_face: .

kellertuer:
Thanks, I think this is a good solution – we only have to check in the long run where the optional manifold is suitable are where the mandatory one should ne used.

kellertuer:
The reduction in code coverage is not your fault (just a mishap in stepwise that irregularly happens in Wolfe Power that I still have to fix some day). Can you bump version so we can merge this?

mateuszbaran:
Sure, I will bump version.

PR #159, opened 2022-10-18
Fix Tutorials

kellertuer:
PlutoStaticHTML 6.0.4 introduces some strange error, so we fix it to 6.0.3 for now.


docs/Project.toml
```diff
@@ -22,4 +22,4 @@ BenchmarkTools = "1.3"
 Documenter = "0.27"
 Manifolds = "0.8.27"
 ManifoldsBase = "0.13"
-PlutoStaticHTML = "6"
+PlutoStaticHTML = "6.0.3"
```
rikhuijzer, line 25:
```
PlutoStaticHTML = "=6.0.3"
```
This would have fixed it. `6.0.3` means `≥6.0.3` just as `6` means `≥6` in a `Project.toml`. And yes I know this because I have done a few hours of debugging to figure it out before

kellertuer: oh, well, I might just register _another_ version then.

rikhuijzer: I'm working on a patch already, so probably no need to pin to =6.0.3 for now. 

I have made the patch in https://github.com/rikhuijzer/PlutoStaticHTML.jl/pull/153 and are currently trying it out on Manopt docs

kellertuer: Well, but I want to have docs that are not completely broken as the current release-version docs, so I might just register another patch – also then you are not in a hurry.

kellertuer: > Are you sure registering versions is necessary by the way? If you let your documentation default to the `dev` branch, then there is no need to register changes to docs

I think it is because I expect that people do click on the stable ones because...they are stable. But currently they are horribly broken (also because I was not careful enough before merging, but it worked fine locally and I missed this detail – also today is not really my day it seems).

PR #163, opened 2022-11-11
Introduce the Riemannian ALM and EPM

kellertuer:
Based on the master thesis of @r-dornig, which included this branch, this PR introduces the

* Riemannian Augmented Lagrangian Method (RALM)
* Riemannian Exact Penalty Method (REPM)

based on https://arxiv.org/abs/1901.10000

## ToDo

* [x] Check Docstrings
* [x] Check test coverage
* [x] Add examples / Tutorial -> postponed to this version being registered to add a Pluto notebook covering constrained optimisation.
* [x] Check for the case that the constraints are a vector of functions instead of a function returning a vector – and the same for their gradients – such that evaluations can be done more efficiently if just one element is needed (as mentioned in the literature above.
* [x] rewrite options to be able to just set the inner stopping criterion without setting the options (but remove most max_iter kwargs)


kellertuer:
Atepsize with Armco / WolfePower is a little unstable still, we should fix that test to not use rand.

PR #168, opened 2022-11-21
Introduce memory efficient versions for ALM and Huber-EPM

kellertuer:
Here the gradient elements are only evaluated if necessary to spare computational time and memory. WE might still have to check code coverage – and a tutorial will check the improvements (but only be part of a next PR after this is registered).

### DIFF
--- a/src/plans/alm_plan.jl
+++ b/src/plans/alm_plan.jl
@@ -157,19 +157,74 @@ mutable struct AugmentedLagrangianGrad{Pr,R,T}
     μ::T
     λ::T
 end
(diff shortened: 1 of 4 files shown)


kellertuer:
On the tutorial the new variants finally work, in short instead of g(p) as one function for the constraints (returning a vector) there is now a second variant where g is a vector of functions (that can be separately evaluated). This spares time (about a factor of 3 for the small Tutorial example on S^99); the same for all gradients, as well as h and grad_h.

This PR is just missing The remaining tests for these functions.

Issue #139, opened 2022-07-27
A combined CostGrad structure


kellertuer:
Depending on some feedback from the talk yesterday:

It would be nice to have a function that can reuse parts from the computation of the cost for the gradient and vice versa.

In manopt there is such a method available ( https://www.manopt.org/tutorial.html#costdescription ) in pymanopt there is an issue open ( https://github.com/pymanopt/pymanopt/issues/65 )

A first idea would be to introduce a struct

```
abstract type AbstractCombinedCostAndGrad{P,T} where {P,T} end

struct CombinedCostAndGradient{P,T,F} <: AbstractCombinedCostAndGradient{P,T} where {F}
    p::P
   grad::T
   cost::F
end
```

and a combined function

```
function (gc:: CombinedCostAndGradient )(M,p)
...
end
```
which would compute `c` and rhe `grad` if `p` is different from the `gc.p` and  store `p, grad, c` in the struct. The actual cost and gradient then depend on an initialised `gc` as

```
function cost(M,p)
    gc(M,p)
return gc.cost
end
```

```
function grad(M,p)
    gc(M,p)
return gc.grad
end
```

and one could write a constructor for all abstract gradient problems (or all that have a gradient) that if they are called with a `CombinedCostAndGrad` structures, these two functions would be generated automatically within the generated problem.

Then the user really only has to implement the above new struct type thing as a combined function.

kellertuer:
Thanks for the link, maybe it can also be done easier then (with their last `fg!` idea, though I would also have `fg` there I think – and I want it to easily be compatible with the current state we have here (so just with a preprocessing use the existing gradient problems here).

kellertuer:
The interesting part I have not yet written above that they cover is to only evaluate one even when having `costgrad`(or their `costgrad!`, for the non mutating one I might need an extra argument and not just a `Nothing` like them.

kellertuer:
sure I also know a few cases where I can use that, I am planning to do this issue somewhen, I just do not see the point of using `Nothing` to deactivate that.

kellertuer:
I think I might have an idea based on a discussion I had with a student today, where we actually needed something like this.
The idea is as follows; it would also be an approach to solve #171, since it also covers the cache case. 

We introduce a

```
struct ProblemCache{P;G,C} <: AbstractCache
    p::P # current point
    X::T # current gradient (at p)
    value::F # current cost (at p)
    # maybe further fields.
end
```
[...]

PR #174, opened 2022-12-02
Rework Problem structure to have a Objective for reusing interim computations and a Cache

### LINKED ISSUE #171: Make `DebugWarnIfCostIncreases` not call objective when possible
Since objective is usually calculated in `perform_step!`, it should be possible to reuse the computed value for `DebugWarnIfCostIncreases` instead of recomputing it.

### LINKED ISSUE #172: ManoptBase.jl (or ideas for improving the basic interfaces for Manopt.jl)
Maybe it would be a good idea to introduce a generic interface for algorithms – ManoptBase.jl

I am not yet 100% sure, but something like the Problem and the options and the decorator pattern for the options would be nice to have in a generic form; maybe also the debug/record/stepsize abstract ideas.
One could use this to rethink a few design decisions (most of them were made when I learned Julia). On the other hand, with the sub solvers now being covered, I think the main interface / design ideas are available, so Manopt.jl is mature enough to think about this step

Then Manopt.jl would be more of a “Library of Optimisation Algorithms” (similar to Manifolds.jl).

This issue is meant to keep track of this idea and collect necessary things to think / redesign for that.

# Ideas / Things to change for Manopt 0.4

* rename `Problem` to `AbstractProblem` or maybe even better `RiemannianProblem`
* rename `Options` to `AbstractOptions` or maybe even better `SolverOptions`
* use the new Decorator ideas https://juliamanifolds.github.io/ManifoldsBase.jl/stable/decorator.html for the Options and the stepsize. Maybe we could even make said Decorator pattern its own package before?
* the main function should be called `solve!` not `solve`
* it might be easier (and more flexible) to store the evaluation type with the function(s) not the problem. I thought  about this already, that it might be nice to combine in-place and allocating functions. So the default would still be the allocation type, but the mutating one is a type `<: MutatingXFunction`, so a functor. This way quite a few `get_gradient` and such functions would be much easier, I think (less boiler plate code).
* the functions and differentials could be exported to ManoptExamples.jl
* remove deprecated constructors.
* reorganise code: Move Options to their solvers.
* reorganise Docs similarly

### LINKED ISSUE #177: Differentials and adjoint differentials in ManifoldDiff.jl
Since we are thinking about a breaking release for Manopt.jl, there are a few points that we could discuss. First, some time ago I've copied and improved a bit differentials and adjoint differentials to ManifoldDiff.jl. I think this duplication could be removed for 0.4?

### LINKED ISSUE #178: `random_point` and `random_tangent`
We now have `rand` methods in Manifolds.jl so most (if not all) uses of `random_point` and `random_tangent` can be replaced by `rand` for more consistency. Eventually it would be nice to support custom distributions in Manopt.jl but that's something that needs to be improved over what we currently have in Manifolds.jl.

### LINKED ISSUE #179: Nelder-Mead seems to be much worse than in Optim.jl
See the following code, optiomization of the Rosenbrock function using Optim.jl and Manopt.jl:
```julia
using Optim, Manopt
using Manifolds

f(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
f_manopt(::Euclidean, x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2

optimize(f, [0.0, 0.0], Optim.NelderMead())

M = Euclidean(2)
opts = Manopt.NelderMead(M, f_manopt; return_options=true)
```
when executed I get:
```julia
julia> using Optim, Manopt

julia> using Manifolds

julia> f(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
f (generic function with 1 method)

julia> f_manopt(::Euclidean, x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
f_manopt (generic function with 1 method)

julia> optimize(f, [0.0, 0.0], Optim.NelderMead())
 * Status: success

 * Candidate solution
    Final objective value:     3.525527e-09

 * Found with
    Algorithm:     Nelder-Mead

 * Convergence measures
    √(Σ(yᵢ-ȳ)²)/n ≤ 1.0e-08

 * Work counters
    Seconds run:   0  (vs limit Inf)
    Iterations:    60
    f(x) calls:    117


julia> M = Euclidean(2)
Euclidean(2; field = ℝ)

julia> opts = Manopt.NelderMead(M, f_manopt; return_options=true)
NelderMeadOptions{Vector{Float64}, StopAfterIteration, Float64, Float64, Float64, Float64, ExponentialRetraction, LogarithmicInverseRetraction}([[0.12499192873081919, 0.009135702816670005], [0.12499192873081919, 0.009135702816670005], [0.12499192873081919, 0.009135702816670005]], StopAfterIteration(200000, "The algorithm reached its maximal number of iterations (200000).\n"), 1.0, 2.0, 0.5, 0.5, [0.12499192873081919, 0.009135702816670005], [0.7698476042280331, 0.7698476042280331, 0.7698476042280331], ExponentialRetraction(), LogarithmicInverseRetraction())

julia> f(opts.x)
0.7698476042280331
```
So Manopt calls objective much more times than Optim and still returns a much worse solution. Could that be a bug in Manopt?

### LINKED ISSUE #180: Introduce `default_stepsize(M, O)`
The following code demonstrates how conjugate gradient descent fails on Rosenbrock function using default parameters:
```julia
using Manopt
using Manifolds

f_manopt(::Euclidean, x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2

function g_manopt!(::Euclidean, storage, x)
    storage[1] = -2.0 * (1.0 - x[1]) - 400.0 * (x[2] - x[1]^2) * x[1]
    storage[2] = 200.0 * (x[2] - x[1]^2)
    return storage
end

M = Euclidean(2)
cg_opts = conjugate_gradient_descent(M, f_manopt, g_manopt!, [0.0, 0.0]; evaluation=MutatingEvaluation(), return_options=true)
```
Could we have default parameters that make it work?

kellertuer:
This is a little bit of a (slowly progressing) rework of the Problem structure to enable that a cost and a grad can be the same function and we might be able to cache interims results. This resolves #172.

# Features introduced in Manopt 0.4
* [x] Rename `Problem` -> `AbstractManoptProblem`
* [x] Introduce an `AbstractManifoldObjective` to store all information about the objective 
* [x] Have a `DefaultManoptProblem` when it (just) constains the manifold and an `AbstractManifoldObjective` 
* [x] Introduce new objectives for the old problems, most of the become `<:AbstractManifoldObjective`s within a new `DefaultManoptProblem` (which has the manifold and the objective). 
  * [x] `AlternatingGradientProblem` is now a `ManifoldAlternatingGradientObjective` 
  * [x] `CostProblem` now is a `ManifoldCostObjective`
  * [x] `GradientProlem` now is a `ManifoldGradientObjective`
  * [x] `StochasticGradientProblem` is now a `ManifoldStochasticGradientObjective`
  * [x] `SubgradientProblem` is now a `ManifoldSubgradientObjective`
  * [x] `PrimalDualProblem` is now a `TwoManifoldProblem` with two new cost structures
  * [x] Frank Wolfe Problem
  * [x] `ConstrainedProblem` is now a `ConstrainedManifoldObjective`
  * [x] `ProximalProblem` is now a `ManifoldProximalMapsObjective`
  * [x] `HessianProblem` is now a `ManifoldHessianObjective`
  * [x] `NonlinearLeastSquaresObjective` is now a `NonlinearLeastSquaresObjective`
* [x] Rename ` Options` to `AbstractManoptSolverState`
* [x] Rename `solve` to `solve!`
* [x] be more consistent with the variables following the https://juliamanifolds.github.io/Manifolds.jl/latest/misc/notation.html
* [x] rework docs
  * [x] update all docs
  * [x] rework the solvers start page to give an easier access to which solver to use when.
  * [x] check why a few links are still broken
  * [x] Check Tutorials on 0.4.
* [x] remove old `@deprecated` constructors (those without manifold first – so without manifold defaults)
* [x] Direction update rules (post processors for gradient descent directions like AverageGradient) now have a unified constructor, the manifold is the only mandatory argument, all others are kwargs.
* [x] ~unify usage of `FieldReference` in the accessor functions.~
* [x] Introduce a Caching idea – `SimpleCacheObjective` that stores one recent point and/or gradient in a `storage`, resolves #171 
* [x] remove `random_point` and `random_tangent` in favour or the `rand` functions from Manifolds.jl (if that is loaded), resolves #178
* [x] Introduce a `default_stepsize(M, mp)` that returns a default stepwise functor based on the Manifold and problem combination, resolves #180
* [x] Introduce a `ManifoldCostGradObjective` with one function computing cost and gradient simultaneously #139 
* [x]  ~Think about a generic `CachedManifoldObjective` with a generic cache with more than one value stored.~ postponed to later.
* [x] Refactor `check_gradient` to return the figure but not show it necessarily. That way the plot can be suppressed.
* [x] wrap warnings in tests so they are captured when checked
* [x] depend on `ManifoldDiff` and remove the corresponding functions from here (resolves #177)
* [x] refactor Nelder Mead to be more performant (resolves #179 and #183)
* [x] write a small test for the cache and give text coverage a final check


kellertuer:
Status I: After introducing the new structure, the cost problems (Nelder Mead, particle swarm) seem to work already with only reasonable rework of code.

src/solvers/quasi_Newton.jl
```diff
@@ -125,64 +125,65 @@ function quasi_Newton!(
...
+        qns.gradient,
+        p_old,
+        qns.gradient,
+        qns.x,
+        get_update_vector_transport(qns.direction_update),
     )
-    o.yk = get_gradient(p, o.x) / β - o.gradient
-    update_hessian!(o.direction_update, p, o, x_old, iter)
-    return o
+    qns.yk = get_gradient(mp, qns.x) / β - qns.gradient
```
mateuszbaran, line 171: Could we save this calculated gradient in `qns.gradient` at the end of this function? I'm currently seeing about 4 repeated calculation of gradient at the same point in quasi-Newton and this is one of them.

kellertuer: Is it ok if we use the planned `CachedObjective` for that? Otherwise we would have t introduce a `qns.X`-or-so field.

mateuszbaran: If there are simple algorithmic improvements (like this one) it would be a nice thing to have. At the very least we save a cache lookup.

kellertuer: sure we could do that.

kellertuer: where do you see 4? I see 2 and we could spare that locally with one memory allocation, sure.

mateuszbaran: I've modified my gradient function to print hash of point every time it is evaluated. It prints the same hash value 3 or 4 times in a row, and then it changes to a different one. I haven't tracked the other ones yet, maybe they are more complicated.

kellertuer: if you have debug or a stopping criterion that checks gradient change,... these evaluate as well.

mateuszbaran:
I've fixed the Nelder-Mead issue mentioned here: https://github.com/JuliaManifolds/Manopt.jl/issues/179#issuecomment-1339406316 . Unfortunately, quasi-Newton doesn't seem to work on this branch yet.

kellertuer:
No, because I am slowly working through the solvers, QN (Wolfe-Powell to be precise) and LM are missing, it is really a lot of work.

mateuszbaran:
No problem, I can use the stable version in the meantime :slightly_smiling_face: .

kellertuer:
On Levenberg-Marquard I am currently a little stuck. Most should be done but I am not able to do the different dispatches for the initialisation somehow my assumed signatures for that are not hit and I am not sure why.

```julia
function initialize_solver!(
    dmp::DefaultManoptProblem{mT,NonlinearLeastSquaresObjective{AllocatingEvaluation}},
    lms::LevenbergMarquardtState,
) where {mT}
```

is not hit when coming along with an Allocating LeastSquares Objective. No clue why not, but there is surely some dispatch magic I am missing.

mateuszbaran:
I've fixed that, I will upload the fix in a moment. The issue is that the `<:` was missing. Almost always you want `::SomeType{MyType{...}}` when you specify all type parameters of `MyType` and `::SomeType{<:MyType{...}}` when you don't specify all type parameters of `MyType`.

kellertuer:
Ah just found them myself as well. Would propose to better do the same as for the other accessors where now

```
get_gradient(amp::AbstractProblem, s::AbstractManoptState, ...)
```

passes down to 

```
get_gradient(M::AbstractManifold, amo::AbstractManifoldObjective, o.p, ...)
```

mateuszbaran:
Yes, sure, updating residual and Jacobian accessors makes sense.

kellertuer:
But LevenbergMarquard is the last Test set failing :) Then the restructure is done and we can check the mentioned new features above.

mateuszbaran:
Great! I can help you with this a bit. Caching and CostGrad are the two biggest things, which one would you prefer to do first? They interact a bit since CostGrad computations will require a different caching code.

kellertuer:
My idea was to do the `CostGrad` with a maybe very simple cache first? The simple cache just has one `p` and one `X` that are alway updated by one function (so they are always in sync) and that can be used with CostGrad quite well.

A more compicated Cache would then be the next step.

kellertuer:
Ah and the `SimpleCache` or `CostGradCache` would also just be an objective of course, which internally stores and objective.

mateuszbaran:
I think a very simple cache that stores one or maybe two last evaluations would be perfectly sufficient but potentially it would have to cache objective values, gradients, Hessians, Hessian-vector products, proximal map values, Jacobians etc. and this part could be complicated to handle generically.

kellertuer:
The final problem was the gradient test – is that necessary? The print is already checked in the debug tests, so I feel this is an unnecessary double test.

But we can now at least check for code code for a first time.

Tests are running again.😎

kellertuer:
> I think a very simple cache that stores one or maybe two last evaluations would be perfectly sufficient but potentially it would have to cache objective values, gradients, Hessians, Hessian-vector products, proximal map values, Jacobians etc. and this part could be complicated to handle generically.

I was even thinking simpler here, the CostGradCache would really just do _one_ step and just cost and grad. But we could do more fields (with just one storage) for sure. This would be nearly the same as the storage already does.

mateuszbaran:
> The final problem was the gradient test – is that necessary? The print is already checked in the debug tests, so I feel this is an unnecessary double test.

Which gradient test?

> Tests are running again.

:tada: 

> > I think a very simple cache that stores one or maybe two last evaluations would be perfectly sufficient but potentially it would have to cache objective values, gradients, Hessians, Hessian-vector products, proximal map values, Jacobians etc. and this part could be complicated to handle generically.
> 
> I was even thinking simpler here, the CostGradCache would really just do _one_ step and just cost and grad. But we could do more fields (with just one storage) for sure. This would be nearly the same as the storage already does.

Yes, starting with something simple sounds like a good idea.

kellertuer:
Oh I was not precise enough there, I meant gradient debug test.

mateuszbaran:
I'm not sure why I've added that test, probably it tests that gradient debug can actually extract gradient from RLM.

kellertuer:
but `get_gradient` is the only thing necessary for that and you test that already :)

mateuszbaran:
OK, then I will remove that test :slightly_smiling_face: .

kellertuer:
concerning the `random_point`, `rand` ideas: As far as I can see `rand` is not within the ManifoldsBase definition, so is it useful to switch to that? I do not want to introduce a dependency to Manifolds here.

mateuszbaran:
I see. How about keeping `random_point` and `random_vector` here in Manopt but using `rand` as the default implementation, used for all manifolds from Manifolds.jl?

kellertuer:
I would be open to switching to `rand` here for sure, if it is part of ? ManifoldsBase.jl` – because most randoms here are really more like poor-mans-randoms.

On the other hand it is also not urgent and we can do that later after this PR as well – I can open an issue for that to keep it in mind.

I am nearly done testing and we have a structure for CombinedCostGradObjectives already, so Caching would be next.

mateuszbaran:
OK, then I think it would be best to make `rand` a part of ManifoldsBase.jl.

Do you need some help with designing caching?

kellertuer:
Ok, I might do the rand thing next.

Sure some help with the Cache would be nice. I am still finishing the tests tomorrow and might also start with a little rework of the docs after that.

src/plans/gradient_plan.jl
```diff
@@ -1,82 +1,177 @@
...
+
+# Fields
+
+* `costgrad!!` – a function that computes both the cost ``f\colon\mathcal M → ℝ``
+  and its gradient ``\operatorname{grad}f\colon\mathcal M → \mathcal T\mathcal M``
+
+Depending on the [`AbstractEvaluationType`](@ref) `T` the gradient can have to forms
+
+* as a function `(M, p) -> (c, X)` that allocates memory for the gradient `X`, i.e. an [`AllocatingEvaluation`](@ref)
+* as a function `(X, p) -> (c, X)` that work in place of `X`, i.e. an [`InplaceEvaluation`](@ref)
```
mateuszbaran, line 58:
```suggestion
* as a function `(M, X, p) -> (c, X)` that work in place of `X`, i.e. an [`InplaceEvaluation`](@ref)
```

mateuszbaran:
I think a simple cache could just be something like
```julia
mutable struct CachedCostGradObjective{TP,TCG,TCost<:Number,TGrad}
    p::TP
    costgrad!!::TCG
    cost::TCost
    grad::TGrad
end

function CachedCostGradObjective(M::AbstractManifold, costgrad!!, p0)
    cost, grad = costgrad!!(M, p0)
    return CachedCostGradObjective(p0, costgrad!!, cost, grad)
end

function (cobj::CachedCostGradObjective)(M::AbstractManifold, p)
    if p != cobj.p
        copyto!(M, cobj.p, p)
        cobj.cost, cobj.grad = cobj.costgrad!!(M, p)
    end
    return cobj.cost, cobj.grad
end
```
Plus variants for mutating `costgrad!!` and separate cost and gradient calculation.

At least that's the simplest case. At some points we may only need to evaluate objective without gradient so there should be a nice way to encode that (i.e. I think optimization algorithms should be able tell at which point we need both and at which only the objective).

src/plans/gradient_plan.jl
kellertuer: Ah, right, I have not yet re-read the docs. Thanks for noticing – this was a little inconsistent, since I introduced the M somewhere (from 0.2 to 0.3?) and did not check all docs at that time.

kellertuer:
I was thinking one could maybe wrap the existing `CostGradCache` into a `SompleCache`?

That way one could also wrap a `ManifoldGradientObjective` (the one with 2 functions ;) ) to have a simple cache, then one just has to fine a way to invalidate (cost if grad is updated, and grad if cost is updated).

But for the `ManifoldCombinedCostGradObjective` your approach works very well and might even be the simple version to implement then :)

mateuszbaran:
> I was thinking one could maybe wrap the existing `CostGradCache` into a `SompleCache`?

Hm, what `CostGradCache` and `SimpleCache`? I think I missed something.

kellertuer:
Or I was too inprecise, let me be more precise then (oh maybe I oversubtyped my AbstractManifoldGradientObjetive - we will see)

```julia
mutable struct SimpleCache{E<:AbstractEvaluationType, O<:AbstractManifoldGradientObjective{E,TC,TG}, P, T,C} <: AbstractManifoldGradientObjective{E,TC,TG}
   objective::O
   p::P # a point
   X::T # a vector
   c::C # a value for the cost
end
```

That way we can also use that cache for any other objective that has a gradient and a cost.
We can still dispatch on a certain type of O, sure and do the same as you do for the `ManifoldCostGradientObjective`, but a `ManifoldGradientObjective` (separate cost/grad) could also use this … indeed any objective that has cost and grad (though only to cache these and not other stuff).

The only thing that is a little tricky is to invalidate `c` or `X` if the `p` is updated, i.e.

* if the gradient is evaluated and cached, the point `p` is updated and `c` has to be set to invalid (`NaN`?)
* if the cost is evaluated and cached, the point `p` is updated and `X` has to be set to invalid (maybe with a bool flag in the struct? That's probably the easiest).

Do you think that is too complicated to be called simple ;) ?

kellertuer:
Tested the new `rand` introduced in https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/141 and it was really easy to switch (though tests here will now fail until that is merged/registered).

kellertuer:
I just implemented what I had in mind for the SimpleCache. For the generic case the invalidation is a little tricky (good that I directly wrote tests) - for costGrad that is easier.

I might have overparametrized my Objectives (including TCost and TGrad), but besides that the tests (not yet running on CI, missing ManifoldsBase 0.13.28) run fine and I like the result.

mateuszbaran:
I will take a look but it might not exactly be "simple" :wink: . I will check if this resolves all multiple calls in my PCA example. I suspect it might not.

kellertuer:
It is simple in the sense that it only caches cost and grad and that also only once - I hope it resolves most of those calls, since that was the start of all this ;) And in the tests it works fine as long as one uses get_cost/get_gradient/get_gradient! on the problem or objective.

src/plans/cache.jl
```diff
@@ -0,0 +1,172 @@
...
+#
+function get_cost(M::AbstractManifold, sco::SimpleCacheObjective, p)
+    if sco.p != p || !sco.c_valid
+        # else evaluate cost, invalidate grad if p changed
+        sco.c = get_cost(M, sco.objective, p)
+        # if we switched points, invalidate X
+        (sco.p != p) && (sco.X_valid = false)
+        sco.p = p
+        sco.c_valid = true
+    end
```
mateuszbaran, line 57:
```suggestion
    scop_neq_p = sco.p != p
    if scop_neq_p || !sco.c_valid
        # else evaluate cost, invalidate grad if p changed
        sco.c = get_cost(M, sco.objective, p)
        # if we switched points, invalidate X
        scop_neq_p && (sco.X_valid = false)
        # we need to copy in case the given point gets modified
        copyto!(M, sco.p, p)
        sco.c_valid = true
    end
```

kellertuer: Is that comparison so expensive? If so we should change all other functions in that file as well.

kellertuer: Oh and the copy breaks a lot of things, let's not do that. In the test it suddenly made p equal to q and that should not happen that p changes in the tests. So I am against the copyto, since it has strange outside side effects.

kellertuer: Ah the side effects might come from me passing `p` in when constructing it, then we should at least do a copy in the constructor.

mateuszbaran: It can be somewhat expensive, so I think all functions should be adapted like this.

> Oh and the copy breaks a lot of things, let's not do that. In the test it suddenly made p equal to q and that should not happen that p changes in the tests. So I am against the copyto, since it has strange outside side effects.

> Ah the side effects might come from me passing `p` in when constructing it, then we should at least do a copy in the constructor.

Yes, if the constructor makes a copy then there should be no problem with reusing that. If we don't do `copyto!` on a point, then imagine something like this:
```julia
p = some_point
c = get_cost(M, obj, p)
retract!(M, p, some_vector)
new_cost = get_cost(M, obj, p) # we get the old cached value `c` because `p` is still `p`
```

kellertuer: good catch, but then we should really copy in the constructor, because otherwise the retract will also update the reference in the cache and you always get the initial (cached) cost.

mateuszbaran: Yes, you're right.

kellertuer: I adapted the code. We were both right to some extend that is nice :)

I kept your variable for the comparison, though it should not be necessary since the last comparison is done (on purpose) before the copyto! – together with the copy in the constructor this now works fine.

kellertuer:
I'll propose to leave the more advanced cache for later, since I fear it might take quite some design thinking (how to store multiple points, gradients or anything else, invalidation).

So then only updating the docs (started with the Readme today) and some discussion / decisions about `default_stepsize` are left.

mateuszbaran:
OK, caching should be non-breaking so it can be done later. Are you going to tag 0.4 immediately after merging this PR? There is at least one other breaking change being considered (#177), and maybe #179 also to a small extent.

kellertuer:
Hm, I had hoped this would turn into a PR of finite length, but then maybe it does not. My problem with #177 is that currently Manopt.jl does not depend on Manifolds.jl (that is only weakly with the requirements trick) and would depend explicitly on Manifolds.jl (that is introducing quite some load overhead) when Manopt depends on ManifoldDiff.jl, that's not so nice. I am not sure whether fixing that is of finite time.

For NelderMead I will have to take a closer look at what you mean there, but if you see/can fix that faster, feel free to do #179 on this branch as well.

mateuszbaran:
> Hm, I had hoped this would turn into a PR of finite length, but then maybe it does not. My problem with #177 is that currently Manopt.jl does not depend on Manifolds.jl (that is only weakly with the requirements trick) and would depend explicitly on Manifolds.jl (that is introducing quite some load overhead) when Manopt depends on ManifoldDiff.jl, that's not so nice. I am not sure whether fixing that is of finite time.

I see. I was hoping removing this duplication wouldn't be that hard :sweat_smile: . I guess I could move the function definitions Manopt.jl requires to ManifoldsBase.jl?

> For NelderMead I will have to take a closer look at what you mean there, but if you see/can fix that faster, feel free to do #179 on this branch as well.

Sure, I mostly mean reworking creation of initial simplex. I will try to do it today or tomorrow.

kellertuer:
Thanks for the discussion on `default_stepsize` – that rework worked like a charm and I even found a few more nice effects of that: For a few sub solvers (unified in this PR as well) the sub_state is initialised as a Keyword, and now that one gets also the same default stepsize (which is now only defined in exactly one spot).

Some defaults have kwargs, since I wanted to pass on retraction_method or vector_transport to the default as well.

But I really like the result.

mateuszbaran:
> I already wrote in #177 about the things you import and I think the easiest is to actually define these in ManifoldsDiff and import them to Manifolds.jl - does that sound reasonable? It is anyways about Diff Backends and such.

Yes, I will try to rework that soon (after the Nelder-Mead thing).

> Thanks for the discussion on `default_stepsize` – that rework worked like a charm and I even found a few more nice effects of that: For a few sub solvers (unified in this PR as well) the sub_state is initialised as a Keyword, and now that one gets also the same default stepsize (which is now only defined in exactly one spot).
> 
> Some defaults have kwargs, since I wanted to pass on retraction_method or vector_transport to the default as well.
> 
> But I really like the result.

Cool :sunglasses:

mateuszbaran:
Hmpf... Nelder-Mead is still worse than in Optim. Maybe it's parameter adaptation? I will investigate it but not today.

src/solvers/NelderMead.jl
```diff
@@ -1,6 +1,152 @@
...
+        p,
+        B::AbstractBasis=DefaultOrthonormalBasis();
+        a::Real=0.025,
+        retraction_method::AbstractRetractionMethod=default_retraction_method(M),
+    )
+
+Construct a simplex from a basis `B` with one point being `p` and other points
+constructed by moving by `a` in each principal direction defined by basis `B` of the tangent
+space at point `p` using retraction `retraction_method`. This works similarly to how
+the initial simplex is constructed in the Euclidean Nelder-Mead algorithm.
```
kellertuer, line 24: ...just that the construction here is done in the tangent space at `p` – is what one could add here maybe?

src/solvers/NelderMead.jl
```diff
@@ -1,6 +1,152 @@
...
+            ρ,
+            σ,
+            p,
+            [],
+            retraction_method,
+            inverse_retraction_method,
+        )
+    end
+end
+function NelderMeadState(M::AbstractManifold; kwargs...)
```
kellertuer, line 136: does this method still work to do we have to turn this population into a NelderMeadSimplex first?

src/solvers/NelderMead.jl
```diff
@@ -80,62 +227,75 @@ function NelderMead!(
...
-    if Costr < o.costs[first(ind)] # reflected is better than fist -> expand
-        xe = retract(p.M, m, -o.γ * o.α * ξ, o.retraction_method)
-        Coste = get_cost(p, xe)
+    if Costr < s.costs[first(ind)] # reflected is better than fist -> expand
+        xe = retract(M, m, -s.γ * s.α * ξ, s.retraction_method)
+        Coste = get_cost(mp, xe)
         # successful? use the expanded, otherwise still use xr
-        o.population[last(ind)] .= Coste < Costr ? xe : xr
-        o.costs[last(ind)] = min(Coste, Costr)
+        s.population.pts[last(ind)] .= Coste < Costr ? xe : xr
```
kellertuer, line 263: Maybe we should use `copyto!` here?

src/solvers/NelderMead.jl
```diff
@@ -80,62 +227,75 @@ function NelderMead!(
...
-        Costc = get_cost(p, xc)
-        if Costc < o.costs[last(ind)] # better than last ? -> store
-            o.population[last(ind)] = xc
-            o.costs[last(ind)] = Costc
+    if continue_steps && Costr > s.costs[ind[end - 1]] # even worse than second worst
+        step = (Costr < s.costs[last(ind)] ? -s.ρ : s.ρ)
+        xc = retract(M, m, step * ξ, s.retraction_method)
+        Costc = get_cost(mp, xc)
+        if Costc < s.costs[last(ind)] # better than last ? -> store
+            s.population.pts[last(ind)] = xc
```
kellertuer, line 273: Maybe `copyto!`would again be better / more general?

kellertuer:
Hmpf, but I have too less experience in NelderMead – I just implemented that once because I saw the paper by Dreisigmeyer, so it might also be that Optim.jl uses a different variant maybe?

src/solvers/NelderMead.jl
```diff
@@ -80,62 +227,75 @@ function NelderMead!(
...
-        )
-        # update cost
-        o.costs[ind[i]] = get_cost(p, o.population[ind[i]])
+    if continue_steps
+        for i in 2:length(ind)
+            retract!(
+                M,
+                s.population.pts[ind[i]],
+                s.population.pts[ind[1]],
+                inverse_retract(
```
kellertuer, line 285: I am not sure whether we lose in number of iterations or “just” in time. For time we could allocate one tangent vector (maybe even a temporary vector in the state) where we do this inverse retract inplace  (then in the beginning of the loop, that is before 281), so that we have an allocation less here.
For the rest we might need a large Cache for the cost at least.

src/solvers/NelderMead.jl
kellertuer: ah yes as the tests (that currently fail) indicate we still would need the step via the constructor `NelderMeadState(M, population::Array{P})` or adapt tests and docs (and this function).

src/solvers/NelderMead.jl
mateuszbaran: Sure, I will add a note.

src/solvers/NelderMead.jl
mateuszbaran: Sorry, I didn't notice there is a second file with tests to check. I will adapt it so that population always needs to be passed by `NelderMeadSimplex`. What I don't like about `NelderMeadState(M, population::Array{P})` is that `population` could be confused with initial point.

src/solvers/NelderMead.jl
mateuszbaran: Or maybe just `s.population.pts[last(ind)] = Coste < Costr ? xe : xr`?

src/solvers/NelderMead.jl
mateuszbaran: I don't see how `copyto!` would be more general? This is just stores the entire point in an array.

mateuszbaran:
> Hmpf, but I have too less experience in NelderMead – I just implemented that once because I saw the paper by Dreisigmeyer, so it might also be that Optim.jl uses a different variant maybe?

Yes it does use a different variant, that would be one possible explanation for the difference.

src/solvers/NelderMead.jl
mateuszbaran: Compared to Optim, our Nelder-Mead uses orders of magnitude more objective evaluations and still reaches a worse minimum so optimizing allocation is definitely not the biggest problem here.

kellertuer: Well, without caching we are reevaluating the cost several times for each point I fear, but that would only be solved by the complicated cache.

src/solvers/NelderMead.jl
kellertuer: but FixedRankMatrices have points that are not arrays for example.

src/solvers/NelderMead.jl
kellertuer: see below, I think a copyto would be fine as well, if we want to avoid the allocation.

src/solvers/NelderMead.jl
kellertuer: ah but as a simple solution you could use `@memoize` for your cost in your example that should help already.

mateuszbaran: Our Nelder-Mead had another bug. After fixing it, the convergence it actually similar to Optim.jl.

src/solvers/NelderMead.jl
kellertuer: but population is not of type `P` as initial points are but an array of `P`, but since we are breaking change here anyways, choose what you prefer.

src/solvers/NelderMead.jl
mateuszbaran: `s.population.pts[last(ind)] = Coste < Costr ? xe : xr` does not allocate. It changes what is stored in the vector `pts`. We would need `copyto!` if we had a cache for intermediate points but currently we don't.

src/solvers/NelderMead.jl
kellertuer: Bug? I do not do bugs, I always(!) implement without any errors 😬 – nice that you found it!

kellertuer: Looking forward to what that was ;)

mateuszbaran: :smile:

src/solvers/NelderMead.jl
kellertuer: ah ok, then let's do that.

src/solvers/NelderMead.jl
mateuszbaran: The issue was that the mean calculated at the beginning should not include the worst point in simplex :slightly_smiling_face: .

kellertuer: Indeed, Dreisigmeyer writes that as well, see https://optimization-online.org/2007/08/1742/ page 4, where he excludes p_{n+1}, from the mean \bar p; I must have missed that.

mateuszbaran: I've just fixed it, and also #183, because it was bugging me.

src/solvers/NelderMead.jl
mateuszbaran: What type `P`? Note that `NelderMeadSimplex` is parametrized by point type, not a vector of points type.

src/solvers/NelderMead.jl
mateuszbaran: I think that was resolved and discussed above.

src/solvers/NelderMead.jl
```diff
@@ -280,22 +284,45 @@ function step_solver!(mp::AbstractManoptProblem, s::NelderMeadState, ::Any)
...
+)
+    (i == 0) && (c.reason = "") # reset on init
+    M = get_manifold(mp)
+    max_cdiff = maximum(cs -> abs(s.costs[1] - cs), s.costs[2:end])
+    max_xdiff = maximum(
+        p -> distance(M, s.population.pts[1], p, s.inverse_retraction_method),
+        s.population.pts[2:end],
+    )
+    if max_cdiff < c.tol_f && max_xdiff < c.tol_x
+        c.reason = "The simplex has shrunk below the assumed level (maximum cost difference is $max_cdiff, maximum point distance is $max_xdiff).\n"
```
kellertuer, line 324: NIce. Could we maybe add the iteration number in this string? I find it helpful to read “After 57464 iterations” somewhere in a stopping criterion string.

src/solvers/NelderMead.jl
```diff
@@ -280,22 +284,45 @@ function step_solver!(mp::AbstractManoptProblem, s::NelderMeadState, ::Any)
...
+mutable struct NelderMeadPopulationCriterion{TF<:Real,TX<:Real} <: StoppingCriterion
+    tol_f::TF
+    tol_x::TX
+    reason::String
+    function NelderMeadPopulationCriterion(tol_f::Real=1e-8, tol_x::Real=1e-8)
+        return new{typeof(tol_f),typeof(tol_x)}(tol_f, tol_x, "")
+    end
+end
+
+function (c::NelderMeadPopulationCriterion)(
```
kellertuer, line 313: I usually prefer names with “StopWhen” and it seems to be a “If the costs differ less enough from the first and the distances differ less enough”?

Maybe `StopWhenPopulationConcentrated` or something?

src/solvers/NelderMead.jl
kellertuer: For me there is no confusion of the `population::AbstractArray{P}` compared to a single initial point `p::P` since the array indicates it is an array of points and nor a single one

kellertuer: (where `P` is the point type as usual in Manopt.jl)

mateuszbaran: My confusion comes from the fact that `method_name(manifold, function, some_vector)` for many other method means "use `some_vector` as the starting point". `NelderMead(M, f, some_array)` could equally well mean `NelderMead(M, f, NelderMeadSimplex(some_array))` and `NelderMead(M, f, NelderMeadSimplex(M, some_array))` which have very different meanings.

src/solvers/NelderMead.jl
mateuszbaran: OK, I'll rename it to `StopWhenPopulationConcentrated`.

src/solvers/NelderMead.jl
kellertuer: oh, well, the subtle part is that its not `some_vector` usually but `p::P` (not even unified to that) compared to NelderMead where it is `population::AbstractArray{P}` for me that would be fine – but if you feel that with the simplex type it is more clearer, that is also fine with me; though I do not see why the first one would not be clear.

src/solvers/NelderMead.jl
```diff
@@ -1,6 +1,152 @@
...
+"""
+    NelderMeadSimplex
+
+A simplex for the Nelder-Mead algorithm.
+
+# Constructors
+
+    NelderMeadSimplex(M::AbstractManifold)
+
+Construct a simplex using random points from manifold `M`.
```
kellertuer, line 11:
```suggestion
Construct a  simplex using ``n+1`` random points from manifold `M`, where ``n`` is the manifold dimension of `M`.
```

src/solvers/NelderMead.jl
mateuszbaran: That would be fine if you could infer `P` from somewhere else but here `P` is inferred from either population or the initial point. We could validate that what we get is indeed a correct population but wrapping it in a simplex type seems easier to me.

kellertuer: no problem, let's wrap it - jus be careful to update the documentation and also add doc-strings to the `NelderMeadSimplex` constructor(s)-

kellertuer:
Hm, I am a little bit confused by the Documentation CI, which now reports that Manopt is missing?! Not sure what that means, since I mainly returned with a few notebooks back to the old style of not using Plutos package management but the environment on CI, which previously lists Manopt as installed.

Besides that on the Documentation side, the name changes have to be checked next carefully again in the docs (which is mainly a lot of reading and adapting names/notation in docstrings).

mateuszbaran:
Unfortunately, I can't really help with these Pluto problems :confused: .

I'm currently looking into updating my PCA code to cached objective and I'm not sure what is the intended way to use `SimpleCacheObjective`. It wraps an objective but for example in quasi-Newton the objective is constructed along the state, so I either need to copy a significant part of `quasi_Newton!` function or adjust that function to optionally wrap the objective in the cache. I'd also have to copy it if I wanted to use quasi-Newton optimizers with a combined cost-gradient objective. Is that intended?

mateuszbaran:
I see, no problem, I've switched to working on ManifoldDiff.jl for now (removing Manifolds.jl dependency) and I'll come back to this PR after that's done.

kellertuer:
Now we have (similar to `decorate_state!` to process `debug=`, `record=` and `return_state=` a `decorate_objective!` that processes `cache=` and currently can to `cache=:Simple` for our simple cache (the advanced one to pass down keywords is sketched already as well, but neither of them covered in tests).

kellertuer:
Meh. after another loooooong evening I am back to tutorials that work okay-ish locally (a few optical things not so nice) now even with a possible switch from checking for a development version or the last registered one...

But that fail on CI, now because it seems its evaluated out of order...

Sometimes I am really not happy I switched tutorials to Pluto – I spent so so so so so so much time debugging these that I could have written another complete documentation in that time - in Weave or what I used before. ☹️

kellertuer:
On 3 files in the docs we include 2 code files and for the second it seems some references/links within the page are broken – at least that is what I find as a common thing among the broken links. I am not yet sure why.

kellertuer:
So just the ManifoldDiff dependency is missing :)

mateuszbaran:
> Sometimes I am really not happy I switched tutorials to Pluto – I spent so so so so so so much time debugging these that I could have written another complete documentation in that time - in Weave or what I used before. frowning_face

I've tried Pluto for a a couple of hours in total and concluded that it's to much hassle for what it does so I definitely won't be against removing Pluto :wink: .

> On 3 files in the docs we include 2 code files and for the second it seems some references/links within the page are broken – at least that is what I find as a common thing among the broken links. I am not yet sure why.

Some broken links are sometimes hard to fix.

> So just the ManifoldDiff dependency is missing :)

Yes, it's mainly that. I'd also like to finish my PCA example to see that this rework works fine with it, in case it suggested some minor changes.

kellertuer:
I like Pluto in general, just within docs, especially with tutorials that should be up to date to the most recent version that is a hustle and the results don't look that nice.

For ManoptExamples I might use mainly Pluto Notebooks to illustrate how the examples work, since there its fine if that refers to an old version.

I fixed the docs and even see roughly how I could fix the remaining 13 lines of coverage – but I also would like to finish this somewhen.

Sure a test with the PCA would be great, as soon as the ManifoldDiff PR is merged we can check here what we can remove when we introduce the dependency.

kellertuer:
I started a-roping most of the code to ManifoldDiff – however my `diff_forward_logs`, which basically is the differential of foraward logs on an array of points now fails with one very strange number in its entries and I get one allocation error from ManifoldsDiff when calling `adjoint_Jacobi_field` from ManifoldDiff.

kellertuer:
Found my bug – now the only thing left is Jacobi fields for circle and such...

kellertuer:
For the remaining tests here to pass, we would need https://github.com/JuliaManifolds/ManifoldDiff.jl/issues/11, but the rest is done.

mateuszbaran:
Sure, I will fix nonmutating manifolds in a moment.

mateuszbaran:
Hm, it doesn't look like Jacobi fields actually work for PositiveNumbers. I will just fix Circle and Euclidean for now.

mateuszbaran:
That is, there is no diagonalizing basis for PositiveNumbers currently.

kellertuer:
Ok, that basis is a very boring one, but sure, my main application is actually the circle, though I covered `PositiveNumbers`and `Euclidean{Tuple{}}` in my non mutating cases as well.

mateuszbaran (approved):
I think caching requires some additional work but I don't really have enough time to work on it now. It will have to wait. In the meantime maybe it would be a good idea to add a note that it's an experimental feature for now?

PR #182, opened 2022-12-06
Small cleanup

mateuszbaran:
A small cleanup of a few things I've noticed while looking at Manopt recently. There should be nothing breaking or changing anything substantial.

### DIFF
--- a/src/plans/alm_plan.jl
+++ b/src/plans/alm_plan.jl
@@ -186,7 +186,7 @@ end
 # Allocating vector -> we can omit a few of the ineq gradients.
 function (
     LG::AugmentedLagrangianGrad{
-        <:ConstrainedProblem{<:AllocatingEvaluation,<:VectorConstraint}
+        <:ConstrainedProblem{AllocatingEvaluation,<:VectorConstraint}
     }
 )(
     ::AbstractManifold, X, p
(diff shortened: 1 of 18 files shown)


src/plans/alm_plan.jl
```diff
@@ -186,7 +186,7 @@ end
 # Allocating vector -> we can omit a few of the ineq gradients.
 function (
     LG::AugmentedLagrangianGrad{
-        <:ConstrainedProblem{<:AllocatingEvaluation,<:VectorConstraint}
+        <:ConstrainedProblem{AllocatingEvaluation,<:VectorConstraint}
```
kellertuer, line 189: Oh I am never sure about the `<:`such that here I thought – there is no harm in having that ;)

src/plans/gradient_plan.jl
```diff
@@ -56,15 +57,19 @@ function get_gradient(p::AbstractGradientProblem{AllocatingEvaluation}, x)
 end
 function get_gradient(p::AbstractGradientProblem{MutatingEvaluation}, x)
     X = zero_vector(p.M, x)
-    return p.gradient!!(p.M, X, x)
+    # note: p.gradient!! is user-supplied so let's not force on users that they must return X
```
kellertuer, line 60: ah yes I am switching to that mode in the 0.4 branch as well.

kellertuer (approved):
Thanks for finding these, no does not seem to be breaking, just a few merge conflicts when I merge that over to the rework-branch (where I finished gradient descent yesterday, and I like the new format).

PR #190, opened 2022-12-22
Don't allocate initial_operator for L-BFGS

mateuszbaran:
It can be very large for high-dimensional manifolds.


kellertuer (approved):
Very well spotted, thanks for that. Let‘s just wait for the test to pass

PR #193, opened 2023-01-06
Refactor `docs/make.jl`

rikhuijzer:
Because multiple `files`  are now passed to `PlutoStaticHTML.build_notebooks`, this PR makes sure that the notebooks are evaluated in parallel. This make the notebook building part roughly 2 times as fast in CI (probably less in practice due to overhead from distributed workers and the fact that `build_notebooks` is called twice). Locally you probably have more than 2 cores so the notebook building should be much faster there.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,94 +2,71 @@ using Documenter: DocMeta, HTML, MathJax3, deploydocs, makedocs
 using Manopt, Manifolds, Literate, Pluto, PlutoStaticHTML, Pkg
 # Load an unregistered package (for now) to update exports of Pluto notebooks
 
-tutorial_menu = Array{Pair{String,String},1}()
+struct Notebook
+    name::String
+    title::String
+end
 
-#
-# Generate Pluto Tutorial HTMLs
+function build(dir::String, notebooks::Vector{Notebook})
+    bopts = BuildOptions(
+        dir;
+        output_format=documenter_output,
+        write_files=true,
+        use_distributed=true
+    )
+    files = [nb.name * ".jl" for nb in notebooks]
+    build_notebooks(bopts, files)
+    return nothing
+end
+
+@info " \n      Rendering Tutorials\n "
+
 tutorial_src_folder = joinpath(@__DIR__, "..", "tutorials/")
 tutorial_output_folder = joinpath(@__DIR__, "src/", "tutorials/")
 tutorial_relative_path = "tutorials/"
 mkpath(tutorial_output_folder)
-#
-# Tutorials
-@info " \n      Rendering Tutorials\n "
-tutorials = [
-    Dict(:file => "Optimize!", :title => "Get Started: Optimize!"),
-    Dict(:file => "AutomaticDifferentiation", :title => "Use AD in Manopt"),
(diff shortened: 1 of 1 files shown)


kellertuer:
Interesting! I do not understand much of the changes – but thanks for the help :)
The examples will actually be removed in the next release (due to a separate package for them later), but that's easy to merge for sure.

Thanks for the help!

kellertuer:
@rikhuijzer I ran the docs locally (since for external PRs they do not run here) – results look very good, just the `Edit_URL` things are pointing to the wrong files now (the md files that are auto-generated, I tweaked those to link to the Notebooks).
Can that be done in the parallel setting as well? I think that was one of the reasons I stayed in the sequential setting.

kellertuer:
Aha! I got an idea. Would it be possible to do a similar postprocessing as we do with the contributing.md when copying that over?
That is: after rendering all tutorials (no need to do that for the examples) open all tutorials again and append the metadata in the beginning? That would be neat.

rikhuijzer:
> results look very good, just the `Edit_URL` things are pointing to the wrong files now

Oh yes of course that's because of the `mv`. Can the `mv` be avoided to make things simpler? 

Otherwise maybe a replace of the URL in the files although that sounds error prone.

> That is: after rendering all tutorials (no need to do that for the examples) open all tutorials again and append the metadata in the beginning? That would be neat.

I don't think that that is an option I would like to support. You currently already say that the `PlutoStaticHTML`  options are too complex.

kellertuer:
> > results look very good, just the `Edit_URL` things are pointing to the wrong files now
> 
> Oh yes of course that's because of the `mv`. Can the `mv` be avoided to make things simpler?

What are you referring to here? That the `.jl` files are in another place than the `.md` files? But the tutorial files are in the tutorials folder – which does not belong to the documentation, the `.md` files do, so they should be in the docs folder.

> Otherwise maybe a replace of the URL in the files although that sounds error prone.

Besides changing `md` to `jl` it is just a base folder change (I had that before and never troubles with that).

> > That is: after rendering all tutorials (no need to do that for the examples) open all tutorials again and append the metadata in the beginning? That would be neat.
> 
> I don't think that that is an option I would like to support. You currently already say that the `PlutoStaticHTML` options are too complex.

My main problem is maybe more understanding the options – for example `# hideall` does not hide all – it keeps output (sometimes?).

kellertuer:
> > That is: after rendering all tutorials (no need to do that for the examples) open all tutorials again and append the metadata in the beginning? That would be neat.
> 
> I don't think that that is an option I would like to support. You currently already say that the `PlutoStaticHTML` options are too complex.

On the point of supporting that option – I did not mean to imply that. Besides that, of course all design decisions on your package are your own :)

In these docs here I would like to have it – that is one could also post process that in the `make.jl` (Similar to what I do for the contributions page). If one does not set that link, originally the edit link would lead to a 404 page (since the PlutoStaticHTML-rendered-md files were not committed to the Repo). I changed that recently such that rendering is faster (at the cost of the md files being out of date), but even then the edit urls lead to files that should not be edited. The `.jl` files should be edited instead.

kellertuer:
Closed due to the work / transition started in #210.

PR #195, opened 2023-01-11
Introduce the difference of convex algorithms

kellertuer:
Introduce two algorithms for solving difference of convex algorithms:

* the DCA as described in https://arxiv.org/abs/2112.05250
* the DCPPA as described in https://doi.org/10.1007/s10589-020-00173-3

# To Do
* [x] write tests
  * [x] Basic tests done
  * [x] test closed form solution variants for both sub problems (dummy ones, since closed form solutions are not often available?)
  * [x] check sub gradient functions in the cross cases (in-place of AllocEval / alloc of InplEval)
  * [x] check that for a provided gradient the stopping criteria are correct.
* [x] check rendered docs.
* [x] update list of solvers.
* [x] solver state summaries


mateuszbaran:
Interesting methods. If these are only for Hadamard manifolds then maybe there should be a check for that?

kellertuer:
This is still a little bit work in progress, I am not sure whether they work more general, at least the first paper is called “on Riemannian Manifolds”, but I think they ended work only on Hadamard. Sure I will add a note probably (or try them on the sphere ;))

kellertuer:
IN principle this is finished, I just have to find out why Julia 1.6 on ubuntu does not terminate its test, will be hard to spot, since there is also no prints for now. But the rest should be done.

mateuszbaran:
I'd guess that Julia 1.6 has some random issue on Ubuntu, maybe it will fix itself tomorrow?

kellertuer:
I will try again tomorrow, but when I do not have a Ubuntu machine, I can not recreate this here locally (on Mac all works :) )

mateuszbaran:
Huh, so I have a Linux Mint (Ubuntu derivative) and decided to look what is going on. There is actually an infinite loop, right here (`linesearch_backtrack`):
```julia
    while fNew > f0 + decrease * s * search_dir_inner # decrease
        s = contract * s
        retract!(M, p_new, p, η, s, retr)
        fNew = f(p_new)
        (s < stop_step) && break
    end
```
It happens that after some iterations `s=4.4e-323`, `stop_step=0.0`, and `fNew` is barely larger than `f0 + decrease * s * search_dir_inner`, so the `while` condition remains true and the `break` condition remains false. I'd suggest adding a termination condition that is certain to happen. LineSearches.jl throws an error if 1000 iterations turn out to be insufficient. I guess there is also a fundamental error somewhere though, like descent direction not being a descent direction.

kellertuer:
Oh I thought I had switched the default to trust regions everywhere,  but yes that is what I wrote somewhere already, our Armijo (or linesearch backtrack in general) is a bit too academic at the moment, but I also want to do that careful, so that we do nut just introduce magic numbers.

kellertuer:
>  fundamental error somewhere though, like descent direction not being a descent direction.

I am quite sure that is not the case. I have been working with these things since January and fought a lot of fights against linesearch, it is our linesearch that is not that great, in its defaults in small tricks here and there... and also: the error only appears on 1/6 CIs – only Julia 1.6/ubuntu.

mateuszbaran:
This one was the first to loop infinitely: https://github.com/JuliaManifolds/Manopt.jl/pull/195/files#diff-1c4ed67c6a962227c4791cc22277282773b8bfd3dde0d933717ecfb7320fbf3fR100 , I didn't check the later ones.

Anyway, falling into an infinite loop there looks to me like a bad option. It takes some time to figure out what causes the infinite loop. Maybe you could just throw an error if there is more than some number of iterations until a better solution is found?

kellertuer:
Ah, I see, yes since I have a default _with_ an approximate hessian and one can deactivate that, that is the case where we actually use gradient descent for now.

As I said – yes that is something I aim for, but I am still not sure how to best realise this. And sure I can think about that, but maybe I would prefer to first finish this, otherwise this will be (again) an infinite PR, while a paper I am writing in this area should hopefully be finished in a finite amount of time. So I would prefer for now just to fix the test and leave a meter variant of linesearch/Armijo for an issue / a next PR.

kellertuer:
I think I found a good solution (not for Armijo but for the defaults here)

Before the last commit the defaults were
* a) if you kept the default `sub_hess` is a forward difference approximation for the hessian based on the gradient and a trust region solver was used
* b) if you set `sub_hess=nothing` by default a gradient descent was used.

this was a bit complicated to explain and since in practice TR is much more stable and such, I just deleted case b.
You can still use a gradient descent of course by setting that in `sub_state` but it is not a default any longer.
Since these were so many cases, this also simplified the code and the tests.

Issue #194, opened 2023-01-08
More persistent tutorials


kellertuer:
Currently there is two small drawbacks with the tutorials

* it is rather complicated to run them on CI for the documentation if the new version is breaking
* Also on CI the idea of using the local version from the repository seems to have stopped working and the caching works only sometimes in cases that are not yet clear - the docs just work because the cached versions are checked in.
* the rendering takes quite a long time
* the output can to be that well styled

So it would be a good idea to look for a new system.

It should provide a Markdown version (for the docs) and a Pluto version. For best the Markdown version refers to the Pluto version. It would be nice if both would stem from the same source. But a separate generation of the md (with example code and output) is also ok.
Goal: Persistent way to have synced Pluto & md files and a more stable and faster documentation generation.

Maybe [Quarto](https://quarto.org) would be a good alternative?

kellertuer:
Quarto seems to be a good idea, one could adapt the example from here

https://github.com/pat-alt/LaplaceRedux.jl/blob/main/docs/src/intro.qmd

for the tutorials in the docs – and see from there whether cmd can also be transformed (that is not even real rendering) into Pluto notebooks.

kellertuer:
resolved by #210 .

PR #196, opened 2023-01-13
Stepsize limiting on some manifolds

mateuszbaran:
Solves issue https://github.com/JuliaManifolds/Manopt.jl/discussions/191 . Also fixes cache exposing internal state when accessing gradient.

For now I've only applied the limiting step size to `WolfePowellLinesearch` but if you think this a good thing I can apply  this change to other methods as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.4.0"
+version = "0.4.1"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"
--- a/src/plans/cache.jl
+++ b/src/plans/cache.jl
@@ -70,7 +70,7 @@ function get_gradient(M::AbstractManifold, sco::SimpleCacheObjective, p)
         copyto!(M, sco.p, p)
         sco.X_valid = true
     end
-    return sco.X
+    return copy(M, p, sco.X)
 end
 function get_gradient!(M::AbstractManifold, X, sco::SimpleCacheObjective, p)
     scop_neq_p = sco.p != p
@@ -130,7 +130,7 @@ function get_gradient(
         sco.X_valid = true
         sco.c_valid = true
     end
-    return sco.X
+    return copy(M, p, sco.X)
 end
 function get_gradient(
     M::AbstractManifold,
--- a/src/plans/stepsize.jl
+++ b/src/plans/stepsize.jl
@@ -611,10 +611,10 @@ function (a::WolfePowellLinesearch)(
     max_step = max_stepsize(M, cur_p)
     # max_step_increase is the upper limit for s_plus
     max_step_increase = ifelse(isfinite(max_step), min(1e9, max_step / grad_norm), 1e9)
-    step = ifelse(isfinite(max_step), min(1.0, max_step / (2*grad_norm)), 1.0)
+    step = ifelse(isfinite(max_step), min(1.0, max_step / (2 * grad_norm)), 1.0)
     s_plus = step
     s_minus = step
-    
+
     f0 = get_cost(mp, cur_p)
     p_new = retract(M, cur_p, step * η, a.retraction_method)
     fNew = get_cost(mp, p_new)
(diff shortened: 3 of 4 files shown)


kellertuer:
The main other one (which only in rare cases even increases the step size) would be Armijo. Until now I did not encounter problems with Armijo concerning too large step sizes, though, it usually really only backtracks (i.e. reduces) the step size.

kellertuer (approved):

kellertuer:
Hm and Documenter has again some problem with PlutoStaticHTML, I will work on replacing that probably, I just have to find a good alternative, see #194.

mateuszbaran:
It primarily isn't about increasing stepsize but about the initial guess. 1 is sometimes way too much. Maybe my PCA example is somewhat atypical though.

kellertuer:
But then we could also set the initial guess to the minimum of 1 and the infectivity radius instead of introducing a max_step? Or maybe we could do both?

mateuszbaran:
At the moment `max_stepsize` is just calling `injectivity_radius`. Maybe there is no need for a new function for stepsize. This PR actually sets the initial stepsize to a smaller value for Wolfe-Powell when it looks like a good idea.

kellertuer:
Ah I missed that one, sure that is a good idea – and that part we could also do for Armijo, just to be safe, but the rest seems to much I think.

mateuszbaran:
Could you help me choose the right place to limit the initial guess in Armijo? This function: https://github.com/JuliaManifolds/Manopt.jl/blob/522668b7184693ce9cf5a5472281189c6ab738ad/src/plans/stepsize.jl#L194-L214 does some weird things with gradients. For example I don't understand what `get_gradient!(mp, get_gradient(s), get_iterate(s))` is supposed to do, which number should I limit and based on which tangent vector.

kellertuer:
That is because Armijo usually continues with the last stepsize, the initial one is here https://github.com/JuliaManifolds/Manopt.jl/blob/522668b7184693ce9cf5a5472281189c6ab738ad/src/plans/stepsize.jl#L176

kellertuer:
Ah and if you provide an initial guess function it gets ignored, but then we can not do anything, because if the user provides a function that computes the initial stepsize to test, its their responsibility.

mateuszbaran:
That's quite unfortunate because at that place we don't have the gradient yet. This limiting needs to happen after the gradient is calculated.

> Ah and if you provide an initial guess function it gets ignored, but then we can not do anything, because if the user provides a function that computes the initial stepsize to test, its their responsibility.

Yes, I wouldn't worry about that case.

kellertuer:
Well _after_ you computed the gradient, so in your position you have the `last_stepsize` field.
That is the time where the last step size is used as the start (if no function is provided). So what you could do is redefine the default of the initial guess?
Why do you need the gradient? Ah because you want to habe step * gradient to be in norm less than infectivity radius?

The guess I am referring to is
https://github.com/JuliaManifolds/Manopt.jl/blob/522668b7184693ce9cf5a5472281189c6ab738ad/src/plans/stepsize.jl#L181

where p, o, i, l (were not updated and) are the problem, the state, the iterate and the last step.

mateuszbaran:
> Why do you need the gradient? Ah because you want to habe step * gradient to be in norm less than infectivity radius?

Yes, exactly that.

> where p, o, i, l (were not updated and) are the problem, the state, the iterate and the last step.

So, initial_guess should be something like this?
```julia
function ig(p, s, i, l)
    M = get_manifold(p)
    X = get_gradient(s)
    p = get_iterate(s)
    grad_norm = norm(M, p, X)
    max_step = max_stepsize(M, p)
    return ifelse(isfinite(max_step), min(1.0, max_step / (2 * grad_norm)), 1.0)
end
```
Should `get_gradient!(mp, get_gradient(s), get_iterate(s))`  be executed before or after `initial_guess`?

kellertuer:
It should be executed before. We first need the gradient before we look for the step. So your variant is completely correct.

mateuszbaran:
OK, it seems to work -- so this PR should be complete now.

kellertuer (approved):
Looks good, thanks for the updates and fixes :)

mateuszbaran:
Should we wait until you resolve the Pluto problem with tagging a new version of Manopt?

kellertuer:
No, that takes too long still. I will not be able to Fix PlutoStaticHTML, since it is not reproducible enough to track that down. And rewriting all Tutorials (probably in Quarto) will still take some time.

It is now what it is. PlutoStaticHTML is sadly not so static, as the name suggests.

PR #198, opened 2023-01-15
Max stepsize on tangent bundle

mateuszbaran:
I've done this parallel to https://github.com/JuliaManifolds/Manifolds.jl/pull/567 because that change doesn't resolve the underlying issue of needing a good guess for maximum stepsize. This change should be reasonable and hopefully fix the tutorial docs.


kellertuer (approved):
😎

PR #200, opened 2023-01-16
Fixes grad_distance functions

hajg-ijk:
Fixes grad_distance functions to make them behave like explained in the docs.

### DIFF
--- a/src/functions/gradients.jl
+++ b/src/functions/gradients.jl
@@ -194,11 +194,23 @@ corresponding zero tangent vector, since this is an element of the subdifferenti
   distance
 """
 function grad_distance(M, y, x, p::Int=2)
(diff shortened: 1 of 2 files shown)


kellertuer:
Cool. Can you maybe add a test that checks that these new cases (we fixed) are actually doing the correct thing?

hajg-ijk:
I edited the test_gradients.jl file adding test cases for p=1 and x=y and x!=y both for the allocating and the mutating versions. Is this what you meant?

kellertuer:
Yes, that was what I was asking for – as you can see in the report in this PR, the code coverage now covers the new lines (all eight of them). Thanks for adding the tests :) That way we also make sure that we do not accidentally break that later.

kellertuer:
I will merge this already, it will be included in the next release then, but I want to wait for the other bug report just opened this morning and fix that as well in the next version (and also hope we have TagBot back until then).

PR #202, opened 2023-01-16
Allow the CG coefficient β to be complex.

### LINKED ISSUE #201: vector_transport_to function in the grassmann manifolds  doesn't work unless using it with ProjectionTransport()
Hi, I want to use CG-method in grassmann manifolds. but it cannot work, because of the vector_transport_to function of grassmann.
```
M = Grassmann(2, 2, ℂ)
p = rand(M)
X = rand(M; vector_at=p)
q = rand(M)
vector_transport_to(M, p, X, q)
```
error is below.
```
ERROR: MethodError: parallel_transport_to!(::AbstractManifold, ::Matrix{ComplexF64}, ::Matrix{ComplexF64}, ::Matrix{ComplexF64}, ::Matrix{ComplexF64}) is ambiguous. Candidates:
```
this called in  CG-methods (& the others) expects the parallel_transport_to!() for ::AbstractManifold or using it with ProjectionTransport()

kellertuer:
Fixes #201 (the part of that which is Manopt-related).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.4.2"
+version = "0.4.3"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"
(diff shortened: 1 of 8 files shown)


mateuszbaran:
I've added a simple test but it doesn't converge to the solution :disappointed: .

kellertuer:
Now this PR even directly resolves the same issue on the limited memory QN code.

kellertuer:
Not yet sure why adding a test reduced the code coverage and one method was not used anymore.

mateuszbaran (approved):
I think this works fine, and additional improvements to CG can be made separately.

PR #205, opened 2023-01-19
Introduce RestartCG as a decorator.

### LINKED ISSUE #203: Descent direction reset for CG
Some variants of CG have descent direction reset option when it diverges too much from gradient (for example `inner(gradient, descent_direction` is negative). It would be nice to add this option to Manopt.jl.

### LINKED ISSUE #204: Rename `HeestenesStiefelCoefficient` to `HestenesStiefelCoefficient`
It looks like a typo (two 'e' instead of one).

kellertuer:
This PR fixes #204 and fixes #203.

There is still a restart test missing, but I hope the formula and how it is implemented is correct. Will add the test the next days.

You can already have a look, maybe, @mateuszbaran :)


joss/paper.md
```diff
@@ -65,7 +65,7 @@ In the current version 0.3.17 of `Manopt.jl` the following algorithms are availa
...
+  [`SteepestDirectionUpdateRule`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.SteepestDirectionUpdateRule),   [`ConjugateDescentCoefficient`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.ConjugateDescentCoefficient). [`DaiYuanCoefficient`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.DaiYuanCoefficient), [`FletcherReevesCoefficient`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.FletcherReevesCoefficient), [`HagerZhangCoefficient`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.HagerZhangCoefficient), [`HestenesStiefelCoefficient`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.HestenesStiefelCoefficient), [`LiuStoreyCoefficient`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.LiuStoreyCoefficient), and [`PolakRibiereCoefficient`](https://manoptjl.org/v0.3/solvers/conjugate_gradient_descent.html#Manopt.PolakRibiereCoefficient)
```
mateuszbaran, line 68: This shouldn't have to be updated (it explicitly links to Manopt v0.3).

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,75 @@ function (u::SteepestDirectionUpdateRule)(
 )
     return 0.0
 end
+
+@doc raw"""
+    ConjugateGradientRestart <: DirectionUpdateRule
+
+An update rule might require a restart, that is one gradient step, if the last two gradients
```
mateuszbaran, line 515:
```suggestion
An update rule might require a restart, that is using pure gradient as descent direction, if the last two gradients
```
I think this explanation is more clear.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,75 @@ function (u::SteepestDirectionUpdateRule)(
...
+    DUR<:DirectionUpdateRule,VT<:AbstractVectorTransportMethod,F
+} <: DirectionUpdateRule
+    direction_update::DUR
+    storage::StoreStateAction
+    threshold::F
+    vector_transport_method::VT
+    function ConjugateGradientRestart(
+        direction_update::D,
+        threshold=Inf;
+        manifold=DefaultManifold(),
```
mateuszbaran, line 555:
```suggestion
        manifold::AbstractManifold=DefaultManifold(),
```

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,75 @@ function (u::SteepestDirectionUpdateRule)(
...
+```
+where ``P_{a\gets b}(⋅)`` denotes a vector transport from the tangent space at ``a`` to ``b``,
+and ``ξ`` is the `threshold`.
+
+# Constructor
+
+    PolakRibiereCoefficient(
+        direction_update::D,
+        threshold=Inf;
+        manifold = DefaultManifold(),
```
mateuszbaran, line 535:
```suggestion
        manifold::AbstractManifold = DefaultManifold(),
```

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,75 @@ function (u::SteepestDirectionUpdateRule)(
 )
     return 0.0
 end
+
+@doc raw"""
+    ConjugateGradientRestart <: DirectionUpdateRule
+
+An update rule might require a restart, that is one gradient step, if the last two gradients
+are nearly orthogonal, cf. [^HagerZhang2006], page 12 (in the pdf, 46 in Journal page numbers).
```
mateuszbaran, line 516: This paper isn't linked anywhere (though there is `HagerZhang2005`).

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,75 @@ function (u::SteepestDirectionUpdateRule)(
...
+        manifold=DefaultManifold(),
+        vector_transport_method::V=default_vector_transport_method(manifold),
+        a::StoreStateAction=StoreStateAction((:Iterate, :gradient, :δ)),
+    ) where {D<:DirectionUpdateRule,V<:AbstractVectorTransportMethod}
+        return new{D,V,typeof(threshold)}(
+            direction_update, a, threshold, vector_transport_method
+        )
+    end
+end
+function (u::ConjugateGradientRestart)(
```
mateuszbaran, line 564: It's a bit sad that this method has to repeat so much work that is already performed in the wrapped update rule.

src/plans/conjugate_gradient_plan.jl
mateuszbaran: By the way, I've also seen restarting when beta gets negative, see here for example: https://en.wikipedia.org/wiki/Nonlinear_conjugate_gradient_method .

joss/paper.md
kellertuer: Ah, yes, you are right, since it links to 0.3.

src/plans/conjugate_gradient_plan.jl
kellertuer: the PR-CG includes that, but currently it is not a decorator. But we could add that as an option as well.

src/plans/conjugate_gradient_plan.jl
kellertuer: HagerZhang2005 is the HZ-Coeff paper linked in the HZ docs. The 2006 one is given in the same string below and is an overview article.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,75 @@ function (u::SteepestDirectionUpdateRule)(
...
+mutable struct ConjugateGradientRestart{
+    DUR<:DirectionUpdateRule,VT<:AbstractVectorTransportMethod,F
+} <: DirectionUpdateRule
+    direction_update::DUR
+    storage::StoreStateAction
+    threshold::F
+    vector_transport_method::VT
+    function ConjugateGradientRestart(
+        direction_update::D,
+        threshold=Inf;
```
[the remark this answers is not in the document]
kellertuer: My problem here was, that the two values I had in mind are not that reasonable (Inf and 0 – never and always restart), but also that there is no good default in between - or are you aware of one?

src/plans/conjugate_gradient_plan.jl
kellertuer: Even more, the internal one might (somehow) work without storage, then the storage here is really required.

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Right, I didn't notice.

kellertuer: ...I actulally also checked 3 times that they were not the same year and not the same ref. ;)

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Sure, not every update rule has to maintain storage and do the vector transports but some do, and then the work is done twice. But it can't be generically simplified.

src/plans/conjugate_gradient_plan.jl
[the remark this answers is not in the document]
mateuszbaran: This paper: https://link.springer.com/article/10.1007/BF01593790 (Section 4) suggests using value 0.2.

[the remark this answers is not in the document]
kellertuer: Cool, I added the default and the reference as well.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,81 @@ function (u::SteepestDirectionUpdateRule)(
...
+```math
+    \frac{ ⟨X_{k+1}, P_{p_{k+1}\gets p_k}X_k⟩}{\lVert X_k \rVert_{p_k}} > ξ,
+```
+where ``P_{a\gets b}(⋅)`` denotes a vector transport from the tangent space at ``a`` to ``b``,
+and ``ξ`` is the `threshold`.
+The default threshold is chosen as `0.2` as recommended in `
+
+# Constructor
+
+    PolakRibiereCoefficient(
```
mateuszbaran, line 533:
```suggestion
    ConjugateGradientRestart(
```

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,81 @@ function (u::SteepestDirectionUpdateRule)(
...
+    > W. W. Hager and H. Zhang, A Survey of Nonlinear Conjugate Gradient Methods
+    > Pacific Journal of Optimization 2, 2006, pp. 35-58.
+    > url: [http://www.yokohamapublishers.jp/online2/pjov2-1.html](http://www.yokohamapublishers.jp/online2/pjov2-1.html)
+
+[^Powell1977]:
+    > M.J.D. Powell: Restart Procedures for the Conjugate Gradient Method,
+    > Methematical Programming 12, 1977, pp. 241–254
+    > doi: [10.1007/BF01593790](https://doi.org/10.1007/BF01593790)
+"""
+mutable struct ConjugateGradientRestart{
```
mateuszbaran, line 551: Since there are many different restart strategies and this one is very specific, maybe let's call it `ConjugateGradientRestartBeale` (after the name from Powell1977)?

kellertuer: why `Beale`?

mateuszbaran: That's what Powell calls this method.

kellertuer: Ah I see, but then we should also cite Beale ;)

mateuszbaran: Sure, why not :slightly_smiling_face: .

kellertuer: Will try to find the exact reference then later today.

mateuszbaran: Powell refers to this paper:
E.M.L. Beale, "A derivation of conjugate gradients", in: F.A. Lootsma, ed., Numerical methods for nonlinear optimization (Academic Press, London, 1972) pp. 39-43.

kellertuer: ...yes, but does it have a DOI? Or at least a URL? Will check.

kellertuer: Could not find a URL but at least added an ISBN.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -507,3 +508,87 @@ function (u::SteepestDirectionUpdateRule)(
 )
     return 0.0
 end
+
+@doc raw"""
+    ConjugateGradientBealeRestart <: DirectionUpdateRule
+
+An update rule might require a restart, that is using pure gradient as descent direction,
+if the last two gradients are nearly orthogonal, cf. [^HagerZhang2006], page 12 (in the pdf, 46 in Journal page numbers).
+this method is names after E. Beale [^Beale1972].
```
mateuszbaran, line 517:
```suggestion
This method is named after E. Beale [^Beale1972].
```

mateuszbaran (approved):

PR #208, opened 2023-01-21
Use the new default_...(M, typeof(p)) where possible.

kellertuer:
This makes it easier using defaults on manifolds with different representations


kellertuer:
Hm, quite interesting that the tests pass locally – I seem to be quite in a bad luck with CI this year until now. Still waiting for a first PR/commit that directly passes :(

PR #209, opened 2023-01-22
Faster storage handling

mateuszbaran:
This takes runtime of the Rosenbrock example from https://github.com/JuliaManifolds/Manifolds.jl/issues/569 from about 80 us to 30 us. A few points:
1. Some places use `:gradient`, while other use `:Gradient`. I think this should be unified because I couldn't figure out which should be used where.
2. This kind of broadcast: `p_old, X_old, δ_old = get_storage.(Ref(u.storage), [:Iterate, :gradient, :δ])` is quite slow at the timescale of Rosenbrock and has a perfectly fine alternative :wink: .
3. I'm not sure why you avoided `a.values[key] = new_value` but it works and is faster than merging a new dictionary.
4. It would be nice to reduce the usage of `deepcopy` in storage code but it's a fairly ambitious goal and not a big issue.
5. The old code missed one `deepcopy` so this PR is likely also a bugfix.

### DIFF
--- a/src/plans/conjugate_gradient_plan.jl
+++ b/src/plans/conjugate_gradient_plan.jl
@@ -175,7 +175,9 @@ function (u::DaiYuanCoefficient)(
         update_storage!(u.storage, cgs) # if not given store current as old
         return 0.0
     end
-    p_old, X_old, δ_old = get_storage.(Ref(u.storage), [:Iterate, :gradient, :δ])
+    p_old = get_storage(u.storage, :Iterate)
+    X_old = get_storage(u.storage, :gradient)
+    δ_old = get_storage(u.storage, :δ)
     update_storage!(u.storage, cgs)
 
     gradienttr = vector_transport_to(M, p_old, X_old, cgs.p, u.transport_method)
@@ -221,10 +223,11 @@ function (u::FletcherReevesCoefficient)(
     amp::AbstractManoptProblem, cgs::ConjugateGradientDescentState, i
 )
     M = get_manifold(amp)
-    if !all(has_storage.(Ref(u.storage), [:Iterate, :gradient]))
+    if !has_storage(u.storage, :Iterate) || !has_storage(u.storage, :gradient)
         update_storage!(u.storage, cgs) # if not given store current as old
     end
-    p_old, X_old = get_storage.(Ref(u.storage), [:Iterate, :gradient])
+    p_old = get_storage(u.storage, :Iterate)
+    X_old = get_storage(u.storage, :gradient)
     update_storage!(u.storage, cgs)
     return inner(M, cgs.p, cgs.X, cgs.X) / inner(M, p_old, X_old, X_old)
 end
@@ -284,7 +287,9 @@ function (u::HagerZhangCoefficient)(
         update_storage!(u.storage, cgs) # if not given store current as old
         return 0.0
     end
-    p_old, X_old, δ_old = get_storage.(Ref(u.storage), [:Iterate, :gradient, :δ])
+    p_old = get_storage(u.storage, :Iterate)
+    X_old = get_storage(u.storage, :gradient)
+    δ_old = get_storage(u.storage, :δ)
     update_storage!(u.storage, cgs)
 
     gradienttr = vector_transport_to(M, p_old, X_old, cgs.p, u.transport_method)
@@ -355,7 +360,9 @@ function (u::HestenesStiefelCoefficient)(
         update_storage!(u.storage, cgs) # if not given store current as old
         return 0.0
     end
-    p_old, X_old, δ_old = get_storage.(Ref(u.storage), [:Iterate, :gradient, :δ])
+    p_old = get_storage(u.storage, :Iterate)
+    X_old = get_storage(u.storage, :gradient)
+    δ_old = get_storage(u.storage, :δ)
     update_storage!(u.storage, cgs)
     gradienttr = vector_transport_to(M, p_old, X_old, cgs.p, u.transport_method)
     δtr = vector_transport_to(M, p_old, δ_old, cgs.p, u.transport_method)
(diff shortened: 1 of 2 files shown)


kellertuer:
Thanks for this work.
Yes some motivation to storage should be added somewhere in the docs I think

1. For the naming `:Gradient` should refer to storing an evaluation of the gradient `:gradient` to the old) field gradient in the (by then) `Options`. While `:gradient` should now be `:X`, I think in most cases it should be indeed the (semantic) one: `:Gradient`. The idea is that the `get_...` functions have capital letter words, all low letter ones are mapped to fields in the solver state
2.  I was not aware of this - thanks for fixing that
3. I do not follow what you mean here can you give an example where I merge and do something wrong/inefficient?
4. sure, if possible we should avoid it, if we know more we should use `copy(M,...)`
5. oh, maybe, sure

The problem with the storage for me is,

1. I designed that in the very beginning of the package to keep the solver state clean and not clustered with old values stored, just in case some sub function might maybe need it. That is done locally while still possibly having a common storage.
2. I am not that good in general in efficient code, so any recommendation is welcome to improve this.

kellertuer:
hm, I think my idea in 3 with the merge is, that there might be cases, where the key does not yet exist in the dictionary, then the merge creates it (but again might not be the most efficient way of doing it) and was my idea back in 2015 – so sure there are probably more efficient ways of doing this.

mateuszbaran:
> 1. For the naming `:Gradient` should refer to storing an evaluation of the gradient `:gradient` to the old) field gradient in the (by then) `Options`. While `:gradient` should now be `:X`, I think in most cases it should be indeed the (semantic) one: `:Gradient`. The idea is that the `get_...` functions have capital letter words, all low letter ones are mapped to fields in the solver state

So, for example, should `update_storage!(a::AbstractStateAction, s::AbstractManoptSolverState)` use `:gradient` or `:Gradient`?

> 3\. I do not follow what you mean here can you give an example where I merge and do something wrong/inefficient?

> hm, I think my idea in 3 with the merge is, that there might be cases, where the key does not yet exist in the dictionary, then the merge creates it (but again might not be the most efficient way of doing it) and was my idea back in 2015 – so sure there are probably more efficient ways of doing this.
> 
> _edit:_ And I also remember why I did that: Imagine a user creates a storage – maybe for the stepsize – and does not have the CG computation in mind, so it is missing the delta. Then the old variant would just extend the storage - your code would error, since in your variant you assume an entry for key exists.

Did you try assigning something to a key in `Dict` that does not exist yet? It works for me. Maybe it didn't work in 2015 but it does now :slightly_smiling_face: .
```julia
julia> d = Dict()
Dict{Any, Any}()

julia> d[:my_key] = 12
12

julia> d[:my_key]
12
```
The slow merge (not wrong, just slow) was in `update_storage!(a::AbstractStateAction, s::AbstractManoptSolverState)`.

> The problem with the storage for me is,
> 
> 1. I designed that in the very beginning of the package to keep the solver state clean and not clustered with old values stored, just in case some sub function might maybe need it. That is done locally while still possibly having a common storage.
> 2. I am not that good in general in efficient code, so any recommendation is welcome to improve this.

I see. The idea of storing old values separately is fine, just the implementation needs some improvements.

kellertuer:
> > 1. For the naming `:Gradient` should refer to storing an evaluation of the gradient `:gradient` to the old) field gradient in the (by then) `Options`. While `:gradient` should now be `:X`, I think in most cases it should be indeed the (semantic) one: `:Gradient`. The idea is that the `get_...` functions have capital letter words, all low letter ones are mapped to fields in the solver state
> 
> So, for example, should `update_storage!(a::AbstractStateAction, s::AbstractManoptSolverState)` use `:gradient` or `:Gradient`?

`:Gradient` as `:gradient` would refer to `s.gradient` which does not exist. I would say all `:gradient` occurrences in Manopt 0.4 should be `:Gradient` ones (since that abstracts from the field the gradient is actually stored in). Especially for abstract cases. In concrete cases, i.e. when you really habe a concrete subtype you might also directly access a field (`:p` instead of `:Iterate` or if you e.g. use `:𝛿` in CG steps).

> Did you try assigning something to a key in `Dict` that does not exist yet? It works for me. Maybe it didn't work in 2015 but it does now 🙂 .
> 
> ```julia
> julia> d = Dict()
> Dict{Any, Any}()
> 
> julia> d[:my_key] = 12
> 12
> 
> julia> d[:my_key]
> 12
> ```
> 
> The slow merge (not wrong, just slow) was in `update_storage!(a::AbstractStateAction, s::AbstractManoptSolverState)`.

As I said, my implementations are often slow, because I do often not know the fast tricks (though I studies computer science, maybe I am too “mathy” for that these days?) and value any input here. Maybe that did not exist, I am not sure, we should then just make sure it exists in the lowest Julia version we support still (1.6 probably).

> > The problem with the storage for me is,
> > 
> > 1. I designed that in the very beginning of the package to keep the solver state clean and not clustered with old values stored, just in case some sub function might maybe need it. That is done locally while still possibly having a common storage.
> > 2. I am not that good in general in efficient code, so any recommendation is welcome to improve this.
> 
> I see. The idea of storing old values separately is fine, just the implementation needs some improvements.

Cool. And thanks for the ideas improving them :) One could also check to automatically reuse storage that is created (so that maybe RecordChange and DebugChange and a CG computation “automatically” use the same storage and we spare multiple copies).

mateuszbaran:
> As I said, my implementations are often slow, because I do often not know the fast tricks (though I studies computer science, maybe I am too “mathy” for that these days?) and value any input here. Maybe that did not exist, I am not sure, we should then just make sure it exists in the lowest Julia version we support still (1.6 probably).

That's not a problem, I can help make things faster :slightly_smiling_face: . Assigning to keys that don't exist works in Julia 1.6 so we can depend on that.

> Cool. And thanks for the ideas improving them :) One could also check to automatically reuse storage that is created (so that maybe RecordChange and DebugChange and a CG computation “automatically” use the same storage and we spare multiple copies).

That may be complicated to get right and is not a big issue so we can leave it for later.

kellertuer:
Cool, then lets switch to assigning keys

And sure the other idea can be postponed for late and affects usually only the cases where one records/debug anyways, so maybe that is not the most performance critical cases.

kellertuer:
I unified the usage to `:Gradient` for the `get_gradient` function. This seems to have been mixed up when moving from `:gradient` (the field in the options) to the accessor function (which still in most cases accesses `s.X` but it provides the freedom that the gradient might be somewhere else and maybe not explicitly stored).

kellertuer:
...we could even check the gradient and stochastic gradient updates (ADAM, Nesterov and such) whether they can be optimised the same way :)

mateuszbaran:
Yes, I've looked for similar patterns and improved most of them.

mateuszbaran:
Yes, sure, I was going to work a bit more on improving performance but I can do that in a new PR.

kellertuer:
This one is not urgent, we can also do that here :)

mateuszbaran:
I've already bumped the version so we can merge this and I will continue in a new PR :)

PR #210, opened 2023-01-23
Move Tutorials to be written in Quarto

kellertuer:
This PR (WIP) works on transferring the tutorials to Quarto

* [x] Rewrite tutorials
* [x] Check whether Quarto can run (including freeze / cache) on CI


kellertuer:
Just checked CI with Quarto and I do not directly see how to best resolve all dependencies to render/evaluate Julia code.

kellertuer:
I checked again for eventually rendering tutorials on CI, but the dependencies are a bit tricky, like Quarto, Python, Jupyter. but also Asymptote. So for now the tutorials should be locally generated and checked in. This was already the case for the previous tutorials anyways.

kellertuer:
Hm, seems like since yesterday we are back to one randomly failing test :( Locally on 1.8 the test runs fine but sometimes it fails on Windows.

PR #211, opened 2023-01-27
State Summary (or: `show()` methods for everyone!)

kellertuer:
I had an idea how to introduce more informative information from a solver run – when one looks at the state (i.e. uses `return_state=true`).
This mainly involves writing quite a few `show()` methods.

Here is three examples

## 1 Just looking at a state

```Julia
using Manifolds, Manopt
gst = GradientDescentState(Euclidean(2), [0.0,0.0]; stopping_criterion=StopAfterIteration(20), stepsize=ConstantStepsize(1.0))
```
prints
```JuliaREPL
# Solver state for gradient_descent

## Parameters
* retraction method: ExponentialRetraction()

## Stepsize
ConstantStepsize{Float64}(1.0)

## Stopping Criterion
Max Iteration 20:       not reached
This indicates convergence: No
```

## 2 A solver run that was succesfull

```Julia
using Manifolds, Manopt
p = [4.0, 2.0]
f(M, q) = distance(M, q, p) .^ 2
grad_f(M, q) = -2 * log(M, q, p)
s = gradient_descent(Euclidean(2), f, grad_f, zero(p); return_state=true)
```

prints

```JuliaREPL
# Solver state for gradient_descent
After 70 iterations

## Parameters
* retraction method: ExponentialRetraction()

## Stepsize
ArmijoLinesearch{ExponentialRetraction, typeof(Manopt.armijo_initial_guess)}(1.0, ExponentialRetraction(), 0.95, 0.1, 0.8573749999999999, 0.0, Manopt.armijo_initial_guess)

## Stopping Criterion
Stop When _one_ of the following are fulfilled:
    Max Iteration 200:  not reached
    |Δf| < 1.0e-9: reached
Overall: reached
This indicates convergence: Yes
```

## 3 A solver run that was not sucessful

```julia
s2 = gradient_descent(Euclidean(2), f, grad_f, zero(p); return_state=true, stopping_criterion=StopAfterIteration(2))
```

which prints

```
# Solver state for gradient_descent
After 2 iterations

## Parameters
* retraction method: ExponentialRetraction()

## Stepsize
ArmijoLinesearch{ExponentialRetraction, typeof(Manopt.armijo_initial_guess)}(1.0, ExponentialRetraction(), 0.95, 0.1, 0.8573749999999999, 0.0, Manopt.armijo_initial_guess)

## Stopping Criterion
Max Iteration 2:        reached
This indicates convergence: No
````

# ToDo
* [x] show-methods for stopping criteria
[...]

PR #212, opened 2023-01-31
Even faster storage handling

mateuszbaran:
These are the changes we've discussed yesterday. Right now a bit WIP still but with the fused retractions my CG example is down to about 6 us (from about 80 us IIRC) and looks relatively solid now.


src/plans/conjugate_gradient_plan.jl
```diff
@@ -157,36 +174,36 @@ default vector transport and a new storage is created by default.
...
-               DirectionUpdateRule
+struct DaiYuanCoefficient{TVTM<:AbstractVectorTransportMethod} <: DirectionUpdateRule
     transport_method::TVTM
-    storage::StoreStateAction
-    function DaiYuanCoefficient(
-        t::AbstractVectorTransportMethod=ParallelTransport(),
-        a::StoreStateAction=StoreStateAction((:Iterate, :Gradient, :δ)),
-    )
-        return new{typeof(t)}(t, a)
+    function DaiYuanCoefficient(t::AbstractVectorTransportMethod=ParallelTransport())
```
kellertuer, line 179: This one we could change to having the manifold as an optional (positional) parameter and the transport as a keyword (unified with all other places where we specify a transport) and then use the default

so
```
DaiYuanCoefficient(M::AbstractManifold=DefaultManifood(2); vector_transport_method = default_vector_transport(M))
```

because that way just providing the manifold „improves“ the default and if you want to be more precise you can also just specify the actual transport you want to use. I introduced this in a few other places as well and like this as a default scheme actually

src/plans/conjugate_gradient_plan.jl
```diff
@@ -1,3 +1,22 @@
+
+struct DirectionUpdateRuleStorage{TC<:DirectionUpdateRule,TStorage<:StoreStateAction}
+    coefficient::TC
+    storage::TStorage
+end
+
+function DirectionUpdateRuleStorage(
+    M::AbstractManifold, dur::DirectionUpdateRule, p0=rand(M), X0=zero_vector(M, p0)
```
kellertuer, line 8: We could also just use p and X here

src/plans/conjugate_gradient_plan.jl
```diff
@@ -67,7 +86,7 @@ function ConjugateGradientDescentState(
     p::P,
     sC::StoppingCriterion,
     s::Stepsize,
-    dU::DirectionUpdateRule,
+    dU::DirectionUpdateRuleStorage,
```
kellertuer, line 89: since you have functions that automatically „know“ the storage requirements, we could keep the constructor nonbreaking and internally create the new storage rule?

src/plans/conjugate_gradient_plan.jl
```diff
@@ -274,32 +289,31 @@ default vector transport and a new storage is created by default.
 mutable struct HagerZhangCoefficient{TVTM<:AbstractVectorTransportMethod} <:
                DirectionUpdateRule
     transport_method::TVTM
-    storage::StoreStateAction
-    function HagerZhangCoefficient(
-        t::AbstractVectorTransportMethod=ParallelTransport(),
-        a::StoreStateAction=StoreStateAction((:Iterate, :Gradient, :δ)),
-    )
-        return new{typeof(t)}(t, a)
+    function HagerZhangCoefficient(t::AbstractVectorTransportMethod=ParallelTransport())
```
kellertuer, line 292: Same as for DYCoeff, lets use an optional manifold here.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -346,35 +362,38 @@ See also [`conjugate_gradient_descent`](@ref)
...
-mutable struct HestenesStiefelCoefficient{TVTM<:AbstractVectorTransportMethod} <:
-               DirectionUpdateRule
+struct HestenesStiefelCoefficient{TVTM<:AbstractVectorTransportMethod} <:
+       DirectionUpdateRule
     transport_method::TVTM
-    storage::StoreStateAction
     function HestenesStiefelCoefficient(
-        transport_method::AbstractVectorTransportMethod=ParallelTransport(),
-        storage_action::StoreStateAction=StoreStateAction((:Iterate, :Gradient, :δ)),
+        transport_method::AbstractVectorTransportMethod=ParallelTransport()
```
kellertuer, line 369: The same as for the other coifs with a transport.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -414,33 +433,33 @@ default vector transport and a new storage is created by default.
...
-               DirectionUpdateRule
+struct LiuStoreyCoefficient{TVTM<:AbstractVectorTransportMethod} <: DirectionUpdateRule
     transport_method::TVTM
-    storage::StoreStateAction
-    function LiuStoreyCoefficient(
-        t::AbstractVectorTransportMethod=ParallelTransport(),
-        a::StoreStateAction=StoreStateAction((:Iterate, :Gradient, :δ)),
-    )
-        return new{typeof(t)}(t, a)
+    function LiuStoreyCoefficient(t::AbstractVectorTransportMethod=ParallelTransport())
```
kellertuer, line 438: same as for the other transports.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -485,31 +504,30 @@ See also [`conjugate_gradient_descent`](@ref)
     > USSR Comp. Math. Math. Phys., 9 (1969), pp. 94–112.
     > doi: [10.1016/0041-5553(69)90035-4](https://doi.org/10.1016/0041-5553(69)90035-4)
 """
-mutable struct PolakRibiereCoefficient{TVTM<:AbstractVectorTransportMethod} <:
-               DirectionUpdateRule
+struct PolakRibiereCoefficient{TVTM<:AbstractVectorTransportMethod} <: DirectionUpdateRule
```
kellertuer, line 507: change constructor here as well.

src/plans/solver_state.jl
```diff
@@ -211,58 +211,104 @@ Initialize the Functor to a set of keys, where the dictionary is initialized to
 be empty. Further, `once` determines whether more that one update per iteration
 are effective, otherwise only the first update is stored, all others are ignored.
 """
-mutable struct StoreStateAction <: AbstractStateAction
-    values::Dict{Symbol,<:Any}
-    keys::NTuple{N,Symbol} where {N}
+mutable struct StoreStateAction{TPS,TXS,TPI,TTI} <: AbstractStateAction
+    values::Dict{Symbol,Any}
+    keys::Vector{Symbol} # for values
```
kellertuer, line 216: nice idea to map them to fields of the state :) Not yet sure why a vector is enough, will have to check the code closer.

src/plans/solver_state.jl
```diff
@@ -211,58 +211,104 @@ Initialize the Functor to a set of keys, where the dictionary is initialized to
...
+    point_values::TPS
+    tangent_values::TXS
+    point_init::TPI
```
kellertuer, line 219: what is the init used for? Below it is accessed at some key? This should be documented maybe.

src/solvers/conjugate_gradient_descent.jl
```diff
@@ -107,7 +107,7 @@ function conjugate_gradient_descent!(
         p,
         stopping_criterion,
         stepsize,
-        coefficient,
+        DirectionUpdateRuleStorage(M, coefficient, p, X),
```
kellertuer, line 110: as written above in the CG constructor - this could also be done in the CG constructor maybe, because then one does not have to change all the tests below :)

kellertuer (commented):
This looks super!

I notier after our discussion that this way the CG coefficients are even more similar to the QuasiNewton rules we have, which I think is super nice.

I just have a few first comments, but mainly usability ideas.

src/plans/conjugate_gradient_plan.jl
kellertuer: Maybe even a stronger argument here is that the coefficient can not be used alone _and_ the storage is an internal thing we do, so we should keep that also more internal.

kellertuer:
Oh, and one thing that I did notice, the storage until now was also used in DebugActions if I remember correctly so is it possible to keep the storage that general? If it is just a storage for CG coefficient-usage then we can also store 2 variables in the coefficients, since we will always only have one coefficients and there is no need for a shared storage. But maybe this is also just a Naming-Thing for the new storage and it can already be used in the other places?

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Sure, this makes sense. Just instead of `DefaultManifood` I will use `DefaultManifold` :wink: .

src/plans/solver_state.jl
mateuszbaran: Hm, I thought that was your idea? I've just changed it from tuple to vector to make it type stable and easier to use :slightly_smiling_face: .

src/plans/solver_state.jl
mateuszbaran: This is here as a substitute to `has_storage` from the old solution, i.e. tells if a particular storage key has actually been saved already. I will add some docs later.

src/plans/conjugate_gradient_plan.jl
kellertuer: Yes, this was written shortly before my breakfast ;)

src/plans/solver_state.jl
kellertuer: I am not 100% sure my implementation was the best achievable one. since the symbols are anyways stored in `keys(values)` I am not sure we actually need `keys` here?

kellertuer: (which is more reflecting on my choices back long long time ago then your current rework, sure).

mateuszbaran:
I've made the storage changes reusable in other parts such as debug actions, so feel free to use these capabilities elsewhere as well. Just storing two or three variables in coefficient storage would be much easier to code but IIRC you wanted to avoid that.

src/plans/solver_state.jl
kellertuer: looking forward to the docs then, since for now I do not understand what this field Is for.

kellertuer:
Maybe I am mainly arguing about the name the storage then currently has? For now it implements the general (i.e. hopefully reusable) case but is called `directionUpdateStorage` which (in name at least, not sure for the rest) seems to restrict the storage to CG and maybe average/Nesterov

src/plans/solver_state.jl
mateuszbaran: Separate keys are useful for differentiating "things that are stored" from "things we want to store but we don't yet" so I've left them separate. Basically what my `init` fields are doing for points and tangent vectors.

kellertuer: Ah! Then I also understand your `Init`s (and was maybe confused that I could not see their types).

mateuszbaran:
Yes, `DirectionUpdateStorage` is CG-specific but holds very little logic in itself. All new reusable parts are in `StoreStateAction`.

mateuszbaran:
Maybe we could extend `DirectionUpdateStorage` to something else but I haven't checked that yet. Anyway that looks like an internal change that could also be done later.

kellertuer:
> Yes, `DirectionUpdateStorage` is CG-specific but holds very little logic in itself. All new reusable parts are in `StoreStateAction`.

Ah, now I see, then maybe I misunderstood the code this morning. You are right, the DUS is just a nice encapsulation to keep the “oldCG-types” simple (and now even without own storage) and the main change is in the storage. I somehow read that differently this morning. Then all is fine :)

mateuszbaran:
> Ah, now I see, then maybe I misunderstood the code this morning. You are right, the DUS is just a nice encapsulation to keep the “oldCG-types” simple (and now even without own storage) and the main change is in the storage. I somehow read that differently this morning. Then all is fine :)

Yes, this is correct :+1: .

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Good idea, I've reverted this change.

src/plans/solver_state.jl
mateuszbaran: I've added some docs, let me know if they are sufficient.

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Done :slightly_smiling_face:

mateuszbaran:
I think this is more or less ready. This work could be continued for other parts of Manopt but for this PR I think these changes are more or less sufficient.

My last measurement of Rosenbrock benchmark is below 6 us and about half of all time is performing some actual computations. There is still at least one significant avoidable allocation (in line search) and some potentially avoidable copying but we are entering diminishing returns territory here. I'm quite happy with the overall improvement.

BTW, I've quickly checked peformence of quasi-Newton for Rosenbrock and the largest issue is too many allocations. There should be buffers that persist between iterations instead of constantly allocating for everything.

src/plans/solver_state.jl
kellertuer: HM, I still do not understand, what I should pass to `point_values` and what to `point_inits` - both are named tuples but of what? I mean it seems this PR leaves another (huge?) PR to update all the rest – and for now I do not understand how to use this, so that makes a huge PR even more work.

So – what are the values what are the inits? Are they keys? Could they then be named keys? If none of them are keys, how to I find out which value is stored where? For example if two points are stored, which one is which field from state?

In short: Yes, all fields have a documentation, but they read like the comment `#this is a for loop` in front of a for loop for me, sorry :/

kellertuer: also: tangent values and point values are documented twice, but different in the list.

kellertuer:
> I think this is more or less ready. This work could be continued for other parts of Manopt but for this PR I think these changes are more or less sufficient.

That reads for me a bit like this is half a PR, since it leave the rest – probably to me – somewhen after I both have understood how this works and I find time to work through all other storages.
I would prefer to be sure that this last point actually worked and we did not miss anything. And also deprecate older constructors might help.

> 
> My last measurement of Rosenbrock benchmark is below 6 us and about half of all time is performing some actual computations. There is still at least one significant avoidable allocation (in line search) and some potentially avoidable copying but we are entering diminishing returns territory here. I'm quite happy with the overall improvement.

line_search Is luckily just one function, but I had designed it to have as less as possible allocations already. So input on which is avoidable and how would be great, if you have spotted it already.

> BTW, I've quickly checked peformence of quasi-Newton for Rosenbrock and the largest issue is too many allocations. There should be buffers that persist between iterations instead of constantly allocating for everything.

Again, more detail here would be great. I am not an expert on allocations, so just saying – there is something wrong/ inefficient – sure can be done, but then it stays inefficient until someone more clever than me comes along. Also it does definelty not allocate “everything”, that reads like we never reuse anything, which is wrong.

mateuszbaran:
> That reads for me a bit like this is half a PR, since it leave the rest – probably to me – somewhen after I both have understood how this works and I find time to work through all other storages.

This PR fully handles CG which I originally intended but then I noticed that similar changes could be applied in many other places. This is probably more like a quarter of what could be improved about storage. I can go back to that later but I think finishing the fused retraction rework may be more important.

> I would prefer to be sure that this last point actually worked and we did not miss anything. And also deprecate older constructors might help.

Which last point do you mean?
I will take a look at what needs a deprecation.

> line_search Is luckily just one function, but I had designed it to have as less as possible allocations already. So input on which is avoidable and how would be great, if you have spotted it already.

This is a topic worth a separate issue and also involves functions like
```julia
(a::WolfePowellLinesearch)(
    mp::AbstractManoptProblem,
    ams::AbstractManoptSolverState,
    ::Int,
    η=-get_gradient(mp, get_iterate(ams));
    kwargs...,
)
```

> Again, more detail here would be great. I am not an expert on allocations, so just saying – there is something wrong/ inefficient – sure can be done, but then it stays inefficient until someone more clever than me comes along. Also it does definelty not allocate “everything”, that reads like we never reuse anything, which is at least exaggerating.

Yes, sorry, I was exaggerating. Look at the flame graph below though:
image
The ten orange blocks in red curves correspond to allocations performed by `quasi_Newton` solver each step that could be preallocated, so there is still quite a bit to do. I can point at specific lines that cause these allocations but maybe let's not put too much unrelated things here.

Maybe the thing to consider would be making a single storage for a solver that could be shared by perform_step!, line searches, coefficient calculations, stopping criteria, etc.

kellertuer:
Yes, I am sure you did not mean to exaggerate, I was also a bit tired yesterday and maybe overinterpretet.

For an action plan. What about

1. We put the Armijo-Issue into a separate Issue. I have an idea where that is from, the `linesearch_backtrack` that a few linesearches share might allocate
2. We put quasi Newton into its own issue. That sub solver does not yet align with the new subsolver-model/structure we developed last autumn, especially to set parameters in the sub solver is something that currently might allocate.
3. You could first finish the fused retractions, that leaves this PR open for a while and I could
4. check on this PR that the new storage works in the other places as well, after I understood how the fields work, but I then can also help documenting that. If it does not, we would probably need to do changes in this PR anyways. If it does, we can also check the other places and change the storage usage to the more efficient one.

But I not only have the other two PR I am currently not finding much time for, I also have at least 3 additional meetings the next few weeks within a collegial pedagogical mandatory coaching all “new” associate professorsI have to do (besides the language courses). So it might be that step 4 only happens somewhen in March, since currently I only find time to code late in the evenings due to that.

mateuszbaran:
> Yes, I am sure you did not mean to exaggerate, I was also a bit tired yesterday and maybe overinterpretet.

Yes, sorry, I just meant that there are many places where avoidable allocations happen.

> 1. We put the Armijo-Issue into a separate Issue. I have an idea where that is from, the `linesearch_backtrack` that a few linesearches share might allocate

I can make a list of all lines that cause allocation so that's easier to track.

> 2\. We put quasi Newton into its own issue. That sub solver does not yet align with the new subsolver-model/structure we developed last autumn, especially to set parameters in the sub solver is something that currently might allocate.

Sure, I will open an issue.

> 3\. You could first finish the fused retractions, that leaves this PR open for a while and I could

OK, this PR is just an internal change that doesn't block any other work so it can wait.

> 4\. check on this PR that the new storage works in the other places as well, after I understood how the fields work, but I then can also help documenting that. If it does not, we would probably need to do changes in this PR anyways. If it does, we can also check the other places and change the storage usage to the more efficient one.

OK, I think the main issue I need your help with is that each place that wants to store points in the fast storage needs a point as a "template", and similarly for tangent vectors. This requires some redesign unfortunately. Once I have a point or vector at the place creating a storage, I can easily convert it to the fast variant.

> But I not only have the other two PR I am currently not finding much time for, I also have at least 3 additional meetings the next few weeks within a collegial pedagogical mandatory coaching all “new” associate professorsI have to do (besides the language courses). So it might be that step 4 only happens somewhen in March, since currently I only find time to code late in the evenings due to that.

No problem, not having this PR merged doesn't block any other work I'm going to do over the next couple of months so that's fine.

kellertuer:
Cool, that sounds like a plan.
For the records that should be easy, the debugs, e.g. `DebugChange` or so might need an update for worst of cases a breaking one, but I think one can trick that, keeping the old slow Dictionary one as a fallback with a warning/deprecation.

src/plans/solver_state.jl
```diff
@@ -278,14 +361,65 @@ function update_storage!(a::AbstractStateAction, s::AbstractManoptSolverState)
     return a.keys
 end
 
+function _storage_key_true(nt::NamedTuple)
```
kellertuer, line 364: What does this function do? For me the name is mega-cryptic. Does it check whether a key exists? But why does it then only have one parameter?

A small doc and maybe a better function name would help, I am currently not able to understand what it is for and how it is used.

src/plans/solver_state.jl
```diff
@@ -205,65 +225,128 @@ iteration, i.e. acts on `(p,o,i)`, where `p` is a [`AbstractManoptProblem`](@ref
 Initialize the Functor to an (empty) set of keys, where `once` determines
 whether more that one update per iteration are effective
 
-    AbstractStateAction(keys, once=true])
+    function StoreStateAction(
+        general_keys::Vector{Symbol}=Symbol[],
+        point_values::NamedTuple=NamedTuple(),
+        tangent_values::NamedTuple=NamedTuple(),
+        once::Bool=true,
+    ))
```
kellertuer, line 233: I tried for about an hour to understand how I would actually call this, I am not able to understand this.
Can we maybe do a concrete example here how the `NamedTuples` are used? It also implies that we always have to know at-time-of-construction all fields we want to store in points/tuples? This can not grow?

The keys of the NamedTuples are the fields we do storage for (or `:Gradient`/`:Iterate`)? All this would be good to be documented here.

src/plans/solver_state.jl
```diff
@@ -278,14 +361,65 @@ function update_storage!(a::AbstractStateAction, s::AbstractManoptSolverState)
     return a.keys
 end
 
+function _storage_key_true(nt::NamedTuple)
+    return map(key -> NamedTuple{(key,),Tuple{Bool}}(true), keys(nt))
```
kellertuer, line 365: (Sorry for so many questions) – this reads for me as “for every key in nt we return _one_ NamedTuple” so we have an array of NamedTuples afterwards? This seems strange, but probably There is something I am missing here. I also have no clue why this is necessary since there is also nothing filtered or reduced?

And then nt is point or the vector NamedTuple?

src/plans/conjugate_gradient_plan.jl
```diff
@@ -1,3 +1,21 @@
...
+end
+
+function DirectionUpdateRuleStorage(
+    M::AbstractManifold, dur::DirectionUpdateRule, p=rand(M), X=zero_vector(M, p)
+)
+    ursp = update_rule_storage_points(dur)
+    ursv = update_rule_storage_vectors(dur)
+    # StoreStateAction makes a copy
+    sa = StoreStateAction(
+        Symbol[], NamedTuple{ursp}(map(x -> p, ursp)), NamedTuple{ursv}(map(x -> X, ursv))
```
kellertuer, line 14: already more like a detail, should it maybe be `x->copy(M, p)` since we want different memory for every point we store?

kellertuer: ah StoreAction makes a copy, sorry, missed that – one question is of course where the copying fits better here or in the constructor.

kellertuer:
But I looked through remaining ones with Storage (M=manifold exists / X = not necessary to update)

* M, DebugChange has a manifold, so its easy to use `rand(M)` to switch to point_storage
* (M,) DebugGradientChange has to be fixed first and has a manifold then enyways, since we need the retraction anyways.
* X, DebugEntryChange records any entry that is in general not possible to do that more efficient
* X/M, All Primal-Dual-Debug/Records could be updated similar to GradientChange, but is a bit of work, and that Algorithm is not used much
* M, RecordChange had a manifold (one even a p)
* X, RecordEntryChange has to stay inefficient, since the entry might be any state field
* M, NonmonotoneLinesearch has a manifold
* M, StopWhenChangeLess has a manifold

So the ones with M need updates but all besides the PrimalDual ones (we can keep them as is) do have a manifold anyways in their constructors, so one can easily get an p/X.

src/plans/solver_state.jl
mateuszbaran: Hm... when I was writing an answer for you I've noticed that it can be done in a much more simple way. I will change it.

src/plans/solver_state.jl
mateuszbaran: Sorry, I'm haven't seen this post before -- did the previous discussion make it a bit more clear?

src/plans/solver_state.jl
mateuszbaran: I think this will be easier to explain in a video call again :wink: .

The fast storage can't grow in this design but we still can use `general_keys` if we need to add things dynamically. Moreover, we need to have a point/vector for initialization at-time-of-construction. I could make it slightly slower and growable, I could also make it slightly slower and only require type of a point at time of construction if that would be useful.

src/plans/solver_state.jl
mateuszbaran: Questions are perfectly fine :slightly_smiling_face: . As I wrote above, I'm going to remove this function.

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Yes, I've even made a comment about that :slightly_smiling_face: . I'm not sure where the copy would be more fitting but making it in `StoreStateAction` seems safer so that's why I've chosen that.

mateuszbaran:
> I have a (hopefully last) question in usability. As far as I see the point/vector named tuples are fixed in size?
> Could we then maybe implement a `merge(s1, s2)` for two storages, that merges fields of the dictionary and both named tuples? That maybe makes it easier to combine storages into one that can globally be used.

Sure, that should be doable.



> Besides that maybe an easier constructor would be cool, I do not feel I understand how to use the NamedTuple ones. Could we do something like
> 
> ```julia
> StoreStateAction( dictionary_symboles, point_symbols, point_init_value(s), vector_symbols, vector_init_values)
> ```
> 
> where maybe one init value for p/X is enough since we would copy that over? Otherwise maybe even
> 
> ```julia
> StoreStateAction(M, dictionary_symboles, point_symbols, vector_symbols; p_init=rand(M) X_init=zero_vector(M,p_init))
> ```
> 
> ...and all `_symbols` variables are vectors of symbols that refer to fields of the state. This (hopefully, If I understood the new type at least a bit) would make it easier to create the storages. I can help providing these and we could make your constructors the internal ones.

I will check how that could work.

> But I looked through remaining ones with Storage (M=manifold exists / X = not necessary to update)

There is another problematic one with no manifold: `StopWhenChangeLess`. This can actually be used in performance-sensitive scenarios.

> _edit:_ I can help doing these but I would like to first have the points above, especially the constructor questions clarified / implemented before doing this.

Sure, I will improve the code and docs here and explain everything.

src/plans/solver_state.jl
kellertuer: This is also not too old, no I still do not understand that well the constructors, but I hope I understand the NamedTuples by now. They key is a symbol, either the State field or a word with a capital letter like `:Gradient` (where for now the only other we have is `:Iterate`).

src/plans/solver_state.jl
kellertuer: even better – now the code looks fine – I do not yet fully understand it, but it resolved this.

src/plans/solver_state.jl
kellertuer: I think now that I slowly understand it, there is mainly one concern I have (besides the constructor but I that I already mentioned below.

Currently we have a tradeoff: Either we have points/vectors and are fast or we do not have those and are slow. Is there maybe also a trick to resolve the rest?
Now, if we once decide we could do the same with - say - cotangent vectors we again have to write all the accessor functions for example.
I was first thinking of making a dictionary of named tuples? or a named tuple of named tuples? Ah both would not work for the “number-storage” when we just store floats or ints.

src/plans/conjugate_gradient_plan.jl
kellertuer: Ok, sure, hiding it from the user is maybe safer than requiring the user to do copies, when we anyways need them.

kellertuer:
> > I have a (hopefully last) question in usability. As far as I see the point/vector named tuples are fixed in size?
> > Could we then maybe implement a `merge(s1, s2)` for two storages, that merges fields of the dictionary and both named tuples? That maybe makes it easier to combine storages into one that can globally be used.
> 
> Sure, that should be doable.

Cool. Because that was one of my first ideas that – r.g. a stopping criterion, the CG step and maybe also the debug share a storage.
Then any constructor that gets a storage and does not create one you either check that all necessary fields are in or anyways always merge with one such that there is definetly all required fields in.


> > Besides that maybe an easier constructor would be cool, [...]
> 
> I will check how that could work.

Cool, I think that is currently the main point why I hesitate to rework the actions, but once that is easily doable that would improve the usability side here quite a bit – don't get me wrong I trust you and its cool that it is fast already :) 

> > But I looked through remaining ones with Storage (M=manifold exists / X = not necessary to update)
> 
> There is another problematic one with no manifold: `StopWhenChangeLess`. This can actually be used in performance-sensitive scenarios.

I can check that one, but since that might be generated in the factory for stopping criteria, that might actually be tricky (but maybe I have the manifold available in that factory, I'll check).

kellertuer:
I checked `StopWhenChangeLess` – it has the `manifold=` keyword, only that since this is optional, we usually can not use it for your purpose.

We could do two constructors (to stay nonbreaking) – one without a manifold upfront.
If your “lookup” can handle a case where “if the key is not in the point list we look into the Dictionary” then all is fine.

Currently I think this is not the case since both vector/point have their own get/set methods right? Hm. Would one get/set pair be possible instead of the current 3?

src/plans/solver_state.jl
mateuszbaran: > Currently we have a tradeoff: Either we have points/vectors and are fast or we do not have those and are slow. Is there maybe also a trick to resolve the rest?

Sure, we can have efficient storage for other things as well but I haven't noticed other things being used that much so I haven't worked on that.

> Now, if we once decide we could do the same with - say - cotangent vectors we again have to write all the accessor functions for example.

I can unify accessors by differentiating storage pools by a type wrapper, so instead of `get_point_storage(u.storage, :Iterate)` we would have something like `get_storage(u.storage, PointStorage(:Iterate))`.

> I was first thinking of making a dictionary of named tuples? or a named tuple of named tuples? Ah both would not work for the “number-storage” when we just store floats or ints.

Dictionary won't work because it couldn't always be type-stable, for example. Nested named tuples would work but the code would be quite convoluted (even more convoluted than it is now). They could store raw numbers if we modified the tuple itself instead of its content but that's another level of convolution.

mateuszbaran:
> > > I have a (hopefully last) question in usability. As far as I see the point/vector named tuples are fixed in size?
> > > Could we then maybe implement a `merge(s1, s2)` for two storages, that merges fields of the dictionary and both named tuples? That maybe makes it easier to combine storages into one that can globally be used.
> > 
> > 
> > Sure, that should be doable.
> 
> Cool. Because that was one of my first ideas that – r.g. a stopping criterion, the CG step and maybe also the debug share a storage. Then any constructor that gets a storage and does not create one you either check that all necessary fields are in or anyways always merge with one such that there is definetly all required fields in.

That's a neat idea, though it would be a fairly significant rework.

> I checked `StopWhenChangeLess` – it has the `manifold=` keyword, only that since this is optional, we usually can not use it for your purpose.
> 
> We could do two constructors (to stay nonbreaking) – one without a manifold upfront. If your “lookup” can handle a case where “if the key is not in the point list we look into the Dictionary” then all is fine.
> 
> Currently I think this is not the case since both vector/point have their own get/set methods right? Hm. Would one get/set pair be possible instead of the current 3?

Currently there is no fallback in the lookup but I can add that quite easily. I've suggested a possible lookup function unification here: https://github.com/JuliaManifolds/Manopt.jl/pull/212#discussion_r1097268185 . They would still have separate methods, but for a single function.

It would certainly be possible to have slow storage as a backup in `StopWhenChangeLess`. If we went for storage unification, though, then it wouldn't even be necessary I think because storage instantiation could happen later in the pipeline. `StopWhenChangeLess` would just say "I need the previous iterate" and solver initialization could handle storage creation. So we wouldn't even have to merge concrete storage instances, just key sets.

src/plans/solver_state.jl
kellertuer: I like the storage thing – and sure if it is not possible more generally, then I am fine, if we keep the general accessors, like you just prosed; because then, _if_ we come up with an efficient way and can store, for example, cotangents also easily, we do not have to change the access/store function interfaces.

kellertuer:
> > > > I have a (hopefully last) question in usability. As far as I see the point/vector named tuples are fixed in size?
> > > > Could we then maybe implement a `merge(s1, s2)` for two storages, that merges fields of the dictionary and both named tuples? That maybe makes it easier to combine storages into one that can globally be used.
> > > 
> > > 
> > > Sure, that should be doable.
> > 
> > 
> > Cool. Because that was one of my first ideas that – r.g. a stopping criterion, the CG step and maybe also the debug share a storage. Then any constructor that gets a storage and does not create one you either check that all necessary fields are in or anyways always merge with one such that there is definetly all required fields in.
> 
> That's a neat idea, though it would be a fairly significant rework.

Hm, that was the initial design idea of the storage, maybe not well communicated/documented, but that was the idea. Otherwise all 3 could really also just store their own stuff actually (if they have separate storages always anyways).

I do not think it is too much work. We need a `DefaultStorage` type maybe that the keyword is set to, if you do change that, one can merge it with the original Default one had to obtain the common storage. I am just not 100% sure that works.

> 
> > I checked `StopWhenChangeLess` – it has the `manifold=` keyword, only that since this is optional, we usually can not use it for your purpose.
> > We could do two constructors (to stay nonbreaking) – one without a manifold upfront. If your “lookup” can handle a case where “if the key is not in the point list we look into the Dictionary” then all is fine.
> > Currently I think this is not the case since both vector/point have their own get/set methods right? Hm. Would one get/set pair be possible instead of the current 3?
> 
> Currently there is no fallback in the lookup but I can add that quite easily. I've suggested a possible lookup function unification here: [#212 (comment)](https://github.com/JuliaManifolds/Manopt.jl/pull/212#discussion_r1097268185) . They would still have separate methods, but for a single function.
> 
> It would certainly be possible to have slow storage as a backup in `StopWhenChangeLess`. If we went for storage unification, though, then it wouldn't even be necessary I think because storage instantiation could happen later in the pipeline. `StopWhenChangeLess` would just say "I need the previous iterate" and solver initialization could handle storage creation. So we wouldn't even have to merge concrete storage instances, just key sets.

This one with the unification is too fast for me in explanation again.

mateuszbaran:
> Hm, that was the initial design idea of the storage, maybe not well communicated/documented, but that was the idea. Otherwise all 3 could really also just store their own stuff actually (if they have separate storages always anyways).
> 
> I do not think it is too much work. We need a `DefaultStorage` type maybe that the keyword is set to, if you do change that, one can merge it with the original Default one had to obtain the common storage. I am just not 100% sure that works.

I haven't noticed any part that suggested actually attempting to unify storage but maybe I haven't looked in the right place? For one thing, in such design storage wouldn't be kept internally in a field in actions but rather passed to relevant functions because the "default" (global) storage would have to be instantiated later than action creation. You only know what fields are needed after you gather all actions.

> This one with the unification is too fast for me in explanation again.

I think I will just make a commit with the change, it looks like a good idea anyway.

kellertuer:
> > Hm, that was the initial design idea of the storage, maybe not well communicated/documented, but that was the idea. Otherwise all 3 could really also just store their own stuff actually (if they have separate storages always anyways).
> > I do not think it is too much work. We need a `DefaultStorage` type maybe that the keyword is set to, if you do change that, one can merge it with the original Default one had to obtain the common storage. I am just not 100% sure that works.
> 
> I haven't noticed any part that suggested actually attempting to unify storage but maybe I haven't looked in the right place? For one thing, in such design storage wouldn't be kept internally in a field in actions but rather passed to relevant functions because the "default" (global) storage would have to be instantiated later than action creation. You only know what fields are needed after you gather all actions.

Ah then not in your unifying sense. My idea was: You would create a storage upfront and pass it to all that need a storage (e.g. CGState/CGStep, Debug, StoppingCriterion)... but if you would miss a field that would also be added and be fault tolerant in that case. But we can also error if we miss afield, that would maybe be the easier way.

> > This one with the unification is too fast for me in explanation again.
> 
> I think I will just make a commit with the change, it looks like a good idea anyway.

Cool :) Maybe I understand it then.

mateuszbaran:
> > > Hm, that was the initial design idea of the storage, maybe not well communicated/documented, but that was the idea. Otherwise all 3 could really also just store their own stuff actually (if they have separate storages always anyways).
> > > I do not think it is too much work. We need a `DefaultStorage` type maybe that the keyword is set to, if you do change that, one can merge it with the original Default one had to obtain the common storage. I am just not 100% sure that works.
> > 
> > 
> > I haven't noticed any part that suggested actually attempting to unify storage but maybe I haven't looked in the right place? For one thing, in such design storage wouldn't be kept internally in a field in actions but rather passed to relevant functions because the "default" (global) storage would have to be instantiated later than action creation. You only know what fields are needed after you gather all actions.
> 
> Ah then not in your unifying sense. My idea was: You would create a storage upfront and pass it to all that need a storage (e.g. CGState/CGStep, Debug, StoppingCriterion)... but if you would miss a field that would also be added and be fault tolerant in that case. But we can also error if we miss afield, that would maybe be the easier way.

I see. I was thinking about automatic unification, not manual, but now I see that a manual one should be somewhat possible.

kellertuer:
Maybe automatic is too complicated, yes, then lets too manual.

mateuszbaran:
Automatic storage would be much less complicated from user's point of view though.

BTW, I've just commited the storage accessor unification thing.

kellertuer:
Yes, the accessors look nice :)

Sure Automatic is nicer for the user, but how large would the effort do realise this be?

mateuszbaran:
> Sure Automatic is nicer for the user, but how large would the effort do realise this be?

I think it wouldn't bee too much work, there is just lots of other more important work.

kellertuer:
I leave that up to you; I feel we have a good mix here of you seeing the improvements in time and performance in general, while I try to keep it usable.

My main point for this PR for now is a nice constructor, the rest is fine, also if we have errors for missing fields (which with the fallback I think we do not have).
So having a constructor where the user does not have to create the NamedTuple that would be helpful I think, since in your lines of creating these you also always mention the type, which I do not want to tell the user to do, it should be easier for a user.

And after this PR maybe a storage tutorial would also be helpful :)

mateuszbaran:
I've added the easier constructor for `StoreStateAction`, does it look fine for you? 

> also if we have errors for missing fields (which with the fallback I think we do not have).

Yes, there are fallbacks so no errors are necessary.

> And after this PR maybe a storage tutorial would also be helpful :)

I think we should improve global storage usability first 

> So the ones with M need updates but all besides the PrimalDual ones (we can keep them as is) do have a manifold anyways in their constructors, so one can easily get an p/X.

Currently many (if not all) of the actions you've listed that have manifold use it only to get default retraction/inverse retraction/vector transport. `DefaultManifold` is a reasonable fallback in these cases. Fast storage, though, requires accurate manifold information so it would have to be a mandatory argument, or there needs to be a way to distinguish accurate manifold information and the default manifold. In the default case we could fall back to slow storage. Do you think something like
```julia
if M === DefaultManifold(1)
    # use slow storage
else
    # use fast storage
end
```
would make sense?

mateuszbaran:
Should I make a separate file for tests related directly to `StoreStateAction`? Or just extend one of existing files like `test_debug`? There is a couple of features that don't seem to be used currently.

src/plans/solver_state.jl
```diff
@@ -242,6 +242,21 @@ For the point storage use `PointStorageKey`. For tangent vector storage use
...
+        ::Type{TPS},
+        ::Type{TTS};
+        p_init=rand(M),
+        X_init=zero_vector(M, p_init),
+        once=true,
+    ) where {TPS<:Tuple,TTS<:Tuple}
+
+Initialize the general storage keys to `dictionary_symbols`, point storage keys to `TPS` and
+tangent vector storage tu `TTS`. For example you may call
+`StorageStateAction(M, Symbol[], Tuple{:Iterate}, Tuple{:Gradient})` to create efficient
```
kellertuer, line 257: Ah this is good to read here, because just from `Type{TPS}`  I would have thought you pass the point type in here. Why does it have to be the type of a tuple and not a tuple? Just curious, because maybe a tuple might be even easier to use here – but sure this is a good improvement over the more-internal other constructor. Thanks!

kellertuer:
> I've added the easier constructor for `StoreStateAction`, does it look fine for you?
> 
> > also if we have errors for missing fields (which with the fallback I think we do not have).
>
> Yes, there are fallbacks so no errors are necessary.

Cool!

> 
> > And after this PR maybe a storage tutorial would also be helpful :)
> 
> I think we should improve global storage usability first

ok, let me know what you think we can improve. 

> > So the ones with M need updates but all besides the PrimalDual ones (we can keep them as is) do have a manifold anyways in their constructors, so one can easily get an p/X.
> 
> Currently many (if not all) of the actions you've listed that have manifold use it only to get default retraction/inverse retraction/vector transport. `DefaultManifold` is a reasonable fallback in these cases. Fast storage, though, requires accurate manifold information so it would have to be a mandatory argument, or there needs to be a way to distinguish accurate manifold information and the default manifold. In the default case we could fall back to slow storage. Do you think something like
> 
> ```julia
> if M === DefaultManifold(1)
>     # use slow storage
> else
>     # use fast storage
> end
> ```
> 
> would make sense?

This is indeed a complicated point – since making it mandatory has two disadvantages
1. its breaking, but well if that needs to be, then it is what it is
2. We might easily end up with _every_ function needing the manifold first, which might make calling things a bit clumsy again, since we basically end up in a naming scheme like
```Julia 
f(M, g(M, p; storage= h(M, p)), j(M), k(M))
```
(if e.g. f is the solver, g is the step size, j is something with a storage, and k...ok I am running out of spontaneous ideas)

So manu Ms is more like a meta-level argument against making M mandatory, so I think the fallback checking for the default Manifold might be a way to go that keeps the interface a bit nicer for a starting user?

kellertuer:
> Should I make a separate file for tests related directly to `StoreStateAction`? Or just extend one of existing files like `test_debug`? There is a couple of features that don't seem to be used currently.

I think with the current improvements probably and own test file is valid. We could even put all storage functions into their own file in the plan folder if you like.

src/plans/solver_state.jl
mateuszbaran: I actually like the new constructor too :slightly_smiling_face: .

It's `Tuple{:Iterate}` instead of `(:Iterate,)` basically just because I don't trust Julia's constant propagation enough to handle the symbols right, and using a tuple type forces Julia to propagate these keys.

mateuszbaran:
> > > And after this PR maybe a storage tutorial would also be helpful :)
> > 
> > 
> > I think we should improve global storage usability first
> 
> ok, let me know what you think we can improve.

OK, though I think that's a topic for another issue. Feel free to try making a centralized storage and see if it works when it's used by solver, stopping criteria and some debug action at the same time. Currently updating storage is somewhat decentralized so I fear it may be done at incorrect points for a centralized storage. This is one example of what I was thinking about. When this works we may start thinking making a preliminary tutorial but it would be nice to have automatic global storage as well.

> This is indeed a complicated point – since making it mandatory has two disadvantages
> 
> 1. its breaking, but well if that needs to be, then it is what it is
> 2. We might easily end up with _every_ function needing the manifold first, which might make calling things a bit clumsy again, since we basically end up in a naming scheme like
> 
> ```julia
> f(M, g(M, p; storage= h(M, p)), j(M), k(M))
> ```
> 
> (if e.g. f is the solver, g is the step size, j is something with a storage, and k...ok I am running out of spontaneous ideas)
> 
> So manu Ms is more like a meta-level argument against making M mandatory, so I think the fallback checking for the default Manifold might be a way to go that keeps the interface a bit nicer for a starting user?

Yes, I can handle supporting both fast and slow storage. I'm just unsure if using `DefaultManifold` for marking "use the slow storage" is the right idea when it is a perfectly fine manifold to optimize a function on? Maybe we should have a `DummyDefaultManifold` just for this purpose?

kellertuer:
> > > > And after this PR maybe a storage tutorial would also be helpful :)
> > > 
> > > 
> > > I think we should improve global storage usability first
> > 
> > 
> > ok, let me know what you think we can improve.
> 
> OK, though I think that's a topic for another issue. Feel free to try making a centralized storage and see if it works when it's used by solver, stopping criteria and some debug action at the same time. Currently updating storage is somewhat decentralized so I fear it may be done at incorrect points for a centralized storage. This is one example of what I was thinking about. When this works we may start thinking making a preliminary tutorial but it would be nice to have automatic global storage as well.

A centralised storage would probably a singleton the elements would “connect to”? But sure this is something for another issue. I still think even writing the tutorial about how the storage can be used nicely for now might be nice for the user.
Sure the step to a centralised one might be really fancy and nice, but then we would just update the tutorial as well :)

A singleton centralised storage might be disadvantageous if you compare two solvers somewhere or two manifolds or so. I have to think about that.

> 
> > This is indeed a complicated point – since making it mandatory has two disadvantages
> > 
> > 1. its breaking, but well if that needs to be, then it is what it is
> > 2. We might easily end up with _every_ function needing the manifold first, which might make calling things a bit clumsy again, since we basically end up in a naming scheme like
> > 
> > ```julia
> > f(M, g(M, p; storage= h(M, p)), j(M), k(M))
> > ```
> > 
> >   
> > (if e.g. f is the solver, g is the step size, j is something with a storage, and k...ok I am running out of spontaneous ideas)
> > So manu Ms is more like a meta-level argument against making M mandatory, so I think the fallback checking for the default Manifold might be a way to go that keeps the interface a bit nicer for a starting user?
> 
> Yes, I can handle supporting both fast and slow storage. I'm just unsure if using `DefaultManifold` for marking "use the slow storage" is the right idea when it is a perfectly fine manifold to optimize a function on? Maybe we should have a `DummyDefaultManifold` just for this purpose?

Why not just `nothing` or `missing` then? I mean for now I used the `DefaultManifold` to get Exp/Log/PT from the default, but we could also dispatch `nothing` to that (if that does not even work directly now already). `nothing` as an indicator for slow storage is probably easier to check.
Or we could even keep the current constructors (those without manifolds) for the slow case and write new ones below that with a manifold upfront for the fast.

mateuszbaran:
> A centralised storage would probably a singleton the elements would “connect to”?

I would phrase it as: there is just one `StorageStateAction` object used by all every part that needs storage. Currently it can be achieved by passing said object explicitly to relevant constructors.

> Why not just `nothing` or `missing` then? I mean for now I used the `DefaultManifold` to get Exp/Log/PT from the default, but we could also dispatch `nothing` to that (if that does not even work directly now already). `nothing` as an indicator for slow storage is probably easier to check.
> Or we could even keep the current constructors (those without manifolds) for the slow case and write new ones below that with a manifold upfront for the fast.

Yes, right, writing new constructors would likely be the best idea here. IIRC you don't like `nothing` in interfaces so I'm avoiding suggesting that :slightly_smiling_face: .

kellertuer:
> > A centralised storage would probably a singleton the elements would “connect to”?
> 
> I would phrase it as: there is just one `StorageStateAction` object used by all every part that needs storage. Currently it can be achieved by passing said object explicitly to relevant constructors.

That sounds like a singleton pattern. The only problem I have with that is if you have a script/notebook that works by comparing solvers/objectives maybe even on the same manifold with different representations of points (Hyperbolic for example), then we would need 2 storages, one for the group doing the Hyperboloid things, one for the group Hyperplane-ing. So “one” is a bit too short maybe? But let's leave that for a future PR; one could do singletons addressed by manifold/point/vector.

> 
> Yes, right, writing new constructors would likely be the best idea here. IIRC you don't like `nothing` in interfaces so I'm avoiding suggesting that 🙂 .

Totally fine, then we could still use `DefaultManifold()` to really take the simplest one we have.

mateuszbaran:
> > > A centralised storage would probably a singleton the elements would “connect to”?
> > 
> > 
> > I would phrase it as: there is just one `StorageStateAction` object used by all every part that needs storage. Currently it can be achieved by passing said object explicitly to relevant constructors.
> 
> That sounds like a singleton pattern. The only problem I have with that is if you have a script/notebook that works by comparing solvers/objectives maybe even on the same manifold with different representations of points (Hyperbolic for example), then we would need 2 storages, one for the group doing the Hyperboloid things, one for the group Hyperplane-ing. So “one” is a bit too short maybe? But let's leave that for a future PR; one could do singletons addressed by manifold/point/vector.

I don't mean a singleton pattern -- to me singleton means one object globally. Here I mean one object per optimization state + things related to that state, which is not the same thing.

kellertuer:
Ah, then I misunderstood you. That is of course more fitting – though I do not have much of an idea how to realise that, especially how to find out which points/vectors to store. But this is something for another PR/Issue

kellertuer:
I lost a bit of track what might be missing here – tests look fine already, should we adapt the ones I listed or do we do that on a next PR, since this is non-breaking the ones not updated would for now just stay in “slow-mode”?

mateuszbaran:
IIRC the main remaining point was enabling more things to use the fast storage. I was going to finish fused retractions first before going back to this though. Or do you want to have this finished first?

kellertuer:
No, its not that urgent, and an advantage might even be (since the other PRs here are done) that I might find time to enable faster storage for more things.

kellertuer:
I fear we have to improve how to generate an efficient `StoreStateAction`. I still don't understand how that works.

I wanted to check with `DebugChange` whether I can find out what needs to be updated and was wondering why we need the manifold. at all (because I forgot). I am probably confused by this constructor docs

```
    function StoreStateAction(
        general_keys::Vector{Symbol}=Symbol[],
        point_values::NamedTuple=NamedTuple(),
        tangent_values::NamedTuple=NamedTuple(),
        once::Bool=true,
    ))
```
That requires that we _have_ already things stored in the named tuples for points/tangent vectors (compared to the other one that needs init_p and init_X)? I do not understand what said text below that says about the two NamedTuples.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -1,3 +1,19 @@
+
+struct DirectionUpdateRuleStorage{TC<:DirectionUpdateRule,TStorage<:StoreStateAction}
+    coefficient::TC
+    storage::TStorage
+end
+
+function DirectionUpdateRuleStorage(
```
kellertuer, line 7: maybe we could unify this to use keywords for p and x as `init_p=` and `init_X=` similar to the storage itself?

src/plans/conjugate_gradient_plan.jl
```diff
@@ -144,9 +162,9 @@ Then the coefficient reads
 See also [`conjugate_gradient_descent`](@ref)
 
 # Constructor
-    DaiYuanCoefficient(
-        t::AbstractVectorTransportMethod=ParallelTransport(),
-        a::StoreStateAction=(),
+    function DaiYuanCoefficient(
+        M::AbstractManifold=DefaultManifold(2);
+        t::AbstractVectorTransportMethod=default_vector_transport_method(M)
```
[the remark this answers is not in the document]
kellertuer: Maybe for now we could keep the old one

`Coeff(t)` to specify the transport method directly or
`Coeff(M=DefaultManifold(); vector_transport_method=default_vector_transport_method(M))` for the new one that can use the default? We could even deprecate the first.

The first makes it nonbreaking (unless someone ver specified a storage here) the second one is the future/nicer one

kellertuer:
So I spent a whole hour just to update _one_ debug that uses storage. I do believe the approach we have is very performant, but from this try I have one or two questions and maybe remarks on usability

So first of all the change is here the commit https://github.com/JuliaManifolds/Manopt.jl/pull/212/commits/5d5bc8199d495effb5f54b8fe51e27533631c06f
which uses the opportunity to also move to a unified naming scheme with Manifolds.jl – instead of invretr to use `inverse_retraction_method` and deprecate the `manifold=` keyword in favour of a positional first argument for the manifold.

Besides that my question is whether this is how this is intended to be constructed
1) We basically have to use in line 158
https://github.com/JuliaManifolds/Manopt.jl/commit/5d5bc8199d495effb5f54b8fe51e27533631c06f#diff-d8a6071071e9c63dc320bb2cd0fa05ea84f8743be3df3a4dd0234a9e0ca5eb4fR158
nothing as a default since otherwise it is hard to check which field to initialise?

2)Concerning the initialisation of a storage if the user does not pass some (line 179-185): passing always a `Symbol[]` seems also like a detail the user might not be interested in. Maybe we could move all 3 to keywords? My preferred way to initialise storage would maybe be

```
StoreStateAction(M;
   store_fields = [:MyFirstField], #maybe floats or other types here
   store_points = [:Iterate], #or other fields storing points here
   store_vectors = [:X], # or others here
   init_p = rand(M)
   init_X = rand(M; vector_at=p)
)
```
This would be nicely unified, one does not have to pass anything empty somewhere just to clarify others – but sure a question is whether this is easily doable, since one has to convert the `Symbol[]` of `state_points`

We could do the second interface as

```
StoreStateAction(;
   store_fields = [:MyFirstField], #maybe floats or other types here
)
```
which is basically only the old storage?

And a final thing I am not so sure about: What happens if one field is provided twice (e.g. in fields and points?) does points “win”?

I think with such constructors the storage would also be easy to use.

I hope my adaption of the first one here is ok, for me it seems this might be breaking (if anyone ever used storage actively, which not even I did).

mateuszbaran:
> I fear we have to improve how to generate an efficient `StoreStateAction`. I still don't understand how that works.
> 
> I wanted to check with `DebugChange` whether I can find out what needs to be updated and was wondering why we need the manifold. at all (because I forgot). I am probably confused by this constructor docs
> 
> ```
>     function StoreStateAction(
>         general_keys::Vector{Symbol}=Symbol[],
>         point_values::NamedTuple=NamedTuple(),
>         tangent_values::NamedTuple=NamedTuple(),
>         once::Bool=true,
>     ))
> ```
> 
> That requires that we _have_ already things stored in the named tuples for points/tangent vectors (compared to the other one that needs init_p and init_X)? I do not understand what said text below that says about the two NamedTuples.

I think I could even remove that constructor if it confuses you. It doesn't provide anything important over the new one.

But yes, in either case you need to have an example of what will be stored in there later.

Manifold is needed to:
1. Get examples `p` and `X` if not provided.
2. (not currently, but maybe it should) Appropriately copy `p` and `X` for storage using `copy(M, p)`  and `copy(M, p, X)`.

> 1. We basically have to use in line 158
>    [5d5bc81#diff-d8a6071071e9c63dc320bb2cd0fa05ea84f8743be3df3a4dd0234a9e0ca5eb4fR158](https://github.com/JuliaManifolds/Manopt.jl/commit/5d5bc8199d495effb5f54b8fe51e27533631c06f#diff-d8a6071071e9c63dc320bb2cd0fa05ea84f8743be3df3a4dd0234a9e0ca5eb4fR158)
>    nothing as a default since otherwise it is hard to check which field to initialise?

Yes, that part looks fine.

> 2)Concerning the initialisation of a storage if the user does not pass some (line 179-185): passing always a `Symbol[]` seems also like a detail the user might not be interested in. Maybe we could move all 3 to keywords? My preferred way to initialise storage would maybe be
> 
> ```
> StoreStateAction(M;
>    store_fields = [:MyFirstField], #maybe floats or other types here
>    store_points = [:Iterate], #or other fields storing points here
>    store_vectors = [:X], # or others here
>    init_p = rand(M)
>    init_X = rand(M; vector_at=p)
> )
> ```
> 
> This would be nicely unified, one does not have to pass anything empty somewhere just to clarify others – but sure a question is whether this is easily doable, since one has to convert the `Symbol[]` of `state_points`

Sure, we can pass all of them as keywords. I kept `dictionary_symbols` as the first argument because currently it's still the most often used part.

> We could do the second interface as
> 
> ```
> StoreStateAction(;
>    store_fields = [:MyFirstField], #maybe floats or other types here
> )
> ```
> 
> which is basically only the old storage?

Sure, that's easily doable.

> And a final thing I am not so sure about: What happens if one field is provided twice (e.g. in fields and points?) does points “win”?

Both would win, that is you would have the same thing stored twice :slightly_smiling_face: . At least that's how it would currently work.

> I think with such constructors the storage would also be easy to use.
> 
> I hope my adaption of the first one here is ok, for me it seems this might be breaking (if anyone ever used storage actively, which not even I did).

I will take a closer look. Maybe it could be made non-breaking but anyway storage looks like an internal Manopt thing so I really doubt anyone uses that. And if one does, I can help them fix their code.

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Sure, I will change that.

src/plans/conjugate_gradient_plan.jl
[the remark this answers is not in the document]
mateuszbaran: Sure, I will make it accept vector transport method again.

kellertuer:
Thanks for the comments. If we could introduce the constructors like I proposed, I think its not too hard to do the other storage cases liste above as well, so I will work on that afterwards :)

src/plans/conjugate_gradient_plan.jl
mateuszbaran: Done.

src/plans/conjugate_gradient_plan.jl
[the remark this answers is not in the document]
mateuszbaran: This is also done.

kellertuer:
> > ```
> >     function StoreStateAction(
> >         general_keys::Vector{Symbol}=Symbol[],
> >         point_values::NamedTuple=NamedTuple(),
> >         tangent_values::NamedTuple=NamedTuple(),
> >         once::Bool=true,
> >     ))
> > ```
> 
> I think I could even remove that constructor if it confuses you. It doesn't provide anything important over the new one.

I would be fine with removing it

> > ```
> > StoreStateAction(M;
> >    store_fields = [:MyFirstField], #maybe floats or other types here
> >    store_points = [:Iterate], #or other fields storing points here
> >    store_vectors = [:X], # or others here
> >    init_p = rand(M)
> >    init_X = rand(M; vector_at=p)
> > )
> > ```
> > 
> > This would be nicely unified, one does not have to pass anything empty somewhere just to clarify others – but sure a question is whether this is easily doable, since one has to convert the `Symbol[]` of `state_points`
> 
> Sure, we can pass all of them as keywords. I kept `dictionary_symbols` as the first argument because currently it's still the most often used part.

No problem I can help reworking that, I just struggle a bit getting from a `Symbol[]` array for the new storages to the correct way setting them up in this constructor

kellertuer:
Since `DebugGradientChange` is fixed in #211, I would like to wait (a) for Manifolds.jl to work with ManifoldsBase 0.14 (b) update #210 and #211 accordingly and merge them to master, before finalising the last storages here (and finally merging this and publishing a new version including all these three things).

This might yield a few merge conflicts when merging the updated master back here, but I can fix those as well for sure.

kellertuer:
Merging is done and it was good to not work on the other storage updates earlier.

I can now go through the rest but for best of cases the constructor we discussed would make a lot of the rework easier :)

mateuszbaran:
Thanks, I will work on the constructor now.

mateuszbaran:
Hm, it looks like you haven't merged master correctly.

Also, I've quickly added the new constructor :slightly_smiling_face: . The old ones can be removed later.

kellertuer:
What did I miss in merging? Tests ran fine locally...

I will not work on the rest today, I will just register the examples and then call it day.

kellertuer:
Ah, I see there is one test of the Exact Penalty Method that Avery 5th-or-so run on Windows yields NaN instead of a value less than 0.02 – but I am not sure how to debug that if it only happens on windows and only in a few cases (without the test having randomness).

kellertuer:
> Also, I've quickly added the new constructor 🙂 . The old ones can be removed later.

At first glance the new constructor looks exactly like I had in mind. Will try to do a few new Constructors in Debug tomorrow then.

src/plans/solver_state.jl
```diff
@@ -207,67 +244,201 @@ iteration, i.e. acts on `(p,o,i)`, where `p` is a [`AbstractManoptProblem`](@ref
...
+    ::Type{TPS},
+    ::Type{TTS};
+    p_init=rand(M),
+    X_init=zero_vector(M, p_init),
+    once=true,
+) where {TPS<:Tuple,TTS<:Tuple}
+    TPS_tuple = Tuple(TPS.parameters)
+    TTS_tuple = Tuple(TTS.parameters)
+    point_values = NamedTuple{TPS_tuple}(map(_ -> p_init, TPS_tuple))
+    tangent_values = NamedTuple{TTS_tuple}(map(_ -> X_init, TTS_tuple))
```
kellertuer, line 331: I statt correct that this does not copy? Would that mean that all fields share the same point memory? That would not be so good.

kellertuer: Hm and the new constructor still requires to pass `store_points=Tuple{:Iterate})` ? I was hoping wee could unify all 3 to an array of symbols?

kellertuer: ...and when I have understood it I would love to provide docs how to use this ;)

kellertuer: finally: StoreStateAction is one of the few things where the manifold is now mandatory. can we fid a way around that? Without a manifold of course only the slow storage works.

mateuszbaran: `copy`ing is currently done in the inner constructor, though that will have to change a bit.

> Hm and the new constructor still requires to pass `store_points=Tuple{:Iterate})` ? I was hoping wee could unify all 3 to an array of symbols?

`Tuple{:Iterate}`  makes it much easier to optimize. A vector of symbols is definitely an option, though it makes optimization more difficult and adds some unavoidable overhead -- that's why I'd prefer to use `Tuple{:Iterate}`.

I'm not sure what part of usage is unclear though? Apart, of course, from "why does it need to be like that".

kellertuer: I just mean the constructor, not the inner stuff, but for now...I posted this 45 minutes ago and the number of new storage I was able to do including testing is still 0 – maybe I am the stupid here, but if I as a developer of this package, can not use that properly, it might be the case for others as well.

mateuszbaran: > finally: StoreStateAction is one of the few things where the manifold is now mandatory. can we fid a way around that? Without a manifold of course only the slow storage works.

In the current inner constructor you don't need to provide a manifold. You can just use the same old syntax (just with a vector instead of a tuple). Would you prefer a different argument list?

kellertuer: > "why does it need to be like that".

It‘s not the why it's the how. I am mainly guessing and sometimes it works and sometimes not and I am not sure why. Currently my tests that I start writing randomly fail (probably not that random but) without me understanding why.

mateuszbaran: There might be a bug in my code, so maybe I could help if you posted how you make a storage object and what errors you get?

kellertuer: So I got a first one working but I was fighting for a bit with whether its

* `store_points = [:Iterate, :SecondPoint]` (my preferred way
* `store_points = Tuple(:Iterate, :SecondPoint)` my second idea what it might be
* `store_points = typeof(Tuple(:Iterate, :SecondPoint))` (saw that somewhere)
* `store_points = Tuple{:Iterate, :SecondPoint}` <- I think this is the right one?

or is there another `typeof` in between?

That is maybe the summary of my problem with the current constructor, it is not yet documented and does not provide an intuitive way (which in my opinion should be the first). And it is a constructor, so I do not care about performance. For example currently the symbols are extracted from `TTP.parameters` why not directly allow a vector of symbols?

kellertuer:
Since this thread got quite long, here is the (hopefully final) todo list concerning storage updates

* [x] `DebugChange`
* [x] `DebugGradientChange`
* [x]  `DebugEntryChange` default not improvable)
* [ ] `RecordChange`
* [x] `RecordEntryChange` (same as DEbugEntryChange, not improvable)
* [ ] `NonmonotoneLinesearch`
* [ ] `StopWhenChangeLess`

src/plans/solver_state.jl
mateuszbaran: Sorry, I forgot to document it. It goes like this:
```julia
StorageStateAction(M; store_fields=Symbol[], store_points=Tuple{:Iterate}, store_vectors=Tuple{:Gradient})
```

mateuszbaran: > And it is a constructor, so I do not care about performance. For example currently the symbols are extracted from `TTP.parameters` why not directly allow a vector of symbols?

It's... complicated. Mostly: it's me doing what I know works instead of spending hours benchmarking a bit nicer alternative in various scenarios making sure there are no big performance problems. Interactions of type unstable code (and with `[:some_point]` instead of `Tuple{:some_point}` we have type unstable code) with Julia's type manipulations have caused me many problems in the past.

kellertuer: Hm, but now I use hours trying to guess how to use it. And I do not feel that this is really usable. So for best of cases no one else will touch these internals, if they would like to, we might loose that user since that is far away from an intuitive interface design.

mateuszbaran: I will modify the constructor to make it accept both tuple types and vectors of symbols. It can easily accept either with no performance problems so maybe that would be the best solution?

kellertuer: What we definitely need is updated docs for the StateActoinStorage, but I would really prefer if that was more unified, I am still nowhere in the storage sure which parameters to use from the list above (ok the first one never uses, but thats the one I would like to use).

mateuszbaran: Now you can use it like `StoreStateAction(M; store_fields=[:my, :fields, :to, :store], store_points=[:Iterate], store_vectors=[:Gradient])` (fast) or  `StoreStateAction([:my, :fields, :to, :store, :Iterate, :Gradient])` (slow). I don't think it can get any more unified?

kellertuer: That would have spared me an hour today having that ;) I will check it just now and write the docs.

kellertuer:
The new Storage does not work for manifolds like the Circle it seems.

mateuszbaran:
Yes, currently you can't pass a plain number to the fast storage thing constructor. I will think how to solve that. The easiest solution would be wrapping in a zero-dimensional array (in the constructor) I think.

src/plans/solver_state.jl
kellertuer: I love the new interface (but I am lazy for now and since the previous one works as well leave the current code as is).

kellertuer:
All storages should be updated, the docs for the `StoreStateAction` also document (at least the easier) keyword possibilities.

The only thing we could think about is a constructor without a manifold that only uses the old storage, but I am not sure how useful that is, for now we have a manifold at most places, just that M is not needed in the constructor if we only use the old storage type.

kellertuer:
> The new Storage does not work for manifolds like the Circle it seems.

One trick I used for now in one place – if it is not a default manifold, I‘ll check if `rand(M)` is a number and then only use the old storage as well (I only tested one of the storages with Circle). We could generalise this to the others.

mateuszbaran:
> All storages should be updated, the docs for the `StoreStateAction` also document (at least the easier) keyword possibilities.

Thanks!

> The only thing we could think about is a constructor without a manifold that only uses the old storage, but I am not sure how useful that is, for now we have a manifold at most places, just that M is not needed in the constructor if we only use the old storage type.

Feel free to suggest a constructor signature for that :slightly_smiling_face: .

> > The new Storage does not work for manifolds like the Circle it seems.
> 
> One trick I used for now in one place – if it is not a default manifold, I‘ll check if `rand(M)` is a number and then only use the old storage as well (I only tested one of the storages with Circle). We could generalise this to the others.

I will move that trick to the constructor so that you don't have to think about it when constructing storage.

kellertuer:
Actually the old one works fine for the old storage, that is `StoreStateAction( Symbol[])` works and for me looks reasonable for still using the old storage.
This should just be documented as well maybe.

Cool, moving that check with number-Manifolds to something internal is for sure better :)

So besides docs we are only missing a few test cases.

kellertuer:
The only line missing is from `extract_type_from_namedtuple` that it returns any at the end. Could you check for that? With the rest I am fine :) (I even tried to find a more stable solution in the EPM test that sometimes failed).

mateuszbaran:
Yes, I will add a test for that line. I also need to take a closer look at performance again.

kellertuer:
Perfect, then we just have to add a small comment in the (now new) Changeling as well.

kellertuer:
I had to add my two PRs of the next version anyways, so we just have to add the date when we merge.

mateuszbaran:
OK, I will change `tangent` to `vector`. The new changelog looks nice :slightly_smiling_face: .

kellertuer:
Besides the name (you called it News.md) it is heavily inspired by your approach. I think for breaking changes especially it is important to document what broke and how to fix it.
Same for our other packages, let‘s introduce that step by step.

kellertuer:
Hm, the cleanup might have caused a bit of code coverage loss. Maybe the Nonmonotonelinesearch does not have a proper storage for the case that is not covered (and hence stops early with no change)? I will try to check your recent changes for that maybe the early evening (sorry busy until then).

mateuszbaran:
I think I know what might be the problem.

kellertuer:
Nice fix! I am not 100% sure we use `PointStorageKey` and `VectorStorageKey` already in all places where they are necessary.

mateuszbaran:
Thanks! Indeed, they are not used in all places where they should be. I will adapt all places except `primal_dual_plan.jl` where I'm not sure about it.

kellertuer:
Sure, primal-dual plan can stay “slow” for now, since it is quite a bit of work required there which we can do later.

kellertuer (approved):
For me this looks nice now, thanks for all the work.

Before merging, just write a date behind 0.4.8 in the Changelog.

I will write a tutorial that hopefully illustrates the benefits of using storage and even the fast storage as well as how to use them.

mateuszbaran:
Thanks, I'm still going to test this storage a bit more and make an example that illustrates its benefits.

src/solvers/quasi_Newton.jl
```diff
@@ -49,6 +49,7 @@ mutable struct QuasiNewtonState{
     retraction_method::RTR
     stepsize::S
     stop::SC
+    p_old::P
```
kellertuer, line 52: Hm, is it too much effort to use a storage here? In (nearly) all other places where we need the last iterate, we use a storage.

mateuszbaran: No, not too much effort, I just wanted to keep the code simple (and with storage, it's updating is still not intuitive for me). I can change it to storage. Anyway, more storage is needed but I only wanted to make a quick small improvement here.

kellertuer: I have not yet checked how much that affects everything, but the tests do something different now since WolfeLineSearch now has one case that is not covered.

Not that storing `p_old` here would store the last iterate double (once in the storage once here) even if we share a storage among all things.

Sure the updatey thing is a bit difficult and should be simplified to just be called at the end of the step in `solve` I think (though I have to think about this a bit more).

I only feel that we spent so much time on storage that on these last few steps omitting that feels a bit - strange?

src/plans/quasi_newton_plan.jl
```diff
@@ -478,14 +478,15 @@ function (d::QuasiNewtonLimitedMemoryDirectionUpdate{InverseBFGS})(mp, st)
...
+        r .-= d.ξ[i] .* d.memory_y[i]
     end
     r .= 1 / (d.ρ[m] * norm(M, p, last(d.memory_y))^2) .* r
     for i in 1:m
-        r .= r .+ (d.ξ[i] - d.ρ[i] * inner(M, p, d.memory_y[i], r)) .* d.memory_s[i]
+        r .+= (d.ξ[i] - d.ρ[i] * inner(M, p, d.memory_y[i], r)) .* d.memory_s[i]
     end
-    d.project && project!(M, r, p, r)
-    return -r
+    d.project && embed_project!(M, r, p, r)
```
kellertuer, line 487: I am also not sure here. Besides that we currently have 2 different forms of specifying this (I think in TCG we store a function here we have a bool) I am not sure why this switch was necessary. In the old code was no embed. Can you help me understand why that was necessary? This line is only meant to introduce stability of the algorithm.

src/solvers/quasi_Newton.jl
mateuszbaran: Sadly, after all this work, I still don't see any real advantage of storage over a plain variable. Only centralized storage and updating would bring an advantage but it's still a lot of work to have it.

Plus, it's unclear to me if we even can have a single point of updating.

src/plans/quasi_newton_plan.jl
mateuszbaran: Simply, `project` is not idempotent. For example on `Rotations` it isn't. `embed_project!` is idempotent. In generic code you shouldn't call `project` before  you actually `embed` a point or tangent vector.

src/solvers/quasi_Newton.jl
kellertuer: Sure the updating is still a thing, that might not be so well-modelled and might still need work. But if we use storage in all places (even if for now those are separate copies) the step to centralise it is easier / shorter than to look though all code again whether we store some p_old – and sure here it was temporary in the code so having that not in there but in the state is beneficial.

Overall the goal is of course to just store p_old once (for debug/stop/step/stepsize...) instead of multiple copies.
The alternative is of course to have a lot of memory and allocations. But then we can omit this PR? Not my plan.

src/plans/quasi_newton_plan.jl
kellertuer: Hm, then most of my code is probably wrong :(

mateuszbaran: :(

kellertuer: So just historically I up to the discussion we had and this extra function we use – have used `project!` in several places for stability.

src/solvers/quasi_Newton.jl
mateuszbaran: Hm, I've tried converting that to storage and I can't make it look nice. I will just revert back to allocating code I think.

kellertuer: Sure, maybe we should move this to the point where we have optimised the update (I have a few ideas but have to think about that as well) – something with a storage that keeps track of its “users”, they report that they are “ready to update” and if the last one reports that, the storage updates.

mateuszbaran: In the meantime it would be nice to have "just update this thing" that looks nicer than
```julia
copyto!(M, get_storage(qns.storage, PointStorageKey(:p_old)), get_iterate(qns))
```
(and I'm not even sure if that's intended interface or not, but that's my best guess)

kellertuer: currently there is `update_storage!(a,s)` ( see https://github.com/JuliaManifolds/Manopt.jl/blob/abedeeda37adc58a35e667bdbf5af24ccb413cca/src/plans/solver_state.jl#L264 ) but that of course updates the whole storage in the `AbstractStateAction a` (read as: Debug/Step/Stop/Record...) to the values given in the `State s`.
So even my first idea of putting a storage in a state `s´ does not directly work (as I proposed above).

But sure this could be reduced to have a `set_storage!(a, key, value)` that updates a key in the storage to value. With your pinky and vector key types this could even be used for those.
But you are right for now this does not exist, since I designed it to just “update all at the end.”
On the other hand no one than me used the code before, maybe not even looked at it, so it is definetly not an optimal design, for sure.

mateuszbaran: Yes, `set_storage!(a, key, value)` would be very reasonable but let's leave it for another PR. I won't put more changes here.

kellertuer: yes, I think that should be combined with making the storage (of different places) more interoperable as well and that's more for another PR.

Issue #220, opened 2023-02-22
Thoughts on improving state storage (even further)


kellertuer:
Recently there were two great improvement PRs on the state storage and quite a few discussions. I thought a bit about the status and how to even improve this further. I would like to summarise this here and look forward to how this can be improved even further.

# Motivation

the initial motivation comes from the fact that there are several “things” that are not always required, and if they are not required they maybe should not be stored to spare memory.
The easiest example is maybe the gradient descent algorithm, if the just fix

* a max number of iterations as a stopping criterion
* a constant stepsize

(not the best of algorithms, merely to illustrate something). Then the algorithm does not need to store the last iterate nor the last step. This would only be required if we check for the last change (e.g. as a stopping criterion) or store the last gradient to check for a gradient change.
Even the gradient evaluation does not need to be saved, only if the need it in more places then the `exp`.

So with this it would be great to have a “flexible storage” where several components can store stuff or request stuff to be stored if necessary.

# Current State: Decentralized
Currently there is the `StateStorageAction` which is storage attached to `Actions` that is e.g. stopping criteria, stepsize rules, debug output or record actions.

Usually these are currently mainly independent storages, though the initial design was to also have a common one that is passed to all them. The challenge here is the update, which currently might have negative side effect.

## Example
a debug printing the change and a stopping criterion using the change both require the last iterate. If they share a storage for that, one might update the storage before the other one uses the last iterate and hence print a change of 0 afterwards

## Possible Solution
We could “double” the storage which 
1. every current “update” would write into a “shadow storage”
2. one always _reads_ from the “front storage”
3. at the end of an iteration these two storages could be switched. This avoids the side effects mentioned, but is a little technical

## Advantage
One relatively large advantage that I see in this approach is, that anyone implementing a new solver does not have to think about these things, since they are decentralised

## Disandvantage
Getting the communication to work properly (see possible solution) is a bit technical I fear

# Possible other Solution
Instead of having individual storages we could do one central storage in all `AbstractManoptSolverState`s. That way the update would just happen at the end of a step, which might be a bit complicated considering that the decorators usually act _after_ a step is performed, but at least this would lead to a central place where the “update” mentioned above would be performed

On the other hand all actions mentioned above have the state in their functor, so accessing the central storage should not be a problem. 

## Challenge: Flexibility

One challenge in this approach might be flexibility, that is when generating the state, it might not be 100% known which actions (again thing stop/step/debug/record in the future maybe count) use the storage.

[...]

PR #221, opened 2023-02-22
LineSearches.jl wrapper and minor improvements

mateuszbaran:
I've made a wrapper around LineSearches.jl so that is can be used from Manopt now. I think this should be relatively safe for most manifolds.

Also, a couple of minor improvements for discussion. Most notably I've brought back local temporary storage in quasi-Newton. There is no easy way to convert it to `StoreStateAction` and I really don't want to work on it for now. On the plus side, quasi-Newton is now faster on functions that are cheap to evaluate. The main issue with the Rosenbrock-on-sphere example is now the cost of vector transports. That's one of the reasons why I capped the default memory size (I think that's reasonable given how the algorithm works?).


kellertuer:
this looks quite nice! And cool that we are faster (for other functions maybe Caching might be needed?)

Sure as I mentioned in #220 at some point we should do a revision of the state store action (probably as a struct in a state) but that is some major rework and (a) nothing I want to do now and (b) should be carefully discussed before to not waste time. So I think that is fine to do this tradeoff for now.

Formally changing the default is breaking, I think this default is reasonable though, but it should be documented of course.

We would need some more docs that one can use Linesearches.jl now (if that is efficient we can even remove a few here, since I do not like the Wolfe-searches we did or tried to implement). 

tldr: looks good, needs docs ;)

One thing we could check is whether we want to transition for Linesearches already to also alloww for the 1.9 weak dependency approach. I was that Chris had a way of allowing for both ways to connect the packages. Just if you want to give that a try.

src/linesearches_compat.jl
```diff
@@ -0,0 +1,65 @@
...
+    fp=get_cost(mp, get_iterate(s)),
+    kwargs...,
+)
+    M = get_manifold(mp)
+    p = get_iterate(s)
+    p_tmp = copy(M, p)
+    X_tmp = copy(M, p, η)
+    Y_tmp = copy(M, p, η)
+    f = get_cost_function(get_objective(mp))
+    function ϕ(α)
```
kellertuer, line 41: These three functions could also be a functor/struct as a field in the wrapper struct? (Hope this idea is not sketched too short but I am thinking similar to AugmentedLagrangianCost & such).

mateuszbaran:
> this looks quite nice! And cool that we are faster (for other functions maybe Caching might be needed?)
> 
> Sure as I mentioned in #220 at some point we should do a revision of the state store action (probably as a struct in a state) but that is some major rework and (a) nothing I want to do now and (b) should be carefully discussed before to not waste time. So I think that is fine to do this tradeoff for now.

Yes, for functions that are more expensive caching would be a good idea. And thanks for agreeing that we can store things outside of `StoreStateAction` for now :slightly_smiling_face:.

> Formally changing the default is breaking, I think this default is reasonable though, but it should be documented of course.

Sure, I will add docs.

> We would need some more docs that one can use Linesearches.jl now (if that is efficient we can even remove a few here, since I do not like the Wolfe-searches we did or tried to implement).

I haven't done much comparing between Manopt and LineSearches, for this PR I only intend to add the wrapper and leave existing line searches. I will add some docs after I resolve (or at least understand) a couple of weird issues with projection VT.

> One thing we could check is whether we want to transition for Linesearches already to also alloww for the 1.9 weak dependency approach. I was that Chris had a way of allowing for both ways to connect the packages. Just if you want to give that a try.

I can try that, maybe it wouldn't be too much work.

src/linesearches_compat.jl
mateuszbaran: I will try to modify it so that temporaries are only allocated once per optimization, so this is just a first sketch anyway.

kellertuer: Sure, only do that if this works. And I think the same idea as with structs is the way to go afterwards – but since I learned that approach with functors from you, I am sure you know what to do.

mateuszbaran: Hm, this is complicated. I can't easily allocate these buffers at the start of optimization. The easiest way would be by adding a function that is called during state creation that asks line search what kind of buffer it needs, puts it in the state, and allows for access by line searches using another new function. Does that sound reasonable?

I've considered converting these three functions to functors and it seems that it will only make the code unnecessarily convoluted in this case. Functors are great for encapsulating what would otherwise be global state, but accessing local state (here: variables in the outer function) rarely causes performance issues.

kellertuer: I think it would avoid the allocations for X_tmp and Y_tmp though because that would be fields in phi_struct (or in the line search wrapper)

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
+
```
kellertuer, line 1: Maybe it would be a good time to start a `src/compat` folder?

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+    LineSearchesStepsize(
+        M::AbstractManifold,
+        ls,
+        is;
+        retraction_method::AbstractRetractionMethod=default_retraction_method(M),
+        vector_transport_method::AbstractVectorTransportMethod=default_vector_transport_method(M),
+    )
+    LineSearchesStepsize(
+        ls,
```
kellertuer, line 17: I would prefer speaking names and if possible types (even if its the `Optim.` ones.
This is `line_search`? the next is `initial_step`?

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+        is;
+        retraction_method::AbstractRetractionMethod=default_retraction_method(M),
```
kellertuer, line 19: This constructor does not have an `M` so this line and the next are not correct.

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+    ),
+)
+    return LineSearchesStepsize(
+        ls,
+        is;
+        retraction_method=retraction_method,
+        vector_transport_method=vector_transport_method,
+    )
+end
+function LineSearchesStepsize(
```
kellertuer, line 52: At some point Manopt should decide on maybe skipping these, if the sell it as „we need M for defaults“

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+        vector_transport_method::AbstractVectorTransportMethod=default_vector_transport_method(M),
+    )
+    LineSearchesStepsize(
+        ls,
+        is;
+        retraction_method::AbstractRetractionMethod=default_retraction_method(M),
+        vector_transport_method::AbstractVectorTransportMethod=default_vector_transport_method(M),
+    )
+
+Wrap linesearch `ls` (for example `HagerZhang` or `MoreThuente`) and initial step selector
```
kellertuer, line 23: Is there something we can link to? Is it `Optim.HagerZhang()`? Then maybe that also helps?

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+    return min(x, y)
+end
+
+# code below was adapted from https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/initialguess.jl
+
+function get_initial_alpha(
+    M::AbstractManifold, is::InitialStatic{T}, state::StepsizeStorage, p, η, phi_0, dphi_0
+) where {T}
+    PT = promote_type(T, real(number_eltype(η)))
+    if is.scaled == true && (ns = real(norm(M, p, η))) > convert(PT, 0)
```
kellertuer, line 163: So this case never happened, so why do we have it?

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+"""
+function nanmin(x, y)
+    isnan(x) && return y
+    isnan(y) && return x
+    return min(x, y)
+end
+
+# code below was adapted from https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/initialguess.jl
+
+function get_initial_alpha(
```
kellertuer, line 159: this is basically an internal `get_initial_stepsize`? Or the `initial_guess`from my Armijo implementation? I am not sure.

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+    PT = promote_type(T, real(number_eltype(η)))
+    if is.scaled == true && (ns = real(norm(M, p, η))) > convert(PT, 0)
+        # TODO: Type instability if there's a type mismatch between is.alpha and ns?
+        state.alpha = convert(PT, min(is.alpha, ns)) / ns
+    else
+        state.alpha = convert(PT, is.alpha)
+    end
+end
+
+function get_initial_alpha(
```
kellertuer, line 171: Please add a documentation – can this be used outside of Linesearches.jl Linesearches? If so we should fine a better name.

src/linesearches_compat.jl
```diff
@@ -0,0 +1,234 @@
...
+    ::AbstractManifold, is::InitialPrevious, state::StepsizeStorage, p, η, phi_0, dphi_0
+)
+    if isnan(state.alpha)
+        state.alpha = is.alpha
+    end
+    state.alpha = min(is.alphamax, state.alpha)
+    return state.alpha = max(is.alphamin, state.alpha)
+end
+
+function get_initial_alpha(
```
kellertuer, line 181: also here – no clue what this does, also there are comments in the code that read like todos or issues. I would prefer to not have such comments.

src/plans/stepsize.jl
```diff
@@ -959,3 +959,53 @@ function get_last_stepsize(
...
+stepsize_storage_points(::Stepsize) = Tuple{}
+
+stepsize_storage_vectors(::Stepsize) = Tuple{}
+
+"""
+    StepsizeStorage{TS<:Stepsize,TStorage<:StoreStateAction}
+
+Storage for stepsize kept between iterations of a solver.
+"""
+mutable struct StepsizeStorage{TS<:Stepsize,TStorage<:StoreStateAction,T<:Number} <:
```
kellertuer, line 974: While I like this – what is stored here?

alpha and last_f_p (is that the last cost?) could be stored in the wrapper – they are stored there for the usual step sizes anyways.

src/solvers/quasi_Newton.jl
```diff
@@ -37,18 +37,20 @@ mutable struct QuasiNewtonState{
     T,
     D<:AbstractQuasiNewtonDirectionUpdate,
     SC<:StoppingCriterion,
-    S<:Stepsize,
+    S<:StepsizeStorage,
```
kellertuer, line 40: I think this is breaking? Can Quasi Newton still be used with the Manopt-Armijo?

src/solvers/quasi_Newton.jl
```diff
@@ -37,18 +37,20 @@ mutable struct QuasiNewtonState{
...
     p::P
+    p_old::P
     X::T
     sk::T
     yk::T
     direction_update::D
     retraction_method::RTR
     stepsize::S
     stop::SC
+    new_grad::T
```
kellertuer, line 53: can we stay consistent and use `p_old` and `X_old` and not mix old and new? I would be in favour of that.

src/solvers/quasi_Newton.jl
```diff
@@ -284,24 +289,24 @@ end
...
+        M, qns.sk, qns.p_old, αη, qns.p, get_update_vector_transport(qns.direction_update)
     )
     vector_transport_to!(
-        M, qns.X, p_old, qns.X, qns.p, get_update_vector_transport(qns.direction_update)
+        M, qns.X, qns.p_old, qns.X, qns.p, get_update_vector_transport(qns.direction_update)
     )
-    new_grad = get_gradient(mp, qns.p)
-    qns.yk .= new_grad ./ β .- qns.X
-    update_hessian!(qns.direction_update, mp, qns, p_old, iter)
+    get_gradient!(mp, qns.new_grad, qns.p)
```
kellertuer, line 307: yes, calling this `X_old` would require a copy here, I see; I would still prefer a unified naming.

kellertuer (commented):
In the coverage it seems we lost `last_stepsize` for Wolfe being covered?

Coverage for the adapted code (especially the one case I do not understand and a rethrow) is complicated for sure. Is that really necessary to adapt and can not be called in the original form (wrapped)?

src/plans/stepsize.jl
kellertuer: What I do not like of this struct for now is – introducing it here says: This is a Manopt-thing. But it is just used in the wrapper and not in the Stepsizes of Manopt. So is it a Manopt-thing?
Why is it then only required in QN that one uses it? I would have expected the old ones to work without this?

src/linesearches_compat.jl
mateuszbaran: Good idea, I will make it.

src/linesearches_compat.jl
mateuszbaran: Yes, I will expand these names.

src/linesearches_compat.jl
mateuszbaran: Sure, I will add a link to docs.

src/linesearches_compat.jl
mateuszbaran: The nice thing about this constructor is that we don't need a manifold to set non-default methods. Would you prefer to implement this differently?

src/linesearches_compat.jl
mateuszbaran: `StepsizeStorage` should allocate these temporaries but I forgot to redo this part.

src/linesearches_compat.jl
mateuszbaran: This is specific to LineSearches.jl, though maybe after another rework it could also be used by other Manopt.jl line searches? I haven't checked. It is similar to `armijo_initial_guess`  but independent of a particular line search method.

src/linesearches_compat.jl
mateuszbaran: That's copied from LineSearches.jl: https://github.com/JuliaNLSolvers/LineSearches.jl/blob/ded667a80f47886c77d67e8890f6adb127679ab4/src/initialguess.jl#L17 . I haven't checked in what cases it is useful.

src/linesearches_compat.jl
kellertuer: Ah you are right, the ones I did (with just kwargs) do `M=DefaultManifold()` but that is the same as having this one if there is more than just the Manifold-parameter. Then we can keep this for sure.

src/linesearches_compat.jl
kellertuer: Do we need to copy it or is there – in the idea of a wrapper that we follow here – a wrapper also possible?

src/linesearches_compat.jl
kellertuer: ah ok, so in the long run it might. because yes it is similar to the initial guess one – but Nonmonontone&Wolfe (in Moat) do not have that (yet) for the others this does not make sense (e.g. the constant stepsize)

src/linesearches_compat.jl
mateuszbaran: It's not an exact copy. In this case I've replaced a standard `norm` with tangent vector norm, so a wrapper would have to translate that part as well. The code is very short so I think it's easier to copy it and change a little instead of making an elaborate wrapper.

kellertuer: Ok, and we even might have to keep the name then, hm.

src/linesearches_compat.jl
kellertuer: Probably something with linesearch and initial guess?

Maybe
`get_initial_stepsize_guess`?

src/plans/stepsize.jl
mateuszbaran: This is intended to store things stepsize-related information between iteration of an optimization algorithm. Right now it is only used for LineSearches.jl but I expect that it would also make sense to use it in "native" Manopt stepsize algorithms. It is not QN-specific but it's the one solver where I cared about using LineSearches.jl and `StepsizeStorage` is needed for those.

Alpha stores stepsize from the previous line search and last_f_p stores value of objective at the point from which the last line search started (that's what Optim.jl does). They could be stored in the wrapper, I put it here because its something stored and we need `StepsizeStorage` for p_tmp, X_tmp and Y_tmp anyway.

src/solvers/quasi_Newton.jl
mateuszbaran: What name would you prefer? I used the name of variable in `step_solver!` that was used for the same purpose.

src/linesearches_compat.jl
mateuszbaran: That sounds fine :slightly_smiling_face: .

src/solvers/quasi_Newton.jl
mateuszbaran: This is only slightly breaking, someone would have to use the inner (not the outer) constructor to have their code broken by this change, or dispatch on type arguments. And there doesn't seem to be a good reason to do that. It still works for Manopt line searches because by default `StepsizeStorage` just forwards to the wrapped linesearch (only LineSearches.jl compat layer has custom behavior).

src/plans/stepsize.jl
kellertuer: then the fields should be called `last_stepsize` and `initial_cost`.

I think the idea is a good one since we have the storage idea anyways and sure others could use that as well later.

src/solvers/quasi_Newton.jl
kellertuer: I do not like the asymmetry between `X` (above) for the current gradient) and `_grad` here, not that we have `p_old` and `new_` (neither the switch in words nor the old/new usage). So in style I would prefer `X_old`.

src/solvers/quasi_Newton.jl
kellertuer: But this does mean I can not use the Manopt line searches here? I feel that is strange.
I would prefer keeping this to be `StepSize` since I feel the storage is a decorator and hence also a stepsize? Why is that not possible? For Record/Debug in State for example it is exactly the way I wrote.

Having here a type that does not work with the line searches from this package but pony from the compat – that is a bit strange.

src/solvers/quasi_Newton.jl
```diff
@@ -77,15 +79,18 @@ function QuasiNewtonState(
     VTM<:AbstractVectorTransportMethod,
 }
     sk_init = zero_vector(M, p)
-    return QuasiNewtonState{P,typeof(sk_init),D,SC,S,RM,VTM}(
+    stepsize_storage = StepsizeStorage(M, stepsize; p_init=p, X_init=initial_vector)
```
kellertuer, line 82: ...for example why does this have to be done here? If I pass a stepsize into the low level function I do not want it „decorated“ that should happen before. I would assume that the new Linesearches can do that themselves?
This also seems the only change with the new field

Let‘s keep it simple here and allow all step sizes. So I would prefer if the LineSearch.jl Stepsiyzes “wrap themselves” into storage if they need to. I would do the same with the Manopt ones (a constant stepsize never need _any_ storage for example).

Also that way one can share the storage later. If it is always only generated here in that is never possible.

src/solvers/quasi_Newton.jl
mateuszbaran: Sure, I can leave the type here as `Stepsize`.

src/solvers/quasi_Newton.jl
mateuszbaran: I can only create storage when I have manifold, point and vector. I don't have them earlier and forcing users to supply them to create a stepsize would be awkward. Anyway, this goes in the general direction of sharing storage much more than having storage completely separate: search state creation "asking" different things it relies on what kind of storage they need and merging them into one.

src/solvers/quasi_Newton.jl
mateuszbaran: I've changed name to `X_old`.

src/plans/stepsize.jl
mateuszbaran: I've updated this.

src/linesearches_compat.jl
mateuszbaran: I have made an `/ext` folder as suggested in Julia docs: https://docs.julialang.org/en/v1.10-dev/manual/code-loading/#man-extensions .

src/linesearches_compat.jl
mateuszbaran: fixed

src/linesearches_compat.jl
mateuszbaran: Done.

src/solvers/quasi_Newton.jl
kellertuer: We provide them to all other „small“ structures (debug/record) and they all just recently got a manifold.

But doing the storage creation here and leaving all old ones as „pass through“ is completely the opposite of the current modular design _and_ will be a lot of work when trying to work further on storage, since then we (or I) have to fix two completely different modes of how storage is handled.

And this does not go into another direction – storage is a decorator – ok, so is debug (having a storage internally), but it is still separate storage (not in state but in the stepsize now), so this is kind of the two disadvantages of bot approaches combined: It is modular storage (not central) that _never_ can be shared. Then you Can also just introduce 3 fields in the Stesize.

And ArmijoStepsize for example _has_ a manifold, so that is also not strange.

src/solvers/quasi_Newton.jl
```diff
@@ -284,24 +289,24 @@ end
...
+        M, qns.sk, qns.p_old, αη, qns.p, get_update_vector_transport(qns.direction_update)
     )
     vector_transport_to!(
-        M, qns.X, p_old, qns.X, qns.p, get_update_vector_transport(qns.direction_update)
+        M, qns.X, qns.p_old, qns.X, qns.p, get_update_vector_transport(qns.direction_update)
     )
-    new_grad = get_gradient(mp, qns.p)
-    qns.yk .= new_grad ./ β .- qns.X
-    update_hessian!(qns.direction_update, mp, qns, p_old, iter)
+    get_gradient!(mp, qns.X_old, qns.p)
```
kellertuer, line 307: hm, this now looks like you use `X_old` for a new gradient? could we then maybe switch the roles of X and X_old? This way it is confusing, I did not just mean to call new_grad->X_old

src/solvers/quasi_Newton.jl
kellertuer: and you are correct my storage design is not the best of choices. But introducing now in different step sizes and different solvers new ways to store stuff makes the whole package a little too chaotic, unless all models used are in detail documented. And I don‘t want to document 2-3 (or 4 or 5 or how many we prototype this way then) motivations on how to do storage.

I would prefer we follow the current way of storage here as well until we have a nice (maybe state-central) way of doing so.

src/ext/LineSearchesExt.jl
```diff
@@ -0,0 +1,278 @@
...
+    )
+
+Wrap `linesearch` (for example [`HagerZhang`](https://julianlsolvers.github.io/LineSearches.jl/latest/reference/linesearch.html#LineSearches.HagerZhang)
+or [`MoreThuente`)(https://julianlsolvers.github.io/LineSearches.jl/latest/reference/linesearch.html#LineSearches.MoreThuente))
+and initial step selector `initial_stepsize` that will work on manifold `M`. Retraction used
+for determining the line along which the search initial_stepsize performed can be provided
+as `retraction_method`. Gradient vectors are transported between points using
+`vector_transport_method`.
+"""
+struct LineSearchesStepsize{
```
kellertuer, line 30: maybe as my interpretation for the storage. This struct would have a storage field (not be wrapped in a storage), the constructor with M can use the fast storage the other one can not. That is the analogue of what other storage things currently do.

Maybe yours is a bit faster that you have here, but I do not see all implications (and do not want to rewrite all high-level interfaces at the moment) and would like to avoid having multiple solutions for one thing. Similarly I do not like that the solver had to be rewritten at all – it uses a stepsize, it does not have to know the internals of said stepsize

src/solvers/quasi_Newton.jl
mateuszbaran: I'm a bit tired fighting with the storage design and I most likely won't need Manopt-LineSearches compatibility so I think I won't be finishing this for some time.

mateuszbaran: I understand your position on my design, and I agree that storage should be consistent. LineSearches.jl didn't improve the performance as I expected it would, sadly.

Another option for finishing this PR is removing the initial step size selection part. What would remain wouldn't require any storage so there would be no design issues and you could think later about a way to extend initial step size choice to existing Manopt line searches as well.

kellertuer: Ah that also sounds like an idea to finish this, yes. That would be less work at least. Maybe I'll go for that.then when I find time. You are right, its nice to have these new line searches available here.


And sure I can understand your tendency to a new storage, that is why we already have an issue for that :) just that this time I want to spent more thinking about the design first.

src/ext/LineSearchesExt.jl
kellertuer: This is a good tradeoff I think. We could aa a “not supported _yet_” here, since with a proper storage we would do that.

src/linesearches_compat.jl
kellertuer: Do we still wanna store the cantons/functionals in the wrapper struct? Or do you think that does not have much of a advantage allocation wise.

src/ext/LineSearchesExt.jl
mateuszbaran: Sure, feel free to update the docs if you think so.

src/linesearches_compat.jl
mateuszbaran: It has no advantage whatsoever at the moment as far as I can tell. The struct-based code would even have to be fairly convoluted to not be *worse* than the current solution allocation-wise.

kellertuer: ok, then lets leave this as is :)

src/ext/LineSearchesExt.jl
kellertuer: Sure will check that later today hopefully.

kellertuer:
Could you check for the last line we miss and whether that is your fault?

I will check and update the docs a bit hopefully later today.

mateuszbaran:
Yes, the last uncovered line is rethrowing error in LineSearches.jl. I will try to add a test for it.

kellertuer:
Is there a good place we could mention this extension? Maybe a docs page mentioning extensions listing this one?

mateuszbaran:
I've added a small docs page with an example in documentation.

src/solvers/quasi_Newton.jl
kellertuer: could you check for this naming? Currently it might really to be too confusing, since we use a `X_old` field for a new gradient.

mateuszbaran: I think I might have misunderstood you, you just want to switch `X` and `X_old`, right? I can do that.

kellertuer: But then be careful when someone accesses `get_gradient` what to return ;)

mateuszbaran: Then maybe it would be easier to change `X_old` to `X_new`?

kellertuer: That would be a bit inconsistent to what we do in other solvers, since we usually only store `X_old` and not `X_new`, but if you do not see how to best rephrase this, then ok take `X_new`.

mateuszbaran: I've changed usage of `X_old` and `X` a bit and I think it's somewhat consistent now. What do you think?

kellertuer:
This looks good so far, let's wait for the tests as a last point :)

kellertuer (approved):

kellertuer:
If you could just bump version. Then we can merge.

Issue #232, opened 2023-03-31
NaNs in constrained optimization


projekter:
I'm currently considering a small optimization problem with constraints (yes, there are no special manifolds involved - but this is just one of my most simple test instances):

```julia
using Manifolds, Manopt
obj(M, x) = -(x[1] - 1.)^2 - (x[1] - x[2])^2 - (x[2] - 3.)^2
∇obj(M, x) = [-2. * (x[1] - 1.) - 2. * (x[1] - x[2]), 2. * (x[1] - x[2]) - 2. * (x[2] - 3.)]
nonpos = [(M, x) -> (x[1] - 1.)^2 - 1.,
          (M, x) -> (x[1] - x[2])^2 - 1.,
          (M, x) -> (x[2] - 3.)^2 - 1.
         ]
∇nonpos = [(M, x) -> [2. * (x[1] - 1.), 0.],
           (M, x) -> [2. * (x[1] - x[2]), -2. * (x[1] - x[2])],
           (M, x) -> [0., 2. * (x[2] - 3.)]
          ]
```
The constraints define a triangle with edges `(1, 2)`, `(2, 2)`, and `(2, 3)`. The minimal objective value is `-2`, and it is attained precisely at those edges.
Manopt provides two algorithms for constrained optimization. As far as I can see, neither the augmented Lagragian method nor the exact penalty method require an initial feasible point, which is great, since for general complicated problems, I can't easily provide one without a preceding Phase I algorithm. Indeed, even the [tutorial on constrained optimization](https://manoptjl.org/stable/tutorials/ConstrainedOptimization/) uses an initially infeasible point.
So let us just for the sake of reproducibility start with the infeasible `(0, 0)`.
```Julia
julia> augmented_Lagrangian_method(Euclidean(2), obj, ∇obj, zeros(2), g=nonpos, grad_g=∇nonpos)
2-element Vector{Float64}:
 NaN
 NaN
```
This looks bad, what is going on?
```Julia
julia> augmented_Lagrangian_method(Euclidean(2), obj, ∇obj, zeros(2), g=nonpos, grad_g=∇nonpos, debug=[:Iteration, " | ", :Cost, " | ", :Iterate, "\n", :Stop])
Initial  | F(x): -10.000000 |
# 1      | F(x): -2.037791 | p: [0.9813650895161447, 1.990621289079651]
# 2      | F(x): -3.954575 | p: [2.49807785767228, 1.8589897056867168]
[...]

PR #233, opened 2023-04-04
Messages to display warnings / Improve linesearch and L-BFGS

### LINKED ISSUE #229: Improve line searches
currently linesearches like `ArmijoLinesearch` but also the Wolfe ones and the NonMonotone one are a bit unstable for the simple reason that the involved `while` loops in the common `linesearch_backtrack` is maybe a bit too academic and might have an infinite loop.

Ideas
* allow for a minimal stepsize when to stop (already partly available but not yet set as a default=
* allow for a maximal number of steps
  *  (Linesearches.jl has 1000 [src](https://github.com/JuliaNLSolvers/LineSearches.jl/blob/ded667a80f47886c77d67e8890f6adb127679ab4/src/backtracking.jl#L78)
  * Manopt/Matlab has 25 [src](https://github.com/NicolasBoumal/manopt/blob/b8b54a6af8b965f7ae572972ba0d15787427744b/manopt/solvers/linesearch/linesearch.m#L107) works on a cost evaluation budget)
* check whether we hit NaN/Inf as LineSearches [src](https://github.com/JuliaNLSolvers/LineSearches.jl/blob/ded667a80f47886c77d67e8890f6adb127679ab4/src/backtracking.jl#L64)

And of course one question is then what to do – LineSearches errors, while Manopt just continues after 25
and if the step is too small it is also a question what to do
* error
* continue as is
* have a stopping criterion that stops in that case and reports that problem as a stopping reason (a non-convergence indicating of course)

kellertuer:
This PR aims to resolve #229 and #232.
It introduces a small way to obtain messages from sub-steps, e.g. a linesearch or the update of a Hessian – and their display can be activated with a debug action.

If @mateuszbaran and @projekter, you both would have a moment of time to take a look.

The challenge was, that most things in Manopt.jl are quite modular. So I came up with a (warning) message system.

# Todo
* [x] `get_message` still needs to be transparent w.r.t. decorators of the state
* [x] what other places might be nice to have messages?


src/plans/nonmutating_manifolds_plans.jl
```diff
@@ -14,31 +14,58 @@ end
...
+    while fNew < f0 + decrease * s * real(inner(M, p, η, gradF)) # increase
+        i = i + 1
         s = s / contract
+        p_new = retract(M, p, η, s, retr)
+        fNew = F(p_new)
+        if i == max_increase_steps
+            msg = "Max increase steps ($(max_increase_steps) reached"
+            break
+        end
+        if s > stop_when_stepsize_larger
```
mateuszbaran, line 43: Be careful -- `max_stepsize` returns the maximum length of the vector we can go in one step, so I think this should be something like `s * norm(M, p, η) > stop_when_stepsize_larger`

mateuszbaran (commented):
Nice, I think this is a really good improvement :+1: .

src/plans/nonmutating_manifolds_plans.jl
kellertuer: well-spotted. Will fix that.

kellertuer:
After thinking a bit, I will remove the message type again. It adds too much fuzz, and steering which type of report you want can be done anyways. If we want to switch message types we could still to that with different message types later (allowing more than just strings).

src/plans/nonmutating_manifolds_plans.jl
kellertuer: Since the parameter is called `when step_size_larger` I will prefer to scale the default number of said parameter instead

mateuszbaran: Sure, that's also fine.

kellertuer:
So besides

* [ ] documentation

the only other thing one could do is

* [ ] safeguards for the Wolfe searches.

But that would be quite some work.

kellertuer:
Checked Wolfe, they have the safeguards already (smallest stepsize and a binary search s its reached relatively quick).

src/plans/quasi_newton_plan.jl
```diff
@@ -467,21 +465,50 @@ function QuasiNewtonLimitedMemoryDirectionUpdate(
...
+get_message(d::QuasiNewtonLimitedMemoryDirectionUpdate) = d.message
+function status_summary(d::QuasiNewtonLimitedMemoryDirectionUpdate{T}) where {T}
+    s = "limited memory $T (size $(length(d.memory_s)))"
+    (d.scale != 1.0) && (s = "$(s) initial scaling $(d.scale)")
+    d.project && (s = "$(s), projections, ")
+    s = "$(s)and $(d.vector_transport_method) as vector transport."
+    return s
+end
 function (d::QuasiNewtonLimitedMemoryDirectionUpdate{InverseBFGS})(mp, st)
+    (length(d.message) > 0) && (d.message = "") # reset message
```
mateuszbaran, line 480:
```suggestion
    isempty(d.message) || (d.message = "") # reset message
```
`isempty` might be a bit nicer than computing length.

src/plans/quasi_newton_plan.jl
```diff
@@ -467,21 +465,50 @@ function QuasiNewtonLimitedMemoryDirectionUpdate(
...
     M = get_manifold(mp)
     p = get_iterate(st)
     r = copy(M, p, get_gradient(st))
     m = length(d.memory_s)
     m == 0 && return -r
     for i in m:-1:1
-        d.ρ[i] = 1 / inner(M, p, d.memory_s[i], d.memory_y[i]) # 1 sk 2 yk
+        # what if we dvide by zero here? Setting to zero ignores this in the next step
+        #precompute in case inner is expensive
```
mateuszbaran, line 488:
```suggestion
        # what if we divide by zero here? Setting to zero ignores this in the next step
        # precompute in case inner is expensive
```

mateuszbaran (approved):
I have a few small comments, everything else looks nice.

src/plans/quasi_newton_plan.jl
```diff
@@ -467,21 +465,50 @@ function QuasiNewtonLimitedMemoryDirectionUpdate(
...
+        v = inner(M, p, d.memory_s[i], d.memory_y[i]) # 1 sk 2 yk
+        if v ≈ 0
```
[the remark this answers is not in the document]
kellertuer: but then I would use `== 0`?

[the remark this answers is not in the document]
kellertuer: ...besides that I was hoping to just maybe finish it today, but now `Circle` is annoying me in about a trillion cases since I assume the max-stepize to be a float but circles infectivity radius is an irrational 🙄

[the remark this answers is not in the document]
mateuszbaran: Sure, `==0` is also effectively the same.

BTW, `pi` should be converted to a float very easily, it shouldn't cause that many issues?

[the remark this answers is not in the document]
kellertuer: Well, took me a bit to fix the pi.

And I switched to iszero, so I'd call this PR ready :)

PR #234, opened 2023-04-09
Improve particle swarm to work in-place

### LINKED ISSUE #186: Add in-place variant of particle swarm optimization
(as in the title)

kellertuer:
Resolves #186.

Currently this still allocates a tangent vectors and a point per iteration, which one could remove by also adding those to the state.
Was this what you were thinking of @mateuszbaran ?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,7 +5,15 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.4.13] - 06/04/2023
+## [0.4.14] - 06/04/2023
+
+### Changed
+* `particle_swarm` now uses much more in-place operations
+
+### Fixed
+* `particle_swarm` used quie a few `deepcopy(p)` commands still, which are now `copy(M,p)`
+
+## [0.4.13] - 09/04/2023
 
 ### Added
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `particle_swarm` now uses much more in-place operations
 
 ### Fixed
-* `particle_swarm` used quie a few `deepcopy(p)` commands still, which are now `copy(M,p)`
+* `particle_swarm` used quite a few `deepcopy(p)` commands still, which were replaced by `copy(M, p)`
 
 ## [0.4.13] - 09/04/2023
 
--- a/src/solvers/particle_swarm.jl
+++ b/src/solvers/particle_swarm.jl
@@ -13,6 +13,9 @@ Describes a particle swarm optimizing algorithm, with
 * `inertia` – (`0.65`) the inertia of the patricles
 * `social_weight` – (`1.4`) a social weight factor
 * `cognitive_weight` – (`1.4`) a cognitive weight factor
+* `p_temp` – temporary storage for a point to avoid allocaitons on the algorithm
+* `social_vec` - temporary storage for a tangent vector related to `social_weight`
+* `cognitive_vector` -  temporary storage for a tangent vector related to `cognitive_weight`
 * `stopping_criterion` – (`[`StopAfterIteration`](@ref)`(500) | `[`StopWhenChangeLess`](@ref)`(1e-4)`)
   a functor inheriting from [`StoppingCriterion`](@ref) indicating when to stop.
 * `retraction_method` – (`default_retraction_method(M, eltype(x))`) the rectraction to use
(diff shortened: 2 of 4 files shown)


mateuszbaran:
Yes, this is indeed a part of what I was thinking about. Adding `cognitive_infl`, `social_infl` and `xOld` to solver state would solve the issue completely.

kellertuer:
Sure that is merely one single thing to do and the only a bit complicated thing is initialising them (when creating the state) to the right type :) I will check.

src/solvers/particle_swarm.jl
```diff
@@ -259,35 +259,47 @@ function particle_swarm!(
...
+    cognitive_infl = zero_vector(M, first(s.p))
+    social_infl = zero_vector(M, first(s.p))
+    xOld = copy(M, first(s.p))
     for i in 1:length(s.x)
-        s.velocity[i] =
+        inverse_retract!(M, cognitive_infl, s.x[i], s.p[i], s.inverse_retraction_method)
+        inverse_retract!(M, social_infl, s.x[i], s.g, s.inverse_retraction_method)
+        # add v = inertia * v + cw*cog_infl + sw*soc_infl
+        # where the last two are randomly shortened a bit
+        copyto!(
```
mateuszbaran, line 286: This should be a bit faster using broadcasting, like
```julia
s.velocity[i] .= s.inertia .* s.velocity[i] .+ s.cognitive_weight .* rand(1) .* cognitive_infl .+ s.social_weight .* rand(1) .* social_infl
```

kellertuer: Oh, I actually _never_ know which one is faster here; because I always have FixedRankManifolds in mind where the tangent vector is a struct and then I do not know whether `.=` works. Are you sure it works?

mateuszbaran: I've spent quite some time ensuring that broadcasting works for all our tangent vectors, including the ones defined by structs. If it doesn't in some case then let me know and I will take a look.

Broadcasting is almost always faster than `copyto!`.

kellertuer: ok, will change that as well. First the 3 new fields and a bit of allocations / type inference when creating a state.

kellertuer:
That was faster than I thought, now all 3 are temporary fields in the state and documented as such.

kellertuer:
The ±10 lines are some wiggles in the Quasi Newton that I was not yet able to narrow down, and are untreated to this PR.

Changelog.md
```diff
@@ -5,7 +5,15 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.4.13] - 06/04/2023
+## [0.4.14] - 06/04/2023
+
+### Changed
+* `particle_swarm` now uses much more in-place operations
+
+### Fixed
+* `particle_swarm` used quie a few `deepcopy(p)` commands still, which are now `copy(M,p)`
```
mateuszbaran, line 14:
```suggestion
* `particle_swarm` used quite a few `deepcopy(p)` commands still, which were replaced by `copy(M, p)`
```

src/solvers/particle_swarm.jl
```diff
@@ -13,6 +13,9 @@ Describes a particle swarm optimizing algorithm, with
 * `inertia` – (`0.65`) the inertia of the patricles
 * `social_weight` – (`1.4`) a social weight factor
 * `cognitive_weight` – (`1.4`) a cognitive weight factor
+* `p_temp` – temporary storage for a point to avoid allocaitons on the algorithm
```
mateuszbaran, line 16:
```suggestion
* `p_temp` – temporary storage for a point to avoid allocations during a step of the algorithm
```

src/solvers/particle_swarm.jl
```diff
@@ -259,35 +266,40 @@ function particle_swarm!(
...
-            s.cognitive_weight * rand(1) .*
-            inverse_retract(M, s.x[i], s.p[i], s.inverse_retraction_method) +
-            s.social_weight * rand(1) .*
-            inverse_retract(M, s.x[i], s.g, s.inverse_retraction_method)
-        xOld = s.x[i]
-        s.x[i] = retract(M, s.x[i], s.velocity[i], s.retraction_method)
-        s.velocity[i] = vector_transport_to(
-            M, xOld, s.velocity[i], s.x[i], s.vector_transport_method
+            s.cognitive_weight * rand(1) .* s.cognitive_vector +
+            s.social_weight * rand(1) .* s.social_vector
```
mateuszbaran, line 293:
```suggestion
            s.cognitive_weight .* rand(1) .* s.cognitive_vector .+
            s.social_weight .* rand(1) .* s.social_vector
```

src/solvers/particle_swarm.jl
```diff
@@ -259,35 +266,40 @@ function particle_swarm!(
...
     M = get_manifold(mp)
+    # Allocate two tangent vectors
     for i in 1:length(s.x)
-        s.velocity[i] =
+        inverse_retract!(M, s.cognitive_vector, s.x[i], s.p[i], s.inverse_retraction_method)
+        inverse_retract!(M, s.social_vector, s.x[i], s.g, s.inverse_retraction_method)
+        # add v = inertia * v + cw*cog_infl + sw*soc_infl
+        # where the last two are randomly shortened a bit
+        s.velocity[i] .=
             s.inertia .* s.velocity[i] +
```
mateuszbaran, line 291:
```suggestion
            s.inertia .* s.velocity[i] .+
```

mateuszbaran (approved):
I just have a few minor comments, everything else looks fine :slightly_smiling_face:

Issue #235, opened 2023-04-07
Implement a `fill(p)` prequel for every solver, where `p` might be a number.


kellertuer:
Currently we have that the last line in 

```julia
using Manifolds
M = Circle()
p = 0.
X = π/2
q = exp(M,p,X)
exp!(M, q, p, X)
```

fails since that function does not exist for Circle (or `Euclidean()` or other number-type manifolds, I have a few listed at
https://github.com/JuliaManifolds/Manopt.jl/blob/9fc85ccf89f3aacbddcb57e3144fbe69fa23f62f/src/Manopt.jl#L193
This makes generic algorithms for them complicated, indeed quite a while ago I started 
https://github.com/JuliaManifolds/Manopt.jl/blob/master/src/plans/nonmutating_manifolds_plans.jl
which is only loaded if you load manifolds together with Manopt.jl

But that is code duplication (besides using `retract` in all places instead of `retract!`).

What do you think about introducing `retract!`? Is that possible or would that break stuff? We can not modify `q` sure, but can we find some other trick?

kellertuer:
Since it is similar (working through all solvers) this could be done together with #223.

kellertuer:
There is one caveat here (and a way to avoid one copy)
the actual high level interface might return more than just the result `q` (when setting `return_state=true`. Besides that with the filled value one can also directly call `gradient_descent!`. So this would look more like

```julia
function gradient_descent(M::AbstractManifold, f, grad_f, p::Number, kwargs...)
    q = fill(p)
    rs = gradient_descent!(M, f, grad_f, q, kwargs...)
    #return just a number if  the return type is the same as the type of q
    (typeof(q) == typeof(rs)) && (return rs[])
    return rs 
end
```

this is still a quite generic way we just have to intrdoduce to every solver (like currently the `gradient_descent` non-mutating ones are very boiler-plate-ly).

kellertuer:
Hm, it is a bit more complicated, since `grad_f` _also_ returns numbers and has to be adopted. I will slowly continue on this but it might actually take quite some time.

mateuszbaran:
Hm, right, that's actually more complicated than I thought initially.

kellertuer:
Well, I got it to work and now know what to avoid.
The current state is far less code than the old reimplementation of Linesearch; so that's nice.

We should just not use `fill` on Floats it seems.

kellertuer:
This will be resolved in #236.

PR #236, opened 2023-04-10
Nicer Interfaces for `::Number` Manifolds and Objectives

### LINKED ISSUE #223: High-lever interfaces starting with `solvername(M, obj, ...)`
Since `ManoptExamples` will also provide objectives – and the fact that we have objectives, it might be nice to provide high-level interfaces accepting objectives so that we have

```
gradient_descent(M, f, grad_f, x0)
```

but also

```
gradient_descent(M, gradient_objective, x0)
```

The first could just construct the objective and pass to the second. Similarly for the in_place variants.
The lower one (objective) would define all defaults to have them in just one place.

kellertuer:
Trying to resolve #223 and #235

## Todo
* [x] check which algorithms _can_ be provided in the way of #235 way (maybe not all necessary) - it can in most cases
* [x] rework all interfaces according to #223, see list below
* Algorithms to check / rework
  * [x] Alternating Gradient Descent (not for numbers, since it alternates on an array anyways)
  * [x] augmented Lagrangian Method
  * [x] Conjugate Gradient
  * [x] Cyclic Proximal Point Algorithm
  * [x] Difference of Convex Algorithm
  * [x] Difference of Convex Proximal Point Algorithm
  * [x] DouglasRachford
  * [x] EPM
  * [x] FW
  * [x] Gradient Descent
  * [x] LevenbergMarquardt (not (yet) for Numbers, maybe still rename `jacB` kw) 
  * [x] NelderMead
  * [x] PS
  * [x] Quasi Newton
  * [x] Stochastic Gradient Descent
  * [x] Subgradient Method
  * [x] TCG
  * [x] trust regions (internally then already uses the right TCG)
* [x]  check that the docs are fine (since I introduced one small new function here as well


kellertuer:
Ah, I at least see the problem.

The `NonMonotoneLineSearch` uses a storage for a point and a vector, but since it (during initialisation) never “sees” an actual point (i.e. how it is represented) it takes a random point, which is a float.
The storage worked fine with Floats by the way before (I was not even aware of that).

Now since we have a float storage, that of course errors when we com by with 1-arrays. This is not only a problem here but even a general problem if other manifolds use other than their default representation. I did fix the number/vector issue when initialising storage – but will have to check for NMLinesearch. Probably the best is to parr an optional init point and init vector to that as well.

kellertuer:
While this turns out _far_ more complex than I thought, I found out some interesting things.
defining `copy` as above (for now only in weak dependency in Manopt, maybe later in Manifolds)

I noticed

```
julia> 1.0 .* inverse_retract(Circle(), fill(0.0), fill(0.0))
0.0

julia> 1.0 * inverse_retract(Circle(), fill(0.0), fill(0.0))
0-dimensional Array{Float64, 0}:
0.0
```

So I am not so sure I want to use `.*` any longer to scale tangent vectors?

kellertuer:
The switch to 1-dim arrays (using `[p]` instead of `fill(p)`; we could also use `fill(p,1)` I think) makes this more reliable (might even revert the changes in Nesterov)

See https://github.com/JuliaLang/julia/issues/28866 thanks to the super-fast answer in https://discourse.julialang.org/t/broadcast-strange-on-0-dimensional-array/97311/2

mateuszbaran:
> * `copy(Circle(), 0.1)` errors since `0.1` is not a `Vector{Float64}` and of float we have

Hm, `copy(Circle(), 0.1)` should just return `0.1`. This should be fixed in ManifoldsBase.jl. I will make a patch.

> So maybe parts of `Manifolds` are also not so much tested with `::Number``s.

Yes, I very rarely use plain numbers in Manifolds.jl. 

> So I am not so sure I want to use `.*` any longer to scale tangent vectors?
> The same for `.-`

Oh, I didn't know about that problem with 0-dim arrays. Let's just use one-element 1-dim arrays with broadcasting then.

kellertuer:
> > * `copy(Circle(), 0.1)` errors since `0.1` is not a `Vector{Float64}` and of float we have
> 
> Hm, `copy(Circle(), 0.1)` should just return `0.1`. This should be fixed in ManifoldsBase.jl. I will make a patch.

great!

> 
> > So maybe parts of `Manifolds` are also not so much tested with `::Number``s.
> 
> Yes, I very rarely use plain numbers in Manifolds.jl.

Maybe this PR will check a few of those then :) but I think the copy might be the main thing.
 
> > So I am not so sure I want to use `.*` any longer to scale tangent vectors?
> > The same for `.-`
> 
> Oh, I didn't know about that problem with 0-dim arrays. Let's just use one-element 1-dim arrays with broadcasting then.

Well, t was interesting to learn, and we just use 1-dims and then it should be fine :)

kellertuer:
This is a bit more work than I thought, but a nice occasion to unify and test all high-level interfaces. ALM took me about an hour, but the kwargs got nicer in quite a few places.

mateuszbaran:
Yes, it indeed looks like quite a bit of work but I think it's a good change :slightly_smiling_face: .

kellertuer:
Yes, I think both things are good ideas, both the preprocessing for numbers (though it takes a bit with adopting functions),
and I like the interfaces `alg(M, obj, p)` if you have an objective that can be run on different manifolds. And all changes are non-breaking indeed.

So this is definetly useful. Just also takes bit to rework.

kellertuer:
...*phew* just 10 Algorithms left. It is nice to also review a few tests and improve them, but going through _all_ algorithms is indeed a bit of work by now.

But. Both new things are really nice to have (though I will not do the number thing for the more-involved algorithms like RCPA and PDSSN).

kellertuer:
For EPM it even made phrasing the defaults much nicer, because some required the objective already. So this rework not only separates some code nicely, it also makes some code really nice.

kellertuer:
Checked Chambolle-Pock and PDSSN today, they make not so much sense in this new form, since they have both a lot of mandatory functions (which sure could be wrapped in their objectives) but also quite a few initial mandatory steps, so that the interface with the objective would still be too crowded. Will think about something nicer for them in some next PR and removed them from the list here.

kellertuer:
@mateuszbaran I have one small questions on `LevenbergMarquardt`

* Do you think a variant on Numbers (Circle/PositiveNumbers/Euclidan) would be useful? I am not sure how to test this
* `expect_zero_residual` was not documented, I already added a line, can you add an explanation, what it does? I mean I know what it indicates, but now how that changes the algorithm.

mateuszbaran:
> * Do you think a variant on Numbers (Circle/PositiveNumbers/Euclidan) would be useful? I am not sure how to test this

That would be a valid optimization problem but I would expect that there should be a better specialized algorithm for that case.

> * `expect_zero_residual` was not documented, I already added a line, can you add an explanation, what it does? I mean I know what it indicates, but now how that changes the algorithm.

I don't really have a good explanation how the algorithm is changed. BTW, `expect_zero_residual` is documented for `LevenbergMarquardtState`.

kellertuer:
> > * Do you think a variant on Numbers (Circle/PositiveNumbers/Euclidan) would be useful? I am not sure how to test this
> 
> That would be a valid optimization problem but I would expect that there should be a better specialized algorithm for that case.

Then I will not cast the special case here

> > * `expect_zero_residual` was not documented, I already added a line, can you add an explanation, what it does? I mean I know what it indicates, but now how that changes the algorithm.
> 
> I don't really have a good explanation how the algorithm is changed. BTW, `expect_zero_residual` is documented for `LevenbergMarquardtState`.

Ah, I took that one over then for now, though it Dows not help much for a user to read that text ;)

Issue #240, opened 2023-05-05
Make the sub solver of trust regions replaceable 


kellertuer:
Similar to Augmented Lagrangian, Exact Penalty Method or the two DC solvers, the trust region should be reworked to have

* subproblem and substate as fields of the trust region state
* a more flexible approach to setting the initial value of the subsolver
* the subprobmen and substate as parameters so one can provide other sub solvers

PR #241, opened 2023-05-06
Add customizable subsolver to trust_regions

mateuszbaran:
This partially solves #240 . Defining custom sub-problems doesn't seem particularly important for extensibility. See also https://discourse.julialang.org/t/ann-manopt-jl/24906/34 .

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.4.18"
+version = "0.4.19"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"
--- a/src/solvers/trust_regions.jl
+++ b/src/solvers/trust_regions.jl
@@ -41,7 +41,7 @@ where all but `p` are keyword arguments in the constructor
     TrustRegionsState(M,
         p=rand(M),
         X=zero_vector(M,p),
-        tcg_state=TruncatedConjugateGradientState(M, p, X),
+        sub_state=TruncatedConjugateGradientState(M, p, X),
 )
 
 construct a trust-regions Option with all other fields from above being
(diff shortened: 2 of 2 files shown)


kellertuer:
Woah. Superfast. I am just back from touristy things in Bergen.

For the problem – hm – all other solvers habe their own sub problems, so maybe it would be reasonable to come up with the quadratic problem to solver here as well and store that in a sub problem? Maybe that could enable caching at some time or it enables other subsolvers to maybe use more advanced sub problems?
(And I hope it is not too hard to do)

mateuszbaran:
> Woah. Superfast. I am just back from touristy things in Bergen.

Cool, I hope you had fun :slightly_smiling_face: 

> For the problem – hm – all other solvers habe their own sub problems, so maybe it would be reasonable to come up with the quadratic problem to solver here as well and store that in a sub problem? Maybe that could enable caching at some time or it enables other subsolvers to maybe use more advanced sub problems?
> (And I hope it is not too hard to do)

I don't really have a good idea what might be reasonably needed by other subsolvers other than a custom state. I would prefer to go for a simple and minimally breaking solution here but I'm open to specific suggestions :slightly_smiling_face: .

kellertuer:
> Cool, I hope you had fun 🙂
Yeah, I hiked up Ulriken – beautiful view, though from sea level up to 643m height and back is quite a tour.

You might be right, we can switch to a sub problem when necessary as well.
My only concern was that this is the not completely unified, but maybe that is fine until we know for sure what we need here.
Besides that this looks like a much easier fix then I expected ;)

kellertuer (approved):

kellertuer:
Oh we have to bump version on either this or the other PR.

mateuszbaran:
> Yeah, I hiked up Ulriken – beautiful view, though from sea level up to 643m height and back is quite a tour.

Nice, that's quite high indeed.

> You might be right, we can switch to a sub problem when necessary as well.
> My only concern was that this is the not completely unified, but maybe that is fine until we know for sure what we need here.
> Besides that this looks like a much easier fix then I expected ;)

Yes, it's not completely unified but it goes in the right direction. So there is less work for later :slightly_smiling_face: .

> Oh we have to bump version on either this or the other PR.

I can bump it here.

PR #244, opened 2023-05-07
Count and Cache Objectives.

### LINKED ISSUE #176: Work counters
Optim.jl has nice work counters that give number of iterations, calls to objective, gradient, time, etc. It would be nice to be able to obtain such information in Manopt.jl easier than now.

### LINKED ISSUE #192: A Cache for calls to fields of the objective
In #174 we introduce(d) a simple cache to cache the last cost (function call) and last gradient (function call). It would be nice to have that more general a cache of length `n` entries instead of just one caching 

  * `:Cost`
  * `:Gradient` (for the stochastic or alternating gradient also the `k`th gradient
  * `:Hessian` (the tangent vector when thinking of the 1-1form)
  * `:Prox` proximal maps (again possibly the `k`th prox
  * `:Subgradient`
  * and possibly any other function evaluation from an objective that might be worth caching

These could be addressed / specified by a symbol that uniquely identifies one of the `get_` functions.

A simple idea to keep track of valid caches would be to assign to every new point `p` an integer and associate cache entries to these numbers.

My idea for the cache would be as follows: Calling any of the accessor functions (for example `get_gradient(M, obj, p)`  or `get_proximal_map(M, obj, p, k)` would do the following steps
1. look up whether `p` is already cached, if so remember its integer `m`
2. look up whether the to-cache-element (here the gradient) is available for `m`.
  a. If that is the case, return that.
  b. If that is not the case and there `n` gradients in the cache, remove the oldest (lowest integer entry). Then evaluate the gradient and store it together with `m`
3. if we removed a gradient in step 2, check whether the corresponding `p` (or number `k`) has an entry in another cache, otherwise remove `p` from the remembered points

This way we could even store information for more than `n` points (for example `n` costs, but also `n` gradients for different points). 

Datastructure wise, the single caches per function evaluation could be sparse arrays, the points might be more complicated, since we have to be able to lookup p->index _and_ index->p, maybe we could store them either as pairs?

Similar to the simple cache, this cache would of course just be an `AbstractManifoldObjective`.

kellertuer:
This is a PR to resolves #176 and resolves #192, since both require a bit more work on decorated objectives.

# roadmap / Todo

* [x] Sketch a `CountObjective`
* [x] check that it acts as a proper decorator, i.e. all other methods acting on objectives pass down propoerly
* [x] Sketch a `CacheObjective`  – mainly find a good way to organise the cache
* [x] implement tests
* [x] document factories
* [x] introduce the general `count=` and `cache=` keywords for all solvers
* [x] figure out a way to return cache and counters in the high level interfaces.
* [x] write a short tutorial
* [x] make `LRUCaches` a weak dependency
* [x] make all other Requires dependencies also into weak dependencies starting in Julia 1.9
* [x] check that all docs still work and the new extension approach works thoroughly.
* [x] implement a summary for counts (also on the tuple `(objective, state)`
* [x] find a good way to initialise the vectorial counters for constraints and proxes


kellertuer:
Before I start twiddling into the details why currently the inplace grad does not yet modify the given memory – @mateuszbaran could you take a look at my LRU Cache design whether that is efficient with the current idea of the Named Tuple of LRUs?

After having fixed the mutating variant, for sure there is a bit of work for other access functions as well, but I can always use a combination of count and cache to test – that is also why I am doing them together :)

mateuszbaran:
> LRU Cache design whether that is efficient with the current idea of the Named Tuple of LRUs?

Are you sure Manopt actually needs LRU caches? IIRC CG and quasi-Newton with just last usage caching had almost no repeated evaluations. I'm not sure how it is in other solvers but I wouldn't be surprised if that was also the case.

Work counters are a good idea for sure.

kellertuer:
Hm, so you say caches are not useful? Why did we then have the issue #192 with so much discussions about that?

mateuszbaran:
Well, maybe they are useful for something, I just personally haven't found a good use for that. I thought you might have encountered a case where such longer caching would be useful and now I'm curious for which solver it is important? Sorry, I should have engaged more in that issue then.

mateuszbaran:
> Well one of my ideas is to make the Cache a bit longer and reuse it for example when comparing different solvers, when the cost/grad/HEss is costly to evaluate you really do not want to evaluate it more than necessary. Maybe also for sub solvers? Something like that.

Wouldn't different solvers almost always evaluate objective/grad/Hess at different points?

> But now I am not sure its worth the effort (since similar to the count-thingy, one has to write quite a few wrappers).

Yes, that's quite a lot of effort (especially when considering optimizations) so it would be good to have specific benchmarks to optimize (optimization problem and solver that causes repeated evaluations).

kellertuer:
For me the main part is the question whether it's worth the effort. Then we could start with a few (cost/grad/hess) and extend that access part when necessary.

For Count I would cover much more since that seems more straight forward (maybe even a bit stupid work to do for all accessors).

mateuszbaran:
> For me the main part is the question whether it's worth the effort. Then we could start with a few (cost/grad/hess) and extend that access part when necessary.

Currently I don't have a problem where longer LRU caches would help so it might not be worth the effort. But it's possible I will encounter a use for them later, who knows? In any case it doesn't look urgent.

> For Count I would cover much more since that seems more straight forward (maybe even a bit stupid work to do for all accessors).

Yes, that should be straightforward.

kellertuer:
Then maybe the following is a good tradeoff:
I will only do a cache for cost and gradient (for both the gradient_objective as well as the combined costgrad objective) more as a proof of concept, that this can easily be used – and we can extend it when the interest is there.

And for Count I will check what we all can count :)

mateuszbaran:
Sure, that counts as a good compromise :wink:

kellertuer:
I checked a bit what Manopt/Matlab does and they use the storage for cost/grad/hess (they do not have prox) as well as interims results to share between cost and grad (we have another approach for that).

So I will stick with the cost/grad approach and maybe add Hess/prox later as well.

mateuszbaran:
> * Sketch a `CountState` – but which counts to include besides the number of iterations?

I can't think of anything to count other than the number of iterations that isn't solver-specific.

kellertuer:
Though about that as well, and it might even be easier for these counts to just be handled as (simplified) recoded values then, something like `:Step` to count the total number of steps (not per iteration but on the overall thing.

kellertuer:
🥳 the gradient cache works (now also for `get_gradient!`.
So now the rest is just working through a few more counts, maybe write a short tutorial and the combined CostGrad to work as well.

kellertuer:
I noticed one final challenging part when introducing cache and counters to the high level interfaces. 
Since they decorate the objective, they are not returned, when setting return_state to true.
While I would like to get rid of that thing anzways (though that is breaking),
one could for now introduce `return_objective_and_state that returns a tuple of the objective and the state (and this could even get its own summary).

In the long run (Manopt 0.5) Maybe the following might be an idea (will do an issue some day)
' usually the x' is returned
' if any decorator of the state is active, the state is returned
' if any objective is active the tuple of state and objective is returned

one could skip the second and for any decoration always return said tuple, but for the tuple accessing the solution is a but trickier (unless one has a special accessor for that)
I feel this is a bit more natural then actively setting the return value one wants to get.

kellertuer:
While today implementing the Cache for CostGrad (which should work fine already) I noticed that the inerplay between decorators might be complicated for now, since we do not have a `AbstractDecoratedObjective{E,O}` where `O` is the most inner one that gets decorated. I might introduce that, since then the dispatch can still be done according to the undecorated objective, such that when calling `get_cost´ on a `Cache(Count(CostGrad))` still the right cache costgrad is called – though I might have to check how to best ensure that counting still works correctly then.

Complicated – but if it works I think it will add a really nice interoperability.

src/plans/count.jl
```diff
@@ -0,0 +1,200 @@
...
+dispatch_objective_decorator(::CountObjective) = Val(true)
+
+function _count_if_exists(co::CountObjective, s::Symbol)
+    return haskey(co.counts, s) && (co.counts[s] += 1)
+end
+function _count_if_exists(co::CountObjective, s::Symbol, i)
+    return haskey(co.counts, s) &&
+           length(i) == ndims(co.counts[s]) &&
+           all(i .<= size(co.counts[s])) &&
+           (co.counts[s][i] += 1)
```
mateuszbaran, line 60: `co.counts[s]` is an integer, how is it supposed to work?

Project.toml
```diff
@@ -9,6 +9,7 @@ ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+LRUCache = "8ac3fa9e-de4c-5943-b1dc-09c6b5f20637"
```
mateuszbaran, line 12: This might also be a weak dependency?

mateuszbaran:
Yes, `AbstractDecoratedObjective{E,O}` might be a good idea.

src/plans/count.jl
kellertuer: This is still a bit work in progress, in this case it is not just an integer, that is the case above.

The application in mind is the following: CPPA has a whole vector of proxies – the same for the inequality functions, equality functions or elements for stochastic gradient.
In these cases, we want to (hopefully!) count the individual calls to the proxes. That is what this aims for. But I am not yet done sketching this and have written the tests – initialising these “group counts” might be tricky.

kellertuer:
> Yes, `AbstractDecoratedObjective{E,O}` might be a good idea.

Works! Now thinking about even doing the same for the state some time, since its really neat to now also see the type of the (undecorated) state in the parameters, though one has to be a bit more careful in the constructors.

Project.toml
kellertuer: That is now really a case where I think it is a tradeoff between ease of use (“Caches just work”) and that you have to load this, I can think about it. If so, I will probably even rework all dependencies into the way Chris did in DiffEq, to be weak for 1.9 and Require earlier.

src/plans/count.jl
mateuszbaran: I see, then it could be a bit tricky to implement.

Project.toml
mateuszbaran: The only user-facing difference would be whether they have to type `using LRUCache` or not before using caches, and for people who don't need caches the package would be lighter.

BTW, `Manifolds`, `Plots` and `LineSearches` could also be turned into weak dependencies.

kellertuer: Yes I would do that on the same run. And we should check Manifold for that as well.

Ok, maybe weak is a good idea, will do it along the way when most of the modelling is done.

mateuszbaran: I'm going to try doing some weak dependencies for Manifolds.jl soonish.

kellertuer: This is the guide I will follow https://discourse.julialang.org/t/package-extensions-for-julia-1-9/93397/5 (but I think you did that as well somewhere already?

mateuszbaran: Yes, I've seen that guide, it is good. I haven't used it fully yet though because I've just switched to 1.9 a couple of days ago but I've followed the general file structure in the LineSearches PR here.

kellertuer:
I have a question concerning moving LRU to weak dependencies – there is the nice way the cache should be activatable by the `cache=` keyword, but I struggle where to put (a) the struct for the LRUCaches-Cache (since the extension should introduce new ones?) and how to change the processing of the `cache=` thing, since it has _one_ line in there that has to be added in the extension? I am a but lost here.

kellertuer:
Ah, and further confusions:

* the Linesearch Ext _does_ define new structs. – I read this should not be the case?
* the Linesearches extension is in `src/ext` I thought it should have been just in `ext`? 

I fear I am not yet understanding what extensions are “allowed” to do and what not and where they should be.

For the `factory` above I also do not know how to do this in a weak way.

kellertuer:
The newest step is a try to move towards weak deps – but it seems no single extension does actually work. So I am really missing some understanding it seems.

kellertuer:
> Regarding `objective_cache_factory(M::AbstractManifold, o::AbstractManifoldObjective, cache::Symbol)`, I think you should dispatch on the cache type instead of doing things like `(cache[1] === :Simple)`. That way you can have a method in Manopt.jl that handles `:Simple` cache and a method in extension that handles LRU cache.

Can you elaborate on this? The idea is that a vector of symbols comes along and is translated into types, so I am not sure how to dispatch symbols on the cache types?

Example `cache=:Simple` should wrap the objective of a solver in the simple cache, while ? cache=:LRUCache` should warp it in an LRU Cache.

mateuszbaran:
> For now I resolved the first by actually moving the type – but I have no clue how to export that – similarly for the alternating gradient descent (which can only be defined on ProductManifolds so it has to be in the ManoptManifoldsExt – but then it is never exported?)

You need to have `alternating_gradient_descent` and `alternating_gradient_descent!` _functions_ in Manopt proper, and methods that dispatch on `ProductManifold` in the extension. I can help you organize the code.

kellertuer:
Yes help would be appreciated – I fear I am not clever enough for this. I tried my best, but I am quite lost on
*  the alternating solver
* the new LRU Cache should ever work when it is not exported and is defined where it should not be defined
* how the recompilation errors we still have can maybe also be seen when calling `precompile`, since now they just error and I see no errors

I fear I really miss some of the main concepts here.

mateuszbaran:
> Can you elaborate on this? The idea is that a vector of symbols comes along and is translated into types, so I am not sure how to dispatch symbols on the cache types?
> 
> Example `cache=:Simple` should wrap the objective of a solver in the simple cache, while ? cache=:LRUCache` should warp it in an LRU Cache.

For example something you can turn
```julia
function objective_cache_factory(M, o, cache::Tuple{Symbol,<:AbstractArray}) 
     (cache[1] === :Simple) && return SimpleCacheObjective(M, o) 
     (cache[1] === :LRU) && return LRUCacheObjective(M, o, cache[2]) 
     return o 
end
```
into 
```julia
function objective_cache_factory(M, o, cache::Tuple{Symbol,<:AbstractArray}) 
     (cache[1] === :Simple) && return SimpleCacheObjective(M, o) 
     (cache[1] === :LRU) && objective_cache_factory(M, o, Val(cache[1]), cache[2]))
     return o 
end
```
and in the extension
```julia
function objective_cache_factory(M, o, ::Val{:LRU}, cache2::AbstractArray)
    return LRUCacheObjective(M, o, cache2) 
end
```

mateuszbaran:
> Yes help would be appreciated – I fear I am not clever enough for this. I tried my best, but I am quite lost on
> 
> * the alternating solver
> * the new LRU Cache should ever work when it is not exported and is defined where it should not be defined
> * how the recompilation errors we still have can maybe also be seen when calling `precompile`, since now they just error and I see no errors
> 
> I fear I really miss some of the main concepts here.

OK, I more or less know what to do -- give me an hour and I should be able to come up with something.

kellertuer:
Running formatter and the same errors appear here, again. I give up for today. Seems I forgot how to do stuff.

kellertuer:
> ```julia
> function objective_cache_factory(M, o, cache::Tuple{Symbol,<:AbstractArray}) 
>      (cache[1] === :Simple) && return SimpleCacheObjective(M, o) 
>      (cache[1] === :LRU) && objective_cache_factory(M, o, Val(cache[1]), cache[2]))
>      return o 
> end
> ```
> 
> and in the extension
> 
> ```julia
> function objective_cache_factory(M, o, ::Val{:LRU}, cache2::AbstractArray)
>     return LRUCacheObjective(M, o, cache2) 
> end
> ```

Hm, ok, this looks a bit complicated, but I see the reasoning behind that. Hmmmmm. Seems for me nearly that Extensions are more like making everything in coding much more complicated. But maybe that's the area where speed comes at the cost of ugly code.

kellertuer:
> OK, I more or less know what to do -- give me an hour and I should be able to come up with something.

Help would be appreciated, because besides Formatter annoying me on this PR, I am not able to fix the errors I have, since `Pkg.precompile()` says the extensions can not be precompiled due to errors without reporting errors, which seems a bit strange, since without errors messages – fixing them is a bit tough.
But maybe I give up on extensions for now after 3 hours of destroying more than getting forward, leave this to you for now and maybe try understanding Extensions – ... later.

mateuszbaran:
I've sort a part of the errors. The remaining ones should be relatively easily found by looking at test failures. It looks like it's mostly missing imports and wrongly moved (or not moved) things between main Manopt and extensions.

Project.toml
```diff
@@ -17,15 +17,31 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```
mateuszbaran, line 20: I'm not totally sure but it looks like uses of StaticArrays are split between "we also need Manifolds.jl" and "that's easily avoidable".

src/functions/proximal_maps.jl
```diff
@@ -316,7 +316,7 @@ function prox_TV2(
             "Proximal Map of TV2(M, λ, x, p) not implemented for p=$(p) (requires p=1) on general manifolds.",
         ),
     )
-    PowX = SVector(x)
+    PowX = [x...]
     PowM = PowerManifold(M, NestedPowerRepresentation(), 3)
```
mateuszbaran, line 320: Most likely `ProductManifold` would be a better fit but I didn't try yet.

Project.toml
kellertuer: if we can avoid it I would not mind, also if we only need it with manifolds would be great; I am not sure why we needed it.

kellertuer:
> Yes, the errors are quite hard to interpret and not very friendly. You need to know how to trigger the helpful errors 😄 .

If the recommended method (from adding/updating packages) yields zero information about the errors – it is indeed _quite_ hard to interpret ;)

src/functions/proximal_maps.jl
kellertuer: indeed for the fixed case of 3 product is maybe better.

mateuszbaran:
I've fixed almost everything now. There are just 4 failing tests, I will leave them for you :slightly_smiling_face: . You probably need to make some little rearrangements after my fixes, I've been adding and moving stuff without too much thought just to get the tests passing, so things may be in the wrong order. I also haven't checked docs.

kellertuer:
Thanks for the help, docs - don't worry about them; I am a fan of fixing docs anyways :)

ext/ManoptLRUCacheExt.jl
```diff
@@ -0,0 +1,174 @@
...
+    p::P=rand(M),
+    v::R=get_cost(M, objective, p),
+    X::T=zero_vector(M, p),
+    cache_size=10,
+    cache_sizes=Dict{Symbol,Int}(),
+) where {E,O<:AbstractManifoldObjective{E},R,P,T}
+    c = init_caches(caches; p=p, v=v, X=X, cache_size=cache_size, cache_sizes=cache_sizes)
+    return LRUCacheObjective{E,O,O,typeof(c)}(objective, c)
+end
+function LRUCacheObjective(
```
kellertuer, line 56: Would it maybe be reasonable to define this constructor `(m, args...; kwargs...) in `Manopt.jl` directly to be able to export it and inform the user that he needs to load LRU to use this?
That way if you get code that uses LRU but you have not loaded it upfront (because you only copied parts and not the using upfront)?

Something similar could be done for the solvers – and cool that you even spotted Chambolle-Pock.

kellertuer: Cool. Will do that because it also makes the constructor “more visible” - so if you use autocomplete or something you can discover it and then you learn what you have to load to use it :)

kellertuer: even better, with that constructor still being defined, the `objective_cache_factory` can also stay as before I think.

kellertuer: Hm, but if it is already a function that throws an error (on a different unspecified signature) in Manopt I can not define a new function in the extension due to

```
ERROR: LoadError: cannot assign a value to imported variable Manopt.LRUCacheObjective from module ManoptLRUCacheExt
Stacktrace:
```

So an extension should not define new stuff but can also not define new signatures for old stuff? What am I missing here?

kellertuer: So if I define a function with an error with the name of the constructor – I can not define the struct afterwards – if I do _not_ define the function – leading the extension does not make `LRUCacheObjective` available neither explorted not within `Manopt.jl`: Maybe this is all not possible? At least I do not see how structs and extensions presupposed to work together.

mateuszbaran: > So an extension should not define new stuff but can also not define new signatures for old stuff? What am I missing here?

You can add methods to existing functions (and constructors to existing types) in package extensions. I'd have to take a look at what exactly you did in `ManoptLRUCacheExt` to be able to tell what could be wrong.

> So if I define a function with an error with the name of the constructor – I can not define the struct afterwards – if I do _not_ define the function – leading the extension does not make `LRUCacheObjective` available neither explorted not within `Manopt.jl`

If a method in Manopt uses a type, that type cannot be defined in an extension. If a type is defined in an extension, it (effectively) can only be instantiated by methods from said extension. But they can be methods of functions defined elsewhere.

kellertuer: Hm, since the exams I am supposed to grade today. are still not scanned, I will twiddle around with this a bit, but I am not yet convinced that this will be any useful for any of the extensions we for now introduced, since it seems one should actually never define structs (unless for use only within the extension) within an extension.

For now we merely do the exact opposite.

mateuszbaran: Yes, one big limitation is that structs defined in extensions can only be instantiated by methods from that extensions. As demonstrated, it can be worked around by either defining types in the main module or by making "factory"-like functions that instantiate types defined in extensions. So I don't see it as a big limitation, and maybe even a small win: it results in more easily extensible code. Interface enforced by that rule can be more easily adapted to a new library.

mateuszbaran: For example, if someone made a "MyAwesomeCache.jl" library, with this structure it could be very easily used by code that was designed to use LRUCache.jl. If you tie a struct more closely to a cache library and let people instantiate it directly, more adjustments would be needed.

kellertuer: I already switched to a name `CachedManifoldObjective` that might be able to work with arbitrary caches and only the LRU initialisation is done in the extension now.

kellertuer: The newest commit (twiddely....) is an approach that might work. I will now check the other extensions before checking the tests again. I think there should sometimes be a reasonable error-default.

mateuszbaran: > I think there should sometimes be a reasonable error-default.

I think error hints should work fine as a replacement: https://github.com/JuliaManifolds/Manifolds.jl/pull/607/files#diff-22f0e186bf4bcece078fec1aa646b8e18e436bfcf26138178f0a5c43c5dcf571R494 ?

kellertuer: I lost about an hour trying to figure out these hints – I am not able to do that. But tests are back working and I hope my current (just the most recent) commit does something reasonable? Currently the new Cache init-function dispatches on the cache type and the default type (`Nothing`) returns an error. Loading LRU defines a more specialised function that even sets LRU as the default type.

kellertuer:
Hm, still does not work, tests on 1.9 do work but the old ones on 1.8 no longer. I fear this is turning into a much more infinite story than I would like.

kellertuer:
It seems one can only `..Manifolds` load that one that is in the extension but not the rest on Julia 1.8? I am ending this in pure guesswork, since the examples we have (or I could find) do the `..LoadPackage` for exactly one package, and it seems trying that with more just makes everything on 1.8 fail?

kellertuer:
So _if_ we get it to work (maybe now?) We end up with 

Julia 1.8 with Requires
```
julia> @time using Manopt
    0.888875 seconds (2.03 M allocations: 140.309 MiB, 0.99% gc time, 5.34% compilation time: 36% of which was recompilation)
```

Julia 1.9 with extensions
```
julia> @time using Manopt
  0.297500 seconds (617.65 k allocations: 38.261 MiB, 4.22% gc time, 2.97% compilation time)
```

kellertuer:
The final trick was by the way, that within the extension files only the one being loaded should be `using ..Manifolds` for example, the rest can for both cases be loaded with just `using` (if it is a hard dependency of Manopt anyways).

I think I removed StaticArrays, but might check that in 1.9 for now as well (in 1.8 it is not in the `@requires` thingy if I see correctly).
Overall – it took me a while to understand the concept of extensions, what they can and can's to; I think that improved to day by two-epsilon; the technicalities on backward-compatibility are still a bit tricky I think.

mateuszbaran:
> The final trick was by the way, that within the extension files only the one being loaded should be `using ..Manifolds` for example, the rest can for both cases be loaded with just `using` (if it is a hard dependency of Manopt anyways).

Weird, in Manifolds.jl I didn't have that problem but let's not waste too much time on that.

Good to see that you've managed to figure it out a bit :slightly_smiling_face: .

kellertuer:
@Mateusz can you maybe take a look why since we moved to extensions your LineSearch Example in the docs no longer works? It seems though we have `using Linesearches` in said example block the extension is not loaded?

mateuszbaran:
> @mateusz can you maybe take a look why since we moved to extensions your LineSearch Example in the docs no longer works? It seems though we have `using Linesearches` in said example block the extension is not loaded?

Weird, it works for me, both in Julia 1.8 (standalone) and 1.9 (standalone & full docs build). Where did you see it not working?

> Also in the current form it seems quite “hacked” – the struct defined in Manopt only works with LineSeaches.jl and is only documented that way, but not in the extension (since that is not possible) – so it would be better to have it a bit more generic? Maybe the docs and even a more generic function (which is currently in the extension)? That helped for Caches at least and I think here the tricks necessary are maybe the same.

Yes, good idea -- docs could be a bit more generic. What do you mean regarding "a more generic function"?

kellertuer:
If I run the docs locally I get (even when not properly importing the Lineserach thingy instead of just using it)

```
┌ Warning: failed to run `@example` block in src/extensions.md:8-46
│ ```@example
│ using Manopt, Manifolds, LineSearches
│ 
│ # define objective function and its gradient
│ p = [1.0, 100.0]
│ function rosenbrock(::AbstractManifold, x)
│     val = zero(eltype(x))
│     for i in 1:(length(x) - 1)
│         val += (p[1] - x[i])^2 + p[2] * (x[i + 1] - x[i]^2)^2
│     end
│     return val
│ end
│ function rosenbrock_grad!(M::AbstractManifold, storage, x)
│     storage .= 0.0
│     for i in 1:(length(x) - 1)
│         storage[i] += -2.0 * (p[1] - x[i]) - 4.0 * p[2] * (x[i + 1] - x[i]^2) * x[i]
│         storage[i + 1] += 2.0 * p[2] * (x[i + 1] - x[i]^2)
│     end
│     project!(M, storage, x, storage)
│     return storage
│ end
│ # define constraint
│ n_dims = 5
│ M = Manifolds.Sphere(n_dims)
│ # set initial point
│ x0 = vcat(zeros(n_dims - 1), 1.0)
│ # use LineSearches.jl HagerZhang method with Manopt.jl quasiNewton solver
│ ls_hz = Manopt.LineSearchesStepsize(M, LineSearches.HagerZhang())
│ x_opt = quasi_Newton(
│     M,
│     rosenbrock,
│     rosenbrock_grad!,
│     x0;
│     stepsize=ls_hz,
│     evaluation=InplaceEvaluation(),
│     stopping_criterion=StopAfterIteration(1000) | StopWhenGradientNormLess(1e-6),
│     return_state=true,
│ )
│ ```
│   value = MethodError: objects of type Manopt.LineSearchesStepsize{LineSearches.HagerZhang{Float64, Base.RefValue{Bool}}, ExponentialRetraction, ParallelTransport} are not callable
└ @ Documenter.Expanders ~/.julia/packages/Documenter/H5y27/src/Utilities/Utilities.jl:34
```

kellertuer:
More generic in the sense that I am not yet sure the current Linesearch struct we have in Manopt is generic or _only_ for Linesearches? 
To me the code reads a bit like the Simpsons episode where Homer buys a Bowling ball for Marge's birthday, but already engraves his name (since he thinks she will anyways not use it). This is what the struct currently feels like. “Generic” but `Linesearches.jl` already engraved, since it can not be used with anything else.

For Caches I at least got to, that you have to write your own init for caches and these (maybe other caches then) are expected to have the `get!` function implemented.
Whether the struct can the _that_ general I do not know, since Linesearches itself is a bit – not generic in itself.

mateuszbaran:
> If I run the docs locally I get (even when not properly importing the Lineserach thingy instead of just using it)

It works for me though :confused: . A minor note, sometimes I  `up` the environment just to refresh package extensions (i.e. I get errors when I don't do that) -- something you could try maybe?



> More generic in the sense that I am not yet sure the current Linesearch struct we have in Manopt is generic or _only_ for Linesearches?

It is now generic, and `LineSearches` is an example implementation (we could have more).

mateuszbaran:
> To me the code reads a bit like the Simpsons episode where Homer buys a Bowling ball for Marge's birthday, but already engraves his name (since he thinks she will anyways not use it). This is what the struct currently feels like. “Generic” but `Linesearches.jl` already engraved, since it can not be used with anything else.

Huh, I think we just have to erase the engraving :wink:

mateuszbaran:
> > If I run the docs locally I get (even when not properly importing the Lineserach thingy instead of just using it)
> 
> It works for me though confused . A minor note, sometimes I `up` the environment just to refresh package extensions (i.e. I get errors when I don't do that) -- something you could try maybe?

Also note that docs render correctly on CI so it looks like a problem on your computer.

kellertuer:
Hm but _why_ does it then not work locally for. me? I really start to _hate_ extensions. They seem unpredictable. And introduce a bash/Python level to Julia: WOMM – works on my machine.

kellertuer:
Also executing said example on my REPL does not work. Argh. I finally had thought we got extensions to work – this is really annoying.

mateuszbaran:
Extensions were a bit tricky for me too but I thought I have the workflow figured out. I have one more idea: could you delete your Julia's precompilation cache and try again? See here: https://discourse.julialang.org/t/how-to-delete-precompile-cache/13622 . Maybe something went wrong there.

mateuszbaran:
> I just sold 30% of my soul, went for a walk, walked 3x clockwise around a seagull – tadaaa, the extension magic works locally again.

Cool, I will write that down in case my extension magic stops working :smile: .

Nice to see that it's getting close to completion.

kellertuer:
*Phew* quite some new tests to do, but now luckily only

* Count dispatch
* generic objective decorator boilerplate code for primal dual and constraint

are left to do. And maybe introducing a few more count-factory Symbols along the way.

kellertuer:
So, besides a few smaller count-tests, the only feature left is the vectorial-count-thingies – how to best initialise them (user-friendly).
And the tutorial, but that is best to write as the very last use example.

kellertuer:
I went through all Objectives and collected all access functions by now – some tests and a short tutorial are missing, but the rest (so the main part of this code) can already be considered final and reviewed.

Also the remaining tests will be in exactly the same manner as the current tests – checking that counting works.

kellertuer:
We are nearly there, just that when writing the tutorial, I noticed one small bug (that might require a bit more code cos now),

and LRU caches do not yet fully work as intended, since activating the cache on a real world example throws an error where the LRU cache tries to pop on element that is not in the cache, and I have no clue yet how and why that happens.

kellertuer:
The only point left (besides a thorough review of the code) is, whether we should add maybe Hessian and Subgradient or such to the cache features here already as well.

mateuszbaran:
> The only point left (besides a thorough review of the code) is, whether we should add maybe Hessian and Subgradient or such to the cache features here already as well.

I think a simple cache for Hessian and other user-specified functions would be useful.

I will make a small review of this PR today.

kellertuer:
> I think a simple cache for Hessian and other user-specified functions would be useful.

The pint is that every cache takes some extend of work, I think I can do Hessian and Subgradient relatively fast, but for example in the stochastic gradient with a vector of caches – that would be a few days I fear.

I am not sure what you mean with “user specified”, the main problem is to initialise the cache correctly (the overwriting of the access function is not so much work).

mateuszbaran:
We can add caches later as needed.

> I am not sure what you mean with “user specified”

Basically all functions that are passed to optimization functions: cost, gradient, Hessian, subgradient, constraints, proximal maps, Jacobians of constraints etc. Likely not all of them are worth caching though.

kellertuer:
For hessian and subgrad that is doable, for constraints I have to think how to nest do that (see the point-and-index things in counting).

mateuszbaran:
I think caching constraints can be left for later.

kellertuer:
I finished implementations for `:Subgradient`, `:SubtrahendGradient` and `:Hessian`, the only other tangent vector cache we could do (mightiest add that) is `:Preconditioner`; all of these are still missing tests, since that is also quite some part to write

I also sketched `:ProximalMap`, if we get that to work, also the constraint gradients are doable the same way.
The first and easy difference for ProximalMap is that the cache is for points, but I have the point type anyways available when initialising the cache, since that is the key for the cache.
The second, more tricky thing is that there are `n` proximal maps (similar to constraints and their grads) – but one can just add an index to the key and have the tuple `(p,i)` as a key – an idea that just popped into my head somewhen today.

So also proximal maps needs testing but if they work, then all others from counting can also be cached, each objective access function needs about 22 lines of “boilerplatecode” to be cached, and that is even a bit repetitive, but not too much, I hope.

kellertuer:
Caching now also works for
* subgradient
* subtrahend-gradient (also just a tangent vector in one of the elements of DC
* proximal point – which was tough to get right, but now with `(p, λ, i)` as a key works

the last point illustrates, that also constraints and grad constraints are cache-able and merely a bit of work. I would prefer a bit of a review first, since there is one scenario I am not yet sure to have 100% right

If an allocating function (say `get_gradient`) has a cache hit see

https://github.com/JuliaManifolds/Manopt.jl/blob/d1a5f4ed38a8dfa6c0617d0a4e292aa70741e2cd/src/plans/cache.jl#L298-L303

it currently returns what it gets from the cache. I fear that might have side effects if an algorithm continues to modify said element, so I am wondering (and would like some in put on) whether these functions should return a copy of what they get from the cache.

So after some review I might continue doing the rest just as well, since I know how to do it (but it still takes approx 45-60 minutes per cache to do the implementation, documentation (a bit), and especially testing).

kellertuer:
...the question on the function of interest in the last post I can now answer myself: To avoid side effects, where we return a reference that might also point to the cache – these allocating variants have to copy as well. I also added a test line for that.

mateuszbaran:
> ...the question on the function of interest in the last post I can now answer myself: To avoid side effects, where we return a reference that might also point to the cache – these allocating variants have to copy as well. I also added a test line for that.

Yes, that is correct -- allocating variants need to always make a copy.

kellertuer:
Thanks for the feedback.

I think I now also covered the last of the 18 possible caches (that is the same amount we can count stuff), at least the implementation.

I will add tests after a break (or the next days). And wait for coverage to check which ones are actually missing (its easy to loose track when one implements 18 x 2 (inlace/alloc) access functions).
But thanks for encouraging to do that now directly and not just stay with Cost and Grad. I think this shows that the scheme is flexible enough.

docs/src/tutorials/CountAndCache.md
```diff
@@ -0,0 +1,164 @@
...
+[LRUCaches.jl](https://github.com/JuliaCollections/LRUCache.jl) which are (since Julia 1.9) a weak dependency and provide
+the *least recently used* strategy for our caches.
+
+``` julia
+using Manopt, Manifolds, Random, LRUCache
+```
+
+## Counting
+
+We first define our task, the Riemannian Center of Mass from the [Get Started: Optimize!](https://manoptjl.org/stable/tutorials/Optimize!.html). tutorial
```
mateuszbaran, line 53:
```suggestion
We first define our task, the Riemannian Center of Mass from the [Get Started: Optimize!](https://manoptjl.org/stable/tutorials/Optimize!.html) tutorial.
```

docs/src/tutorials/CountAndCache.md
```diff
@@ -0,0 +1,164 @@
...
+    c.count += 1
+    # [ .. Actual implementation of the cost here ]
+end
+```
+
+though this still leaves a bit of work to the user, especially for trakcing more than just the cost.
+
+Similarly caching of function calls can for example be added using [Memoize.jl](https://github.com/JuliaCollections/Memoize.jl) by the user.
+The approach in the solvers of [Manopt.jl](https://manoptjl.org) aims to simplify adding
+both these capabilities on the level of calling a solver.
```
mateuszbaran, line 28:
```suggestion
When the a function like objective or gradient is expensive to compute, it may make sense to cache its results.
Manopt.jl tries to minimize the number of repeated calls but sometimes they are necessary and harmless when the function is cheap to compute.
Caching of expensive function calls can for example be added using [Memoize.jl](https://github.com/JuliaCollections/Memoize.jl) by the user.
The approach in the solvers of [Manopt.jl](https://manoptjl.org) aims to simplify adding
both these capabilities on the level of calling a solver.
```
I think a bit more details as to why a user would want caching is a good idea.

docs/src/tutorials/CountAndCache.md
```diff
@@ -0,0 +1,164 @@
...
+      * :Cost     :  278
+    on a ManifoldGradientObjective{AllocatingEvaluation}
+
+And we see that statistics are shown in the end. To now also cache these calls,
+we can use the `cache=` keyword argument.
+Since now both the cache and the count “extend” the functionality of the objective,
+the order is important: On the high-level interface, the `count` is treated first, which
+means that only actual function calls and not cache look-ups are counted.
+With the proper initialisation, you can use any caches here that support the
+`get!(function, cache, key)! update. All parts of the objective that can currently be cached are listed at [`ManifoldCachedObjective`](@ref). The solver call has a keyword`cache`that takes a tuple`(c, vs, n)`of three arguments, where`c`is a symbol for the type of cache,`vs`is a vector of symbols, which calls to cache and`n`is the size of the cache. If the last element is not provided, a suitable default (currently`n=10\`) is used.
```
mateuszbaran, line 108:
```suggestion
`get!(function, cache, key)!` update. All parts of the objective that can currently be cached are listed at [`ManifoldCachedObjective`](@ref). The solver call has a keyword `cache` that takes a tuple`(c, vs, n)` of three arguments, where `c` is a symbol for the type of cache, `vs` is a vector of symbols, which calls to cache and `n` is the size of the cache. If the last element is not provided, a suitable default (currently`n=10`) is used.
```

ext/ManoptLRUCacheExt.jl
```diff
@@ -0,0 +1,93 @@
...
+"""
+function Manopt.init_caches(
+    M::AbstractManifold,
+    caches::AbstractVector{<:Symbol},
+    ::Type{LRU};
+    p::P=rand(M),
+    v::R=0.0,
+    X::T=zero_vector(M, p),
+    cache_size=10,
+    cache_sizes=Dict{Symbol,Int}(),
```
mateuszbaran, line 44:
```suggestion
    cache_size::Int=10,
    cache_sizes::Dict{Symbol,Int}=Dict{Symbol,Int}(),
```

ext/ManoptLRUCacheExt.jl
```diff
@@ -0,0 +1,93 @@
...
+"""
+    init_caches(caches, T::Type{LRU}; kwargs...)
+
+Given a vector of symbols `caches`, this function sets up the
+`NamedTuple` of caches, where `T` is the type of cache to use.
+
+# Keyword arguments
+
+* `p`   - (`rand(M)`) a point on a manifold, to both infere its type for keys and initialize caches
+* `v`   - (`0.0`) a value both typing and initialising number-caches, eg. for caching a cost.
```
mateuszbaran, line 31: I think I'd prefer a bit more verbose name, maybe `value`?

ext/ManoptLRUCacheExt.jl
```diff
@@ -0,0 +1,93 @@
...
+function Manopt.init_caches(
+    M::AbstractManifold,
+    caches::AbstractVector{<:Symbol},
+    ::Type{LRU};
+    p::P=rand(M),
+    v::R=0.0,
+    X::T=zero_vector(M, p),
+    cache_size=10,
+    cache_sizes=Dict{Symbol,Int}(),
+) where {P,R,T}
```
mateuszbaran, line 45:
```suggestion
) where {P,R<:Real,T}
```

src/plans/cache.jl
```diff
@@ -181,6 +191,551 @@ function get_gradient!(
...
+    cache::C
+end
+function ManifoldCachedObjective(
+    M::AbstractManifold,
+    objective::O,
+    caches::AbstractVector{<:Symbol}=[:Cost];
+    p::P=rand(M),
+    v::R=get_cost(M, objective, p),
+    X::T=zero_vector(M, p),
+    cache_size=10,
```
mateuszbaran, line 256:
```suggestion
    cache_size::Int=10,
```

src/plans/cache.jl
```diff
@@ -181,6 +191,551 @@ function get_gradient!(
...
+    cache_sizes=Dict{Symbol,Int}(),
```
mateuszbaran, line 257:
```suggestion
    cache_sizes::Dict{Symbol,Int}=Dict{Symbol,Int}(),
```

src/plans/cache.jl
```diff
@@ -181,6 +191,551 @@ function get_gradient!(
...
+) where {E,O<:AbstractManifoldObjective{E},R,P,T}
```
mateuszbaran, line 258:
```suggestion
) where {E,O<:AbstractManifoldObjective{E},R<:Real,P,T}
```

src/plans/cache.jl
```diff
@@ -181,6 +191,551 @@ function get_gradient!(
...
+function ManifoldCachedObjective(
+    M::AbstractManifold,
+    objective::O,
+    caches::AbstractVector{<:Symbol}=[:Cost];
+    p::P=rand(M),
+    v::R=get_cost(M, objective, p),
+    X::T=zero_vector(M, p),
+    cache_size=10,
+    cache_sizes=Dict{Symbol,Int}(),
+) where {E,O2,O<:AbstractDecoratedManifoldObjective{E,O2},R,P,T}
```
mateuszbaran, line 273:
```suggestion
    v::R=get_cost(M, objective, p),
    X::T=zero_vector(M, p),
    cache_size::Int=10,
    cache_sizes::Dict{Symbol,Int}=Dict{Symbol,Int}(),
) where {E,O2,O<:AbstractDecoratedManifoldObjective{E,O2},R<:Real,P,T}
```

src/plans/cache.jl
```diff
@@ -192,24 +747,47 @@ on the `AbstractManifold M` based on the symbol `cache`.
...
 @doc raw"""
-    objective_cache_factory(M::AbstractManifold, o::AbstractManifoldObjective, cache::Tuple{Symbol, Array}Symbol)
+    objective_cache_factory(M::AbstractManifold, o::AbstractManifoldObjective, cache::Tuple{Symbol, Array, Array})
+    objective_cache_factory(M::AbstractManifold, o::AbstractManifoldObjective, cache::Tuple{Symbol, Array})
 
 Generate a cached variant of the [`AbstractManifoldObjective`](@ref) `o`
 on the `AbstractManifold M` based on the symbol `cache[1]`,
-where the second element `cache[2]` is an array of further arguments for the cache and
-the third is passed down as keyword arguments.
+where the second element `cache[2]` is are further arguments to  the cache and
```
mateuszbaran, line 773:
```suggestion
where the second element `cache[2]` are further arguments to the cache and
```

src/plans/count.jl
```diff
@@ -0,0 +1,427 @@
...
+    M::AbstractManifold, o::O, s::Symbol, p::P=rand(M)
+) where {P,O<:AbstractManifoldObjective}
+    # vectorial counting cases
+    (s == :EqualityConstraint) && (return length(get_equality_constraints(M, o, p)))
+    (s == :GradEqualityConstraint) && (return length(get_equality_constraints(M, o, p)))
+    (s == :InequalityConstraint) && (return length(get_inequality_constraints(M, o, p)))
+    (s == :GradInequalityConstraint) && (return length(get_inequality_constraints(M, o, p)))
+    # For now this only appears in ProximalMapObjective – so we can access its field
+    (s == :ProximalMap) && (return length(get_objective(o).proximal_maps!!))
+    (s == :StochasticGradient) && (return length(get_gradients(M, o, p)))
```
mateuszbaran, line 98:
```suggestion
    (s === :EqualityConstraint) && (return length(get_equality_constraints(M, o, p)))
    (s === :GradEqualityConstraint) && (return length(get_equality_constraints(M, o, p)))
    (s === :InequalityConstraint) && (return length(get_inequality_constraints(M, o, p)))
    (s === :GradInequalityConstraint) && (return length(get_inequality_constraints(M, o, p)))
    # For now this only appears in ProximalMapObjective – so we can access its field
    (s === :ProximalMap) && (return length(get_objective(o).proximal_maps!!))
    (s === :StochasticGradient) && (return length(get_gradients(M, o, p)))
```

src/plans/count.jl
```diff
@@ -0,0 +1,427 @@
...
+function _get_count(o::AbstractManifoldObjective, ::Val{true}, s, i, m)
+    return get_count(get_objective(o, false), s, i, m)
+end
+
+"""
+    reset_counters(co::ManifoldCountObjective, value=0)
+
+Reset all values in the count objective to `value`.
+"""
+function reset_counters!(co::ManifoldCountObjective, value=0)
```
mateuszbaran, line 190:
```suggestion
    reset_counters(co::ManifoldCountObjective, value::Integer=0)

Reset all values in the count objective to `value`.
"""
function reset_counters!(co::ManifoldCountObjective, value::Integer=0)
```

src/plans/count.jl
```diff
@@ -0,0 +1,427 @@
...
+        else
+            co.counts[s] .= value
+        end
+    end
+    return co
+end
+function reset_counters!(o::AbstractDecoratedManifoldObjective, value=0)
+    return reset_counters!(get_objective(o, false), value)
+end
+function reset_counters!(o::AbstractManifoldObjective, value=0)
```
mateuszbaran, line 203:
```suggestion
function reset_counters!(o::AbstractDecoratedManifoldObjective, value::Integer=0)
    return reset_counters!(get_objective(o, false), value)
end
function reset_counters!(o::AbstractManifoldObjective, value::Integer=0)
```

tutorials/CountAndCache.qmd
```diff
@@ -0,0 +1,128 @@
+---
+title: "How to Cache function calls and Get Started: Optimize!"
```
mateuszbaran, line 2: This appears to somewhat duplicate the `.md` file?

mateuszbaran (commented):
Just a few minor comments. I think the caching looks fine, maybe it could be optimized a bit but likely not much.

docs/src/tutorials/CountAndCache.md
```diff
@@ -0,0 +1,164 @@
...
+Ronny Bergmann
+
+In this tutorial, we want to investigate the caching and counting (i.e. statistics) features
+of [Manopt.jl](https://manoptjl.org). We will reuse the optimization tasks from the
+introductionary tutorial [Get Started: Optimize!](https://manoptjl.org/stable/tutorials/Optimize!.html).
+
+## Introduction
+
+THere are surely many ways to keep track for example of how often the cost function is called,
+for example with a [functor](https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects), as we used in an example in [How to Record Data](https://manoptjl.org/stable/tutorials/HowtoRecord.html)
```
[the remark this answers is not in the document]
kellertuer: This should be changed in the `.qmd` file.

ext/ManoptLRUCacheExt.jl
kellertuer: do you have a preference? It is the value and type o initialise cost or constraint values. So besides points and tangents, we cache arbitrary values – hence the name

tutorials/CountAndCache.qmd
kellertuer: Quite easy: I do not yet have a good way to (a) render quarto on CI (I think that should work now soon-ish since I found some nice features) nor (b) how to properly cache that because tutorials take some time to render and (c) how these caches could be copied into the docs _before_ the docs are rendered. So for now the `.md` files are rendered from these `.qmd` files locally for now and both are checked in.

kellertuer: One could probably change the docs-CI to run quarto upfront and have some artefact cache and such (not that since this PR I know how to run quarto render even on dev versions of Manopt), but it would take me probably 2-3 weeks to get that working in which I would not work on any other code.

ext/ManoptLRUCacheExt.jl
mateuszbaran: I don't really have much preference other than that it should be something longer.

tutorials/CountAndCache.qmd
mateuszbaran: I see, then let's keep it duplicated. Do you want me to re-apply the .md suggestions here?

ext/ManoptLRUCacheExt.jl
kellertuer: Hm, and I am not creative enough to provide something besides `value=` – I thought `v` would be fine besides `p` and `X` though.

mateuszbaran: `p` and `X` are used commonly in the ecosystem so it's an established convention, `v` not so much. Anyway, if you prefer `v` over `value`, then let's keep `v`.

kellertuer: We can also switch to value, since you are right, `v` is not so usual.

tutorials/CountAndCache.qmd
kellertuer: I already just transferred them :)

kellertuer: And in the long run you are right, for best of cases

* Quarto should run before Documenter in CI
* Quarto should have a cache so only changed tutorials should be rerun

PlutoStaticHTML for example managed the first , but I was never able to get the second (and it produced HTML), so at some point I might give this a try.

kellertuer:
🥳 I am now quite happy with this, thanks for motivating me to do all the caches!

Changelog.md
```diff
@@ -5,6 +5,22 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.21] - 22/05/023
```
mateuszbaran, line 8:
```suggestion
## [0.4.21] - 22/05/2023
```

mateuszbaran:
Yes, I think this is a nice new feature :slightly_smiling_face:

kellertuer:
I would like to merge this today. Do you have any objections (for example because you want to take a closer look at something?)

In general I think – In case of a cache-miss we currently do one copy too much: We compute the gradient, store a copy in the cache and return another copy. But to avoid that, we would have to rewrite `get!` I think, which is not what I want to do. Otherwise I think this should even be relatively performant thanks to the NamedTuples and such.

mateuszbaran (approved):
Sure, I think this is ready to be merged. Caching could indeed be slightly optimized with fewer copies but that's not straightforward.

Issue #250, opened 2023-05-22
Introduce Quarto CI


kellertuer:
It would be nice to have the tutorials automatically rendered on CI as well.
For that we would need in the `documentation.yml` CI

* a python dependency for
* Jupyter to run
* Quarto with `quarto render` on the tutorials folder

But also a way to cache the results in order to not render all tutorials on every documenter run, but only if the corresponding `.qmd` file changed.

PR #251, opened 2023-05-22
Typos in CountAndCache.md

DanielVandH:


DanielVandH:
Oh, didn't see it's produced from Quarto! I've made the changes in that file too now, but I don't know how I would render it so if it's not so useful for you no problem.

kellertuer:
Oh, rendering is (sadly) still something performed locally (that is why the md files to exist at all, moving to CI is #250). But rising the typos in both is fine as well.
When now the next time the tutorial is generated from the qmd file, the error is not reintroduced – that is the main thing to care for.

So your solution works fine as well.
As soon as the tests pass we can merge this. Thanks for spotting the typos – and impressive that you opened the PR just hours after we published that version introducing said tutorial.

kellertuer:
Everything looks fine, thanks for spotting them. They will now appear in `/dev` and be part of 0.4.22 when we generate that release.

PR #254, opened 2023-05-27
We can run Quarto on CI.

kellertuer:
to me especially all dependencies to python and Jupyter are not clear whether that is included in any of the quarto CI thingies.


kellertuer:
Hm, the conclusion seems to be that whatever one passes to Quarto it says there is nothing to render – even with the command that from the main directory of the repository does work here locally. I have no clue how this is supposed to work, since the documented versions do not work.

kellertuer:
In principle this works – just that the benchmark tutorial (on my machine 30 seconds) runs for 10+ minutes on CI already). Maybe I'd exclude that if it takes too long (and I can't get caches to work properly).

I also excluded one tutorial from this style of rendering since it requires arymptote and probably some rendering capabilities - which is maybe something one could add later to the ci as well.

kellertuer:
This now just has to wait for #256 to have the CI back as well.

kellertuer:
With the Julia cache and moving to 1.9 the rendering now (when there is nothing new in the tutorials) of the docs is done in 3 minutes now instead of 7 before.

mateuszbaran:
Cool, that's much faster :slightly_smiling_face:

kellertuer:
Maybe I even merge this only to also use the other PR in the next version, since the new tutorial #255 only needs one thorough read and could be included in this scheme easily (actually really by merging both PRs)

PR #255, opened 2023-05-27
Start a Tutorial how to implement a Solver.

kellertuer:


tutorials/ImplementASolver.qmd
```diff
@@ -0,0 +1,340 @@
...
+        copyto!(M, rws.p, rws.q)
+    end
+    return rws
+end
+```
+
+Performance wise we could improve the number of allocations by making `X` also a field of
+our `rws` but let's keep it simple here.
+We could also store the cost of `q` in the state, but we will see how to easily also enable
+this solver to allow for [caching](https://manoptjl.org/stable/tutorials/CountAndCache/#How-to-Count-and-Cache-Function-Calls).
```
mateuszbaran, line 199:
```suggestion
this solver to allow for [caching](https://manoptjl.org/stable/tutorials/CountAndCache/#How-to-Count-and-Cache-Function-Calls).
In practice, however, it is preferable to cache intermediate values like cost of `q` in the state when it can be easily achieved. This way we do not have to deal with overheads of an external cache.
```

tutorials/ImplementASolver.qmd
```diff
@@ -0,0 +1,340 @@
...
+
+solve!(mp, s)
+get_solver_result(s)
+```
+
+The function `solve!` works also in place of `s`, but the last line illustrates how to access
+the result in general; we could also just look at `s.p`, but the function `get_iterate` is
+also used in several other places.
+
+We could for example easilz set up a second solver to work from a specified starting point
```
mateuszbaran, line 230:
```suggestion
We could for example easily set up a second solver to work from a specified starting point
```

tutorials/ImplementASolver.qmd
```diff
@@ -0,0 +1,340 @@
...
+```{julia}
+function random_walk_algorithm!(M::AbstractManifold, f, p=rand(M); kwargs...)
+    mgo = ManifoldCostObjective(f)
+    return random_walk_algorithm!(M, mgo, p; kwargs...)
+end
+```
+
+## Ease of Use II: The State Summary
+
+For the case that zou set `return_state=true` the soplver should return a summary of the run. This is realised providing a `show` method.
```
mateuszbaran, line 303:
```suggestion
For the case that you set `return_state=true` the solver should return a summary of the run. When a `show` method is provided, users can easily read such summary in a terminal.
```

tutorials/ImplementASolver.qmd
```diff
@@ -0,0 +1,340 @@
...
+It should reflect its main parameters, if they are not too verbose an dprovide information
```
mateuszbaran, line 304:
```suggestion
It should reflect its main parameters, if they are not too verbose and provide information
```

tutorials/ImplementASolver.qmd
```diff
@@ -0,0 +1,340 @@
...
+about the reason it stopped and whether this indicates convergence
```
mateuszbaran, line 305:
```suggestion
about the reason it stopped and whether this indicates convergence.
```

mateuszbaran (commented):
A few small improvements :slightly_smiling_face: . The tutorial looks quite nice.

tutorials/ImplementASolver.qmd
```diff
@@ -0,0 +1,340 @@
...
+`f(M,p) = ...`. MOre general it might for example store a gradient function or the Hessian
+or any other information we have about our task.
+
+This is something independent of the solver itself, since it only identifies the problem we
+want to solve independent of how we want to solve it – or in other words, this type contains
+all information that is static and independent of the specific solver at hand.
+
+Usually the problems variable is called `mp`.
+
+### The “Solver” An `AbstractManoptSolverState`
```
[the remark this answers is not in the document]
kellertuer: “ – an” is probably the best choice so

The Solver – An AbstractManoptSolverState

tutorials/ImplementASolver.qmd
kellertuer: Oh, yes, I wrote some par t of that at home on my German keyboard, and the one thing to get consistent then is the switch of z and y on said keyboard ;)

tutorials/ImplementASolver.qmd
```diff
@@ -0,0 +1,340 @@
...
+
+## Implementing the Your solver
+
+There is basically only two methods we need to implement for our solver
+
+- `initialize_solver!(mp, rws)` which initialises the solver before the first iteration
+- `step_solver!(mp, rws, i)` implement the `i`th iteration, where `i` is given to you as the third parameter
+- `get_iterate(rws)` to access the iterate from other places in the solver
+- `get_solver_result(rws)` toi return the final point we reached. Bz default this would
+  return the last iterate, but since our result is `q`, we have to set this
```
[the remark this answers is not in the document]
kellertuer: To be honest I wrote that and the code at the same time, so I even have to read it carefully myself ;)

mateuszbaran (approved):

PR #256, opened 2023-05-30
Switch to `ArrayPartition`

mateuszbaran:
Adapts Manopt to the deprecation in https://github.com/JuliaManifolds/Manifolds.jl/pull/612 . @kellertuer could you regenerate the .md file for GeodesicRegression example? I'm not sure how I should run quarto to do that.


kellertuer (approved):
I think it should be fine to combine this in a version with the new CI, so we could merge this without a version bump, I think.

mateuszbaran:
Sure, that makes sense. Tests are running fine :tada: .

kellertuer:
Yes, restarting them was enough after the new Manifolds.jl was there. Will still wait for the MacOS ones before we merge here.

PR #258, opened 2023-06-02
🎓Extend the Count and Cache Tutorial to also cover exchanging interims results.

kellertuer:


kellertuer:
Hmpf. This quarto thigh is still quite wobbly, now IJulia does not compile on CI, might have to remove a cache maybe?

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
 ```{julia}
 get_solver_result(r)
 ```
+
+## Advanced Caching Examples
+
+Besides caching single calls to elements of the objective.
+A specific scenario might be, that you want to cache intermediate results
+of computing the cost and share that with the gradient. We will present
+three solutions to this:
```
mateuszbaran, line 132:
```suggestion
There are more options other than caching single calls to specific parts of the objective.
For example you may want to cache intermediate results
of computing the cost and share that with the gradient computation. We will present
three solutions to this:
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+The small disadvantage might be, that we always compute _both_, the gradient and the cost. Luckily, the cache we used before, takes this into account and caches both results, such that we indeed end up computing `A*p` only once when asking to a cost and a gradient.
+
+Let's compare both methods
+```{julia}
+p0 = [(1/5 .* ones(5))..., zeros(m-4)...];
+@time s1 = gradient_descent(N, g, grad_g!, p0;
+    stopping_criterion = StopWhenGradientNormLess(1e-5),
+    evaluation=InplaceEvaluation(),
+    count=[:Cost, :Gradient],
+    cache=(:LRU, [:Cost, :Gradient], 25),
```
mateuszbaran, line 190: For `gradient_descent` this looks like way too long cache. I think a simple cache should be enough.

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+    p::T
+end
+function (g::StorageG)(::Val{:Cost}, M, p)
+    if !(p==g.p) #We are at a new point -> Update
+        g.Ap .= g.A*p
+        g.p .= p
+    end
+    return -g.p'*g.Ap
+end
+function (g::StorageG)(::Val{:Gradient}, M, X, p)
```
mateuszbaran, line 243:
```suggestion
function (g::StorageG)(::Val{:Cost}, M::AbstractManifold, p)
    if !(p==g.p) #We are at a new point -> Update
        g.Ap .= g.A*p
        g.p .= p
    end
    return -g.p'*g.Ap
end
function (g::StorageG)(::Val{:Gradient}, M::AbstractManifold, X, p)
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+    if !(p==g.p) #We are at a new point -> Update
+        g.Ap .= g.A*p
+        g.p .= p
+    end
+    copyto!(M, X, p, -2.0 .* g.Ap)
```
mateuszbaran, line 248:
```suggestion
    X .= -2 .* g.Ap
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+    project!(M, X, p, X)
+    return X
+end
+```
+
+Here we use a first parameter do distinguish both functions. For the mutating
```
mateuszbaran, line 254:
```suggestion
Here we use the first parameter to distinguish both functions. For the mutating
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+case the signatures are even different, but for an allocating case, the signatures of the cost and the gradient function are the same.
```
mateuszbaran, line 255:
```suggestion
case the signatures are different regardless of the additional argument but for the allocating case, the signatures of the cost and the gradient function are the same.
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+
+### A shared storage approach using a functor
+
+So compared to the previous approach, which might be beneficial.
+An alternative is the usage of a functor, that introduces a “shared storage”
+of the result of computing `A*p`. We additionally have to store `p` though,
+since we have to check that we are still evaluation the cost and/or gradient
+at the same point, respectively.
+We again consider the (more efficient) inplace variant.
+The can be done as follows
```
mateuszbaran, line 228:
```suggestion
An alternative to the previous approach is the usage of a functor that introduces a “shared storage”
of the result of computing `A*p`. We additionally have to store `p` though,
since we have to check that we are still evaluating the cost and/or gradient
at the same point at which the cached `A*p` was computed.
We again consider the (more efficient) inplace variant.
The can be done as follows
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+    Ap = get!(g.cache, copy(M,p)) do
+        g.A*p
+    end
+    return -p'*Ap
+end
+function (g::CacheG)(::Val{:Gradient}, M, X, p)
+    Ap = get!(g.cache, copy(M,p)) do
+        g.A*p
+    end
+    copyto!(M, X, p, -2.0 .* Ap)
```
mateuszbaran, line 304:
```suggestion
    X .= -2 .* Ap
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,222 @@ result as usual:
...
+and for safety let's check that we are reasonably close
+
+```{julia}
+p4 = get_solver_result(s4)
+g(N, p4) - f_star
+```
+
+For this example, or maybe even [`gradient_descent`](@ref) in general it seems,
+the last this additional cache does not improve the result further,
+it is one the same effort both time and allocation-wise.
```
mateuszbaran, line 338: What does "the last" here refer to?

mateuszbaran (commented):
Nice tutorial :+1: . I'm only surprised that you chose a large LRU cache for gradient descent.

kellertuer:
> Nice tutorial 👍 . I'm only surprised that you chose a large LRU cache for gradient descent.

The size of the cache was set basically by random (not far away from head-hits-keyboard-random-number-generator).

tutorials/CountAndCache.qmd
kellertuer: Indeed, here one could take the simple cache as well. Will try.

tutorials/CountAndCache.qmd
kellertuer: I am never sure when to use which, so I trust you here.

tutorials/CountAndCache.qmd
kellertuer: that is just too much in that line.

tutorials/CountAndCache.qmd
kellertuer: So `SimpleCache` multiplies the evaluations by a factor of 3, sure maybe a 5er cache is fine as well and we do not need 2, but just the `SimpleCache` is not enough.

tutorials/CountAndCache.qmd
kellertuer: rephrased it.

tutorials/CountAndCache.qmd
mateuszbaran: Weird, though let's use LRU cache then.

tutorials/CountAndCache.qmd
mateuszbaran: Thanks, it reads better now.

tutorials/CountAndCache.qmd
kellertuer: I am a bit surprised as well, but there is several places that access the gradient for example.

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,221 @@ result as usual:
...
+#Define the new functor
+storage_g = StorageG(A, zero(p0), zero(p0))
+# and cost and gradient that use this functor as
+g3(M,p) = storage_g(Val(:Cost), M, p)
+grad_g3!(M, X, p) = storage_g(Val(:Gradient), M, X, p)
+@time s3 = gradient_descent(N, g3, grad_g3!, p0;
+    stopping_criterion = StopWhenGradientNormLess(1e-5),
+    evaluation=InplaceEvaluation(),
+    count=[:Cost, :Gradient],
+    cache=(:LRU, [:Cost, :Gradient], 25),
```
mateuszbaran, line 267:
```suggestion
    cache=(:LRU, [:Cost, :Gradient], 2),
```
I've checked, 2 is enough but 1 is not.

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,221 @@ result as usual:
...
+This of course still yields the same result
+
+```{julia}
+p3 = get_solver_result(s3)
+g(N, p3) - f_star
+```
+
+And while we again have a split of the cost and gradient evaluations, we can observer that the allocations are less thatn half of the previous approach.
+
+### A local Cache approach
```
mateuszbaran, line 281:
```suggestion
### A local cache approach
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,221 @@ result as usual:
...
+```
+
+This of course still yields the same result
+
+```{julia}
+p3 = get_solver_result(s3)
+g(N, p3) - f_star
+```
+
+And while we again have a split of the cost and gradient evaluations, we can observer that the allocations are less thatn half of the previous approach.
```
mateuszbaran, line 279:
```suggestion
And while we again have a split off the cost and gradient evaluations, we can observe that the allocations are less than half of the previous approach.
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,221 @@ result as usual:
...
+
+### A local Cache approach
+
+This variant is very similar to the previous one, but uses a whole cache instead
+of just one place to store `A*p`. This makes the code a bit nicer,
+and it is possivle to store more than just the last `p` either cost or gradient
```
mateuszbaran, line 285:
```suggestion
and it is possible to store more than just the last `p` either cost or gradient
```

tutorials/CountAndCache.qmd
```diff
@@ -122,3 +123,221 @@ result as usual:
...
 ```
+
+## Advanced Caching Examples
+
+There are more options other than caching single calls to specific parts of the objective.
+For example you may want to cache intermediate results
+of computing the cost and share that with the gradient computation. We will present
+three solutions to this:
+
+1. An easy approach from within `Manopt.jl`: The [`ManifoldCostGradientObjective`]
```
mateuszbaran, line 134: You probably missed link address here?

mateuszbaran (approved):
OK, with these minor points I think it's a nice description of caching :slightly_smiling_face:

tutorials/CountAndCache.qmd
kellertuer: Actually just a `(@ref)` missing since it end up in the docs.

PR #259, opened 2023-06-03
Allow the stochastic cost to be a vector of costs.

### LINKED ISSUE #249: Splitting stochastic gradient descent objective
Currently the stochastic gradient descent objective is a single function, in contrast with multi-part gradient. It would be reasonable to also return parts of the objective separately that correspond to specific parts of gradient. That would enable for example using line searches to determine step size.

In your example that would correspond to changing 
```julia
F(M, p) = 1 / (2 * n) * sum(map(q -> distance(M, p, q)^2, data))
```
to
```julia
Fs = [(M, p) -> 1 / 2 * distance(M, p, q)^2 for q in data]
```

kellertuer:
I first misread even #249 - but now this resolves #249, it was only necessary to implement `get-cost` for this case and extend the type in the objective (one could even simplify the existing ones).

Do we need more functions here, e.g. a `get_cost(M, ago, p, i)`? It would only work in this vector case, though (as opposed for the gradient where it works in both cases, since both return a vector).


mateuszbaran:
Given that primary use of stochastic gradient descent is something like machine learning where we want to avoid calculation of the entire cost function, I think `get_cost(M, ago, p, i)` is needed. In the non-vector case `get_cost(M, ago, p, 1)` could default to `get_cost(M, ago, p)` perhaps?

mateuszbaran:
Either an error or just `get_cost(M, ago, p)` -- both sound fine. I think it should be consistent with what we do in the vector case when `i` is greater than the number of components.

kellertuer:
there I do not implement anything and the vector access fails :)

mateuszbaran:
OK, so `i>1` should be an error in the non-vector case.

kellertuer:
But the `I=1` could also directly error since we are not in a vector case?

mateuszbaran:
Julia allows getting first element of a scalar so for convenience allowing it here makes sense. This way stochastic optimization code could always use `get_cost` with an index, both in the scalar and the vector case. In particular I think stepsize selection and stopping criteria should have access to the current batch number `i`, for example stochastic gradient norm stopping criterion could refer to gradient for the current `i` instead of the overall gradient.

kellertuer:
That sounds reasonable. Then I will do the new cost as you said.

mateuszbaran (approved):

PR #260, opened 2023-06-04
Fix technical details in doc

blegat:
Not so familiar with the `qmd` format so I just followed
https://github.com/JuliaManifolds/Manopt.jl/blob/a9b0f2c2f8f119df292c2d110bfc332ed4a24498/tutorials/ImplementASolver.qmd#L16-L19


kellertuer:
Wow, nice! Yes I was struggling with the admonitions from Documenter (and how to write them in `.qmd` for a while and the one you noticed is just a few days old. I did not yet have the time to check old admonitions – so great that you directly have a PR to fix them 👍

PR #264, opened 2023-06-06
JuMP interface

blegat:
This is a proof of concept implementation of MathOptInterface so that this package can be used from JuMP.
Here are a few things that are missing:

- [x] How do I get the vectorization of a Manifold ? In JuMP, every set is represented as `Vector{Float64}`. For PSD matrices, it's the vectorization of the upper triangle for instance.
- [x] How do I get the gradient given the ambient gradient ? For a Riemanian submanifold, it should be the orthogonal projection to the tangent space, how do I compute that ?
- [x] Add support `MOI.RawOptimizerAttributes`
- [x] Add support for choosing optimization algorithm (not just hardset gradient as of the current state of the PR)
- [x] `RawStatusString`
- [x] Run against `MOI.Test.runtests`
- [x] Reshape inside `Optimizer` and test with a manifold with a matrix representation

Here are a few questions:

- [x] What is the difference between a decorated state/problem and the non-decorated one ? Is it ok if I use the decorated one to return the result to the user ?
- [x] How do I get the objective value from the problem and state ? Waiting for https://github.com/JuliaManifolds/Manopt.jl/pull/266
- [x] How do I get the primal status from the problem and state ? https://jump.dev/MathOptInterface.jl/dev/reference/models/#MathOptInterface.ResultStatusCode
- [x] How do I get the termination status from the problem and state ? https://jump.dev/MathOptInterface.jl/dev/reference/models/#MathOptInterface.TerminationStatusCode

**Note**: Requires https://github.com/jump-dev/JuMP.jl/pull/3106 for nonlinear objective because `NLPBlock` is incompatible with constrained variables because of https://github.com/jump-dev/MathOptInterface.jl/blob/32dbf6056b0b5fb9d44dc583ecc8249f6fd703ea/src/Utilities/copy.jl#L485-L500 but should already be useable from MOI

PR #266, opened 2023-06-07
A collection of small improvements

kellertuer:
While writing some answers in #264 I noticed that

* `get_stoping_criterion` might be nice
* `indicated_convergence` was not documented
* `get_state` would be nicer when it also had the recursive parameter (like `get_objective`)
* `get_state` replaces a few access functions

Internally

* maybe it would be reasonable to switch the state also to a `AbstractDecoratedManoptState` pattern like the objective, since there seems not much sense to keep the dispatch-with-THTT approach, since with an abstract decorator type it would be nicer to write _and_ unified with the objective. -> this is a bit more work for later
* one could also implement a more. precise way for convergence than just the string returned by `get_reason`. – maybe after we have MOI states internally (with their dependency) but not before.

I will collect a few small internal improvements here for a while.


src/plans/solver_state.jl
```diff
@@ -121,16 +135,26 @@ function get_solver_return(o::ReturnManifoldObjective, s::AbstractManoptSolverSt
 end
 
 @doc raw"""
-    get_state(s::AbstractManoptSolverState)
+    get_state(s::AbstractManoptSolverState, recursive=truue)
```
mateuszbaran, line 138:
```suggestion
    get_state(s::AbstractManoptSolverState, recursive::Bool=true)
```

src/plans/solver_state.jl
```diff
@@ -121,16 +135,26 @@ function get_solver_return(o::ReturnManifoldObjective, s::AbstractManoptSolverSt
...
-return the undecorated [`AbstractManoptSolverState`](@ref) of the (possibly) decorated `s`.
-As long as your decorated state store the state within `s.state` and
-the [`dispatch_state_decorator`](@ref) is set to `Val{true}`,
-the internal state are extracted.
+If `recursive` is set to `false`, only the most outer decorator is taken away instead of all.
 """
-get_state(s::AbstractManoptSolverState) = _get_state(s, dispatch_state_decorator(s))
-_get_state(s::AbstractManoptSolverState, ::Val{false}) = s
-_get_state(s::AbstractManoptSolverState, ::Val{true}) = get_state(s.state)
+function get_state(s::AbstractManoptSolverState, recursive=true)
```
mateuszbaran, line 151:
```suggestion
function get_state(s::AbstractManoptSolverState, recursive::Bool=true)
```

src/plans/solver_state.jl
```diff
@@ -56,6 +56,20 @@ _get_message(s::AbstractManoptSolverState, ::Val{true}) = get_message(s.state)
 #INtroduce a default that there is no message
 _get_message(s::AbstractManoptSolverState, ::Val{false}) = ""
 
+@doc raw"""
+    get_stopping_criterion(ams::AbstractManoptSolverState)
+
+return the [`StoppingCriterion`](@ref) stored within the [`AbstractManoptSolverState`](@ref) `ams`.
```
mateuszbaran, line 62:
```suggestion
Return the [`StoppingCriterion`](@ref) stored within the [`AbstractManoptSolverState`](@ref) `ams`.
```

docs/src/index.md
```diff
@@ -8,13 +8,13 @@ CurrentModule = Manopt
...
 
-For a function $f:\mathcal M → ℝ$ defined on a [Riemannian manifold](https://en.wikipedia.org/wiki/Riemannian_manifold) $\mathcal M$ we aim to solve
+For a function ``f:\mathcal M → ℝ`` defined on a [Riemannian manifold](https://en.wikipedia.org/wiki/Riemannian_manifold) ``\mathcal M`` we aim to solve
 
 ```math
 \operatorname*{argmin}_{p ∈ \mathcal M} f(p),
 ```
 
-or in other words: find the point $x$ on the manifold, where $f$ reaches its minimal function value.
+or in other words: find the point ``x`` on the manifold, where ``f`` reaches its minimal function value.
```
mateuszbaran, line 17:
```suggestion
or in other words: find the point ``p`` on the manifold, where ``f`` reaches its minimal function value.
```

kellertuer:
Hm, I am a bit confused here. I could need this PR a lot on ManoptExamples just now but it seems my most recent commit made that every first run of the CI fails but every second one passes? Strange.

PR #276, opened 2023-07-04
a few typos in comments

flgoyens:


kellertuer:
Thanks for noticing the fixes and great that you are using Manopt.jl :)

Your fix of the stopping criterion typo broke one test, where we use the reason string to check that the right stopping criterion was met.
Could you check in the file `test/solvers/test_truncated_cg.jl` line 21 and fix the typo here as well?

flgoyens:
Hi @kellertuer ! 

I was indeed surprised to break something. Thanks for pointing me towards the fix, I'll get to that :)

Best
Florentin

kellertuer:
Well, we test also the show methods you changed; but sure we could have also spotted the typo when writing the test ;)

kellertuer:
Thanks for the fix; now this can be merged and will be part of the next version. Thanks for your PR :)

PR #277, opened 2023-07-05
Adaptive Regularisation with Cubics (ARC) Solver

kellertuer:
This PR is a slight rework from the work @mathiasrm1 did in his master thesis this semester. Until now

* [x] Variables are renamed according to other solvers
* [x] some temporary variables were introduced to avoid allocations
* [x] Lanczos iteration optimised a bit
* [x] The Cubic subproblem cost and gradient implemented
* [x] add middle-layer solver interface `adaptive_regularization_with_cubics(M, mho, ...; kwargs...)`
* [x] allow / test a CG subsolver as well  (cost and grad already existing
* [x] write a documentation page
* [x] write a tutorial comparing trust region and ARC-Lanczos, and ARC-CG
* [x] check the value theta (maybe in a Newton State?)
* [x] make Newton a state so it can have proper stopping criteria.
* [x] there seems to also be a problem with the second order progress stopping criterion as to when to stop Lanczos. It currently often uses its maximal iterations. The only reason the outer one did not hit, was that it checked versus one too less.
* [x] write tests
* [x] check that the subproblem can be solved with CG on the tangent space.
* [x] write a `status_summary` for the new state.
* [x] improve naming and availability of the sub solver stopping criteria
* [x] carefully check rendered docs
* [x] ~~maybe implement the exact best step size for the gradient descent subsolver~~


kellertuer:
The only thing to maybe consider is, that for the gradient based sub solvers of ARC, there is 
https://github.com/NicolasBoumal/manopt/blob/master/manopt/solvers/arc/solve_along_line.m
which can be considered an exact step size computation. But it is not documented where that is from nor why that holds, so it is maybe not so easy to generalise / transfer here as a step size.
The only effect for now is, that here the gradient based methods require a few more steps than the Lanczos solver.

kellertuer:
Currently code coverage fails because we have one spurious line missed on code not changed here. All lines coded here are covered by tests.

kellertuer:
Let's leave the optimised step size for a later PR.

PR #281, opened 2023-07-13
Use DocumenterCitations.jl for References

### LINKED ISSUE #280: Unify citations to DocumenterCitations.jl
https://github.com/JuliaDocs/DocumenterCitations.jl

just went 1.0, has `citet` and such supports _and_ bibliography per page possibilities, so that I will definelty try this out in order to finally get unified citations, since currently we use probably 2-3 styles and an extra style in the tutorials.

kellertuer:
This changes all references to using a (global) references bibtex file (maybe later CSL-YAML?), so that both the quarto tutorials and the documenter references are based on the same source

What is still to do is to check all tutorials and documentation pages for either `[@CiteKey]` or `[^citekey]` code, look up the reference and rewrite the corresponding text as done exemplarily for ALM and EPM already.
This is a bit of work but I think worth it, since we also gain a global references list.

This resolves #280.

* [x] check all references that are in footnotes
* [x] check all references from the tutorials which for now used the quarto system


kellertuer:
This is not perfect (die to BibTeX compared to a nicer BibLaTeX) but should work so far and is a good basis, once a BibLaTeX renderer should be available.

kellertuer:
I will not bump version here, since I would like to use this also with ARC docs – and then we can also release both features together as a new patch version.

PR #285, opened 2023-08-11
Introduce a :Subsolver debug symbol and a DebugWhenActive

### LINKED ISSUE #189: Improve subsolver debug output
Here is one interesting thing I noticed:
The sub solver debug output (for example in ALM) currently prints every iteration – it would be good to adapt that to be able to take the “subsampling” in output from the outer iteration.

I am just not yet sure how to realise that.

Similarly when only printing every n-th iteration and the cost increases in between, the warning does not contain the right iteration number.

kellertuer:
This way adding :Subsolver to a sub solvers debug keyword – the debug is only printed when the parent solver is “active” as well; for now this is mainly corresponding to `DebugEvery` being on an active iteration.

* [x] unification and documentation of the new `set_manopt_parameter!` variants
* [x] documentation
* [x] Testing
* [x] a Tutorial, that illustrates this behaviour.

This resolves #189 (since I saw a 1 hour idea how to resolve this finally).


kellertuer:
I am a bit confused about documenter on CI, since it should run with the current code base where the error that is thrown is actually fixed. Will have to check, why that is happening on the new tutorial.
So locally on my machine that runs fine and it looks like it has Manopt in dev mode so than that error should be fixed but somehow is not. Me very confused.

PR #286, opened 2023-08-15
Introduce an EmbeddedObjective

### LINKED ISSUE #217: Implement a ManifoldEuclideanGradientObjective
I am not yet sure how to provide this easily in the high-end interfaces, but with the new structure in Manopt.jl 0.4 it is easy to provide a `ManifoldEuclideanGradientObjective`, that is a cost function $f$ together with its _Euclidean_ gradient (usually called egrad in Manopt), since every gradient call would just be encapsulated in a projection onto the tangent space and a `change_representer` call.

The main task here is a nice way to handle this in the high-level interfaces. One way could be to extend the evaluation types as follows:

* Introduce `AbstractAllocatingEvaluation` and `AbstractInplaceEvaluation` as intermediate types that the current ones inherit from
* Introduce further subtypes `AllocatingEuclideanEvaluation` and `InplaceEuclideanEvaluation`

this makes it easy to specify that gradients (and later Hessians) are defined in the Embedding/Euclidean sense and the corresponding  Objectives should be taken. One could maybe even just adapt the `ManifoldGradientObjective` in this way without introducing a new objective.

kellertuer:
This PR resolves #217 by proposing an `EmbeddedManifoldObjective`.

## Idea
* the new objective is a wrapper
* before calling the cost of the original objective it does an `embed(M,p)`
* ...and similarly before calling `get_gradient` – but then also calls `riemannian_gradient` on the gradient the original objective returns

## Interface
since we have the `decorate_objective!` chain already available, just introducing the `objective_type=:Euclidean` keyword there makes it possible for any objective to be wrapped. Currently also `:Embedded` is equivalent, since it actually works more general then just Euclidean – but `:Euclidean` is easier to understand for most users.

## Internals
The `EmbeddedManifoldObjective` has two (optionally used) fields `p` and `X` for a point and tangent vector in the embedding. That way one can use `embed!` or `get_gradient!` in the embedding. This should reduce allocations tremendously.

## ToDo
* [x] `get_cost`
* [x] `get_gradient` / `get_gradient!`
* [x] `get_hessian` / `get_hessian!`
* [x] `get_grad_equality_constraint` / `get_grad_equality_constraint!` (the the plural)
* [x] `get_grad_inequality_constraint` / `get_grad_inequality_constraint!` (and the plural)
* [x] More documentation
* [x] Testing
* [x] A Tutorial


src/plans/objective.jl
```diff
@@ -47,23 +47,155 @@ do not allocate memory but work on their input, i.e. in place.
...
+@doc raw"""
+    get_gradient(M, emo::EmbeddedManifoldObjective, p)
+    get_gradient(M, X, emo::EmbeddedManifoldObjective, p)
+
+Evaluate the gradient function of an objective defined in the embedding, that is
+call [`embed`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions/#ManifoldsBase.embed-Tuple{AbstractManifold,%20Any})
+on the point `p` and call the original cost on this point.
+And convert the gradient using [`riemannian_gradient`]() on the result.
+
+if `emo.p` is not nothing, the embedding is done in place of `emo.p`
```
mateuszbaran, line 158:
```suggestion
If `emo.p` is not `nothing`, the embedding is done in place of `emo.p`.
```

src/plans/objective.jl
```diff
@@ -47,23 +47,155 @@ do not allocate memory but work on their input, i.e. in place.
...
+    return get_cost(get_embedding(M), emo.objective, emo.p)
+end
+
+function get_cost_function(emo::EmbeddedManifoldObjective)
+    return (M, p) -> get_cost(M, emo, p)
+end
+
+@doc raw"""
+    get_gradient(M, emo::EmbeddedManifoldObjective, p)
+    get_gradient(M, X, emo::EmbeddedManifoldObjective, p)
```
mateuszbaran, line 151:
```suggestion
    get_gradient!(M, X, emo::EmbeddedManifoldObjective, p)
```

src/plans/objective.jl
```diff
@@ -47,23 +47,155 @@ do not allocate memory but work on their input, i.e. in place.
...
+end
+
+@doc raw"""
+    get_cost(M, emo::EmbeddedManifoldObjective, p)
+
+Evaluate the cost function of an objective defined in the embedding, that is
+call [`embed`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions/#ManifoldsBase.embed-Tuple{AbstractManifold,%20Any})
+on the point `p` and call the original cost on this point.
+
+if `emo.p` is not nothing, the embedding is done in place of `emo.p`
```
mateuszbaran, line 135:
```suggestion
If `emo.p` is not `nothing`, the embedding is done in place of `emo.p`.
```

src/solvers/solver.jl
```diff
@@ -64,12 +64,13 @@ decorate the [`AbstractManifoldObjective`](@ref)` o` with specific decorators.
...
-* `cache` – (`missing`) specify a cache. Currenlty `:Simple` is supported and `:LRU` if you
+* `cache`     – (`missing`) specify a cache. Currenlty `:Simple` is supported and `:LRU` if you
   load `LRUCache.jl`. For this case a tuple specifying what to cache and how many can be provided,
   i.e. `(:LRU, [:Cost, :Gradient], 10)`, where the number specifies the size of each cache.
   and 10 is the default if one omits the last tuple entry
-* `count` – (`missing`) specify calls to the objective to be called, see [`ManifoldCountObjective`](@ref) for the full list
-
+* `count`     – (`missing`) specify calls to the objective to be called, see [`ManifoldCountObjective`](@ref) for the full list
+* `objective` - (`missing`) specify that an objective is `:Euclidean`, which is equivalent to specifying the default embedding
+                the equivalent statement here is `:Embedded`, which fits better in naming when `M` is an []`
```
mateuszbaran, line 73: That's not a particularly clear explanation IMO. What `[]` was supposed to be? Why have both `:Euclidean` and `:Embedded` when there is no difference?

mateuszbaran (commented):
I think the general idea looks good, I think just a few minor comments.

> This PR resolves https://github.com/JuliaManifolds/Manopt.jl/issues/217 by proposing an EmbeddedObjective– maybe one could also better all it EmbeddedManifoldObjective?

I think `EmbeddedManifoldObjective` would be more consistent with the other names?

> which functions can we extend this to? Hessian for sure, maybe also proximal maps.

Is there actually a generic way to convert a Euclidean proximal map to a Riemannian one?

src/solvers/solver.jl
kellertuer: Well-spotted – I was maybe already in the phase of “Let's first see whether the idea is good”. I meant of course to refer to the `EmbeddedManifold`, since this wrapper would not only work for Euclidean embeddings but for any

kellertuer:
> I think the general idea looks good, I think just a few minor comments.
> 
> > This PR resolves #217 by proposing an EmbeddedObjective– maybe one could also better all it EmbeddedManifoldObjective?
> 
> I think `EmbeddedManifoldObjective` would be more consistent with the other names?

That's what I thought as well when already writing this PR ;)

> 
> > which functions can we extend this to? Hessian for sure, maybe also proximal maps.
> 
> Is there actually a generic way to convert a Euclidean proximal map to a Riemannian one?

You might be right, that that is not the case – but the constraints and their gradients, the differentials, they should be possible as well.

src/solvers/solver.jl
kellertuer: `objective_domain` sounds more precise, that is correct and “missing” would mean “manifold/inherit”, we could maybe even go to `:Manifold` as default then.

kellertuer: The current amount of documentation was mainly done for me to structure (and not forget) my own ideas/thoughts here ;)

mateuszbaran: OK, `:manifold` sounds fine as well.

kellertuer: I went for `:Riemannian` as compared to `:Euclidean` in the symbol here.

mateuszbaran: OK, and what about `objective_domain`?

kellertuer: Oh sorry, forget that. Will do that after my next meeting :) 

Though with `Riemannian` and `Euclidean` objective itself would actually also make sense already, but you are right, domain is more precise here.

kellertuer: Hm, now that I do use `:Riemannian`, `:Euclidean` as symbols, I feel that (similar to gradient and hessian) just stating it is a Riemanian/Euclidean objective should be fine enough (without domain) and still be precise enough.

A Riemannian objective is one that has a Riemannian cost/grad/Hessian
A Euclidean objective has a Euclidean cost/grad/objective (where to be more precise the cost just needs and embed for p before calling the inner cost).

With that, do you thing `objectvie=:Riemannian` would be fine enough?

kellertuer:
I think I now also sketched the hessian, before doing that for the constraints – since that is quite some work (though more in writing cases than in much thinking) I would like to discuss/check whether

* the `get_gradient` cases can maybe be simplified
* whether it might be reasonable to have a second embedded vector in the objective for the Euclidean Hessian (so that the Euclidean Hessian and the embedded direction can be computed without allocations).

src/solvers/solver.jl
mateuszbaran: I still find it a bit weird that there is the actual objective passed an argument and an argument called `objective` that isn't it.

mateuszbaran: So, maybe just call it `domain`? After all, it applies to objective, its gradient and Hessian (when present).

kellertuer: That sounds reasonable, but for me domain would be a set and miss the information whether its Euclidean or not? `domain=:Euclidean` and `domain=:Riemannian` sound/read a bit strange?

mateuszbaran: Hm... what about `embedded::Bool=false`?

kellertuer: I like that one better, just that my idea of making it easier for Manopt(Matlab) people of thinking of the Euclidean objective vanishes in this bool's name. I am also not sure whether we might have other things we want to “transport” in this information? But we could also just to more keywords then, sure.

mateuszbaran: I'm out of ideas so let's maybe just stick with what you prefer.

kellertuer:
In order to have a nice tutorial, this probably also should wait for https://github.com/JuliaManifolds/Manifolds.jl/pull/644

src/solvers/solver.jl
kellertuer: What about `objective_type`?

We could either let `objective` be a nice keyword that `objective_type` defaults to (that way the outer one is nicer but the inner one is not confusing) – or do `objective_type` on both. I feel that `:Riemannian` specifies a bit more the type of objective (than its domain).

mateuszbaran: `objective_type` sounds fine :+1:

kellertuer: Perfect, will change that here. Then first focus on ManifoldDiff, then Manifolds and then only return to this one here.

I think I really now understood most of that conversion stuff (just maybe not yet whether it works as well for non-isometrically embedded ones, we will see ;))

kellertuer:
This is just missing a bit of testing, but the tutorial is at least basically finished – it needs https://github.com/JuliaManifolds/Manifolds.jl/pull/644, and I have to check while for now the Hessian conversion on Grassmann is not (yet) correct.

tutorials/EmbeddingObjectives.qmd
```diff
@@ -0,0 +1,189 @@
...
+
+While they look reasonable here and were already derived – for the general case this derivation
+might be more complicated.
+
+Luckily there exist two functions in [`ManifoldDiff.jl`](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/) that are implemented for several
+manifolds from [`Manifolds.jl`](https://github.com/JuliaManifolds/Manifolds.jl), namely [`riemannian_gradient`](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/library/#ManifoldDiff.riemannian_gradient-Tuple{AbstractManifold,%20Any,%20Any})`(M, p, eG)` that converts a Riemannian gradient
+`eG=`$\nabla \tilde f(p)$  into a the Riemannain one $\operatorname{grad} f(p)$
+ and [`riemannian_Hessian`](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/library/#ManifoldDiff.riemannian_Hessian-Tuple{AbstractManifold,%20Any,%20Any,%20Any,%20Any})`(M, p, eG, eH, X)
+ which converts the Euclidean Hessian `eH=`$\nabla^2 \tilde f(p)[X]$  into $\operatorname{Hess} f(p)[X]$,
+ where we also require the Euclidean gradient `eG=`$\nabla \tilde f(p)$.
```
mateuszbaran, line 96: There are some formatting errors around this part.

kellertuer: Thanks for spotting, for now I am still fighting interactions of sub solvers and embeddings ;)

tutorials/EmbeddingObjectives.qmd
```diff
@@ -0,0 +1,189 @@
...
+    return_objective=true,
+    return_state=true,
+)
+q1 = get_solver_result(r1)
+r1
+```
+
+but if you choose to go for the conversions, then, thinking of the embedding and defining two new functions
+might betedious, there is a shortcut for these, which performs the change internally, when necessary
+by specifying `objective_type=:Euclidean`.
```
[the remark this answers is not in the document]
kellertuer: For now that tutorial was write only, I did not yet read it, it was more to have example code, so before you spent too much time on it (feel free to do so, but it might have quite a few quirks still).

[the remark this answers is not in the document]
mateuszbaran: (after all, you've put the ready-for-review label on the PR)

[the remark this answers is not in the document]
kellertuer: Yes, I should have removed that again before I entered todays afternoon – sorry for that. I was hoping to give it a read today at 10, but with a few meetings and quite some debugging, I think it is working now, but I did not yet read the text – feel free to read :D but I will only do that after a few more administrative duties now.

kellertuer:
Sorry for the tag a bit too early, I missed the “automatic conversion to embedded objectives” for sub solvers (and was wondering why embedded worked so poorly with ARC).

kellertuer:
One thing I am a bit confused about it, that on https://manoptjl.org/previews/PR286/tutorials/EmbeddingObjectives/ the second run (with the euclidean ones converted) sometimes does work (yesterday evening) sometimes not (today) – on my machine that consistently needs 7 iterations and not 40 (which is the ARC default max iter).

Not sure what the CI is doing there.

kellertuer:
Ah, the main point is that I am stupid. Quarto renders on the last registered version. So sure that is the old one and hence fails. I will trick that a bit for this tutorial now.

PR #294, opened 2023-09-25
Make the trust region subsolver exchangeable.

### LINKED ISSUE #240: Make the sub solver of trust regions replaceable

kellertuer:
This should resolve #240 and unifies notation as well.

To keep track:

* `η` the subsolver iterate is now called `Y`
* `η_Cauchy` the Cauchy point to check for the randomised variant is now called `Z`
* `random` will not only be true/false but will allow to set the amount of randomness
* documentation updated as well

This is still an early stage, but I wanted to keep track here. Especially, setting initial values and checking the radius afterwards are still quite intertwined.


kellertuer:
The decoupling is finished – though I would like to maybe improve this a bit further by also introducing a sub_problem – which also means to rewrite TCG a bit to not use the HessianProblem but the new subproblem actually.

Further I will try to provide a second subsolver just to further check the functionality.

But:
The code by now is unified to the rest of Manopt and simplified in several places.

kellertuer:
While this rework per se is done (besides the idea above to add a second subsolver), we still miss a `sub_problem` in the trust region solver for full generality; since that is defined in a tangent space, I would like to wait for https://github.com/JuliaManifolds/ManifoldsBase.jl/pull/169 to be merged, though this might still take a bit of time.

kellertuer:
This still needs a bit of rework, for best of cases also for the other states with sub solvers to be unified in interface and even easy to use with default sub solvers.

kellertuer:
The redesign should be finished, but a few tests run now longer and some are innacurate – maybe I introduced some side-effect or the subsolver is not correctly initialised,
but it is a proper subsolver by now.

kellertuer:
I think one can basically reuse Lanczos, since for ARC it is doing that for an even more involved sub problem. I just have to check how to use that in common and maybe use the chance to also improve the sub solver of Lanczos (Newton) a bit.

This might take a while though, since the current form of Lanczos is a bit involved – i.e. quite closely tied to ARC.

kellertuer:
Hm, in reworking Lanczos / The sub solvers Objective I seem to have introduced a bug that is really hard to narrow down where that comes from.

test/solvers/test_adaptive_regularization_with_cubics.jl
```diff
@@ -126,7 +122,14 @@ include("../utils/example_tasks.jl")
...
         p3 = adaptive_regularization_with_cubics(
-            M, f, grad_f, p0; θ=0.5, σ=100.0, retraction_method=PolarRetraction()
+            M,
+            f,
+            grad_f,
+            p0;
+            θ=0.5,
+            σ=100.0,
+            retraction_method=PolarRetraction(),
+            debug=[:Iteration, :Cost, :σ, :ρ_denonimator, "\n"],
```
mateuszbaran, line 132: `ρ_denonimator` looks like a typo?

kellertuer: Maybe. I am basically lost at this moment, because running the tests on Julia 1.6 locally on my Mac OS the tests pass, running them on CI '1.6 they fail. 
But you are right, the variable should be called nominator actually – still this is just a wrong naming of a variable, but not the reason for this to fail. Locally on 1.6 (on 1.9 anyways) tests pass, just on CI they don't.
And yes I spent about 3 hours with that yesterday, so I might be a bit annoyed by now.

mateuszbaran: It says `ρ_denoNIMator`, not `ρ_denoMINator` :wink: . BTW, the opposite of denominator is numerator. 

Unfortunately I can't help much regarding that error on CI, I know almost nothing about ARC.

kellertuer: Hehe, nice, even better. But that typo we then have for about a few months. Sure I can fix that once I fix the bug.

It seems a tolerance issue for some approximate hessians on exactly Mac OS 1.6 because the value that is not allowed to be negative gets to about `-1e-10`; with the new error system I could also go for just issuing a warning, maybe.

kellertuer:
@blegat since we discussed a possible tutorial about methods required for a solver... I did not yet write such a tutorial here, but since I was changing a few things in the docs anyways, I wrote a short section https://manoptjl.org/previews/PR294/solvers/gradient_descent.html#Technical-Details that we could add to every solver, mentioning, which functions are required. Would that help for the use case you described in https://github.com/JuliaManifolds/Manifolds.jl/issues/658?

kellertuer:
Everything is working now, I might have to check for a missing line in coverage and whether that is fixable, but the rest is done and good to go – the TR subsolver is fully replaceable.

blegat:
> Would that help for the use case you described in https://github.com/JuliaManifolds/Manifolds.jl/issues/658?

Yes, I think that is a very valuable addition

kellertuer:
Cool. I can add more on a next PR, when I write the tutorial then – that might also help further to read those technical details.

PR #295, opened 2023-09-27
Expose more RLM parameters and possible fixes in step solver ρk

Affie:
I'm not sure about the math here, but I'm adding 3 parts that I didn't understand here.
The reason I suspected something was wrong is I get this result:

### DIFF
--- a/src/solvers/LevenbergMarquardt.jl
+++ b/src/solvers/LevenbergMarquardt.jl
@@ -35,6 +35,10 @@ then the keyword `jacobian_tangent_basis` below is ignored
   a functor inheriting from [`StoppingCriterion`](@ref) indicating when to stop.
 * `expect_zero_residual` – (`false`) whether or not the algorithm might expect that the value of
   residual (objective) at mimimum is equal to 0.
+* `η` – parameter of the algorithm, the higher it is the more likely the algorithm will be
+  to reject new proposal points
+* `damping_term_min` – initial (and also minimal) value of the damping term
+* `β` – parameter by which the damping term is multiplied when the current new point is rejected
 * `initial_residual_values` – the initial residual vector of the cost function `f`.
 * `initial_jacobian_f` – the initial Jacobian of the cost function `f`.
 
@@ -159,6 +163,9 @@ function LevenbergMarquardt!(
                                           StopWhenStepsizeLess(1e-12),
     debug=[DebugWarnIfCostIncreases()],
     expect_zero_residual::Bool=false,
+    β::Real=5.0,
+    η::Real=0.2,
+    damping_term_min::Real=0.1,
     initial_residual_values=similar(p, get_objective(nlso).num_components),
     initial_jacobian_f=similar(
         p, get_objective(nlso).num_components, manifold_dimension(M)
@@ -173,6 +180,9 @@ function LevenbergMarquardt!(
         p,
         initial_residual_values,
         initial_jacobian_f;
+        β,
+        η,
+        damping_term_min,
         stopping_criterion=stopping_criterion,
         retraction_method=retraction_method,
         expect_zero_residual=expect_zero_residual,
@@ -254,7 +264,7 @@ function step_solver!(
     nlso = get_objective(dmp)
     basis_ox = _maybe_get_basis(M, lms.p, nlso.jacobian_tangent_basis)
     get_jacobian!(dmp, lms.jacF, lms.p, basis_ox)
-    λk = lms.damping_term * norm(lms.residual_values)
+    λk = lms.damping_term * norm(lms.residual_values)^2
 
     JJ = transpose(lms.jacF) * lms.jacF + λk * I
     # `cholesky` is technically not necessary but it's the fastest method to solve the
@@ -271,9 +281,9 @@ function step_solver!(
     get_residuals!(dmp, lms.candidate_residual_values, temp_x)
 
     ρk =
-        2 * (normFk2 - norm(lms.candidate_residual_values)^2) / (
+        (normFk2 - norm(lms.candidate_residual_values)^2) / (
             -2 * inner(M, lms.p, lms.X, lms.step_vector) - norm(lms.jacF * sk)^2 -
-            λk * norm(sk)
+            λk * norm(sk)^2
         )
     if ρk >= lms.η
         copyto!(M, lms.p, temp_x)


src/solvers/LevenbergMarquardt.jl
```diff
@@ -254,7 +264,7 @@ function step_solver!(
     nlso = get_objective(dmp)
     basis_ox = _maybe_get_basis(M, lms.p, nlso.jacobian_tangent_basis)
     get_jacobian!(dmp, lms.jacF, lms.p, basis_ox)
-    λk = lms.damping_term * norm(lms.residual_values)
+    λk = lms.damping_term * norm(lms.residual_values)^2
```
Affie, line 267: I there possibly `^2` missing here?
image

src/solvers/LevenbergMarquardt.jl
```diff
@@ -271,9 +281,9 @@ function step_solver!(
     get_residuals!(dmp, lms.candidate_residual_values, temp_x)
 
     ρk =
-        2 * (normFk2 - norm(lms.candidate_residual_values)^2) / (
+        (normFk2 - norm(lms.candidate_residual_values)^2) / (
             -2 * inner(M, lms.p, lms.X, lms.step_vector) - norm(lms.jacF * sk)^2 -
-            λk * norm(sk)
+            λk * norm(sk)^2
```
Affie, line 286: Is there possibly a `^2` missing here?
image

src/solvers/LevenbergMarquardt.jl
```diff
@@ -271,9 +281,9 @@ function step_solver!(
     get_residuals!(dmp, lms.candidate_residual_values, temp_x)
 
     ρk =
-        2 * (normFk2 - norm(lms.candidate_residual_values)^2) / (
+        (normFk2 - norm(lms.candidate_residual_values)^2) / (
```
Affie, line 284: Is this this part?
image
So should it include the `1/2`

Affie: ρk is calculated as:
image

src/solvers/LevenbergMarquardt.jl
mateuszbaran: Yes, thanks, there should be `^2`-- sorry for the mistake.

src/solvers/LevenbergMarquardt.jl
mateuszbaran: Yes, I missed that `^2` too, sorry.

src/solvers/LevenbergMarquardt.jl
mateuszbaran: You're again correct, though this shouldn't impact convergence that much.

mateuszbaran:
I think we can expose additional options in `LevenbergMarquardt!` but keep in mind that those high level functions (like `LevenbergMarquardt!`) are mostly for convenience. In my experience, when doing anything nontrivial, it's better to construct the state object and objective directly in your code and call `solve!`. That's what I was doing when using Manopt for some experiments.

BTW, did these fixes help with `damping_term` running away? IIRC you can also have such behavior if you have an incorrect Jacobian. We should probably have a function that tests that, similar to `check_gradient`.

src/solvers/LevenbergMarquardt.jl
Affie: Thanks, it was only the `^2` in ρk that seemed to make a difference in my example not converging.

mateuszbaran (approved):

Affie:
>I think we can expose additional options in LevenbergMarquardt! but keep in mind that those high level functions (like LevenbergMarquardt!) are mostly for convenience. In my experience, when doing anything nontrivial, it's better to construct the state object and objective directly in your code and call solve!. That's what I was doing when using Manopt for some experiments.

Thanks, I did not know that there was an alternative that I could use. I will look into it.

> BTW, did these fixes help with damping_term running away? IIRC you can also have such behavior if you have an incorrect Jacobian. We should probably have a function that tests that, similar to check_gradient.

Yes, it converges quickly to the correct value if I set `damping_term_min` small, with a larger damping term it takes a bit longer but still converges.

mateuszbaran:
> Thanks, I did not know that there was an alternative that I could use. I will look into it.

I have some examples in another repository, for example here is Quasi-Newton: https://github.com/JuliaManifolds/MultivariateDataAnalysis.jl/blob/main/src/solvers.jl#L3-L18 . This lets you save the state between different optimization runs (less allocations), customize more things, or skip parts of high-level functions that you don't need (but may still eat some microseconds).

> Yes, it converges quickly to the correct value if I set `damping_term_min` small, with a larger damping term it takes a bit longer but still converges.

Great!

kellertuer:
I feel like we can easily expose more parameters as keyword arguments (nearly always), since that makes it easier for the user to use (instead of generating the problem and the state and calling `solve!`), but sure, in the end that is a tradeoff, how much “keyword-boilerplate” we want to do. For now I went to some extreme case sometimes, even to some interacting keywords very now and then.

So I am fine with more keywords as long as they are documented.

src/solvers/LevenbergMarquardt.jl
```diff
@@ -35,6 +35,10 @@ then the keyword `jacobian_tangent_basis` below is ignored
   a functor inheriting from [`StoppingCriterion`](@ref) indicating when to stop.
 * `expect_zero_residual` – (`false`) whether or not the algorithm might expect that the value of
   residual (objective) at mimimum is equal to 0.
+* `η` – parameter of the algorithm, the higher it is the more likely the algorithm will be
+  to reject new proposal points
```
kellertuer, line 39: This is a bit vague, can we make this more precise?

mateuszbaran: This is the same docs we have for `LevenbergMarquardtState` actually. I think it would be enough to add a range of allowed values? For example `0 < η < 1`.

kellertuer: Then we should improve that one as well, this here is really vague and at least for me it does not help what the parameter does. But sure a range is also great.

kellertuer: If we add the bound be rephrase it to the rate at which it is rejected or so? Is it deterministic? Then we should write that it is the part of acceptance of a certain value we check?

mateuszbaran, line 39:
```suggestion
* `η` – Scaling factor for the sufficient cost decrease required to accept new proposal points. Allowed range: `0 < η < 1`.
```
How about this?

kellertuer: That's what I find less vague :) thanks.

Affie: Should we include the word `threshold` somehow.
The way I understand it is the threshold parameter for the subproblem to be judged as trustworthy.
from paper:
>Let η ∈ (0, 1) be a prefixed constant. If ρk ≥ η holds, we judge the subproblem is trustworthy

Affie: The range helps a lot.

Affie: A side question, if the range of η is (0,1) should ρk then also only be in that range?

Affie: I'll leave the wording up to you, I can go change the docstring in `LevenbergMarquardtState` as well when everyone is happy.

mateuszbaran: I think it's better to understand it in terms of cost decrease, i.e. instead of 
$\rho_k = \frac{f(x_{old}) - f(x_{new})}{denominator} \geq \eta$
we can have
$f(x_{old} )- f(x_{new}) \geq \eta * denominator$
which exposes the sufficient decrease nicely. I don't see any reason why ρk would also have to be in that range.

mateuszbaran: I've added the word "threshold", it sounds like a good idea.

Affie: I updated the docstring in LevenbergMarquardtState as well

mateuszbaran: Great, thanks :+1:

mateuszbaran (approved):

kellertuer (approved):

kellertuer:
This looks good to me, thanks for reporting, fixing and improving :)

I am currently travelling (👋 from FRA), but will merge this and the other branch later today and register a new version with this.

Affie:
I tried solving a new problem that includes a different point type now and I'm having some problems. It seems like if the `λk * I` term gets to big it just runs away. I set the initial damping term small here and it actually gets the solution to the correct answer, but if I set it larger it runs away from the start, here's a small part of the debug data:

```
# 204   Last Change: 2.458570 | f(x): 0.006255 | s:2.4585704641649606 | |grad f(p)|:2.1675394296577704 | damping_term: 0.005
# 205   Last Change: 0.543011 | f(x): 0.000014 | s:0.5430106773684948 | |grad f(p)|:0.66536082796027 | damping_term: 0.005
# 206   Last Change: 0.011161 | f(x): 0.000000 | s:0.011160858231299897 | |grad f(p)|:0.03149099874481584 | damping_term: 0.005
# 207   Last Change: 0.000001 | f(x): 0.000000 | s:8.73505394589866e-7 | |grad f(p)|:1.2874390097948677e-5 | damping_term: 0.005
# 208   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 0.025
# 209   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 0.125
# 210   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 0.625
# 211   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 3.125
# 212   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 15.625
# 213   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 78.125
# 214   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 390.625
# 215   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 1953.125
# 216   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 9765.625
# 217   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 48828.125
# 218   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 244140.625
# 219   Last Change: 0.000000 | f(x): 0.000000 | s:1.442303629426467e-14 | |grad f(p)|:6.7748367200932954e-12 | damping_term: 1.220703125e6
```
Do you have any advice? There might be something wrong on my side, but I'm not sure where to look.

I don't know if it helps, but pk gets negative and then stops changing.
```
ρk=1.7440555482973672
ρk=1.6902836331506341
ρk=1.6178435916952576
ρk=1.518899252487929
ρk=1.3836565423510514
ρk=1.2049652681201364
ρk=0.995444619183045
ρk=0.8167826989049397
ρk=0.7516999053015216
ρk=0.809156019162061
ρk=0.9462673315667482
ρk=0.9993221597671592
ρk=1.0000004541245415
ρk=0.9999999999999982
ρk=0.5577474106495428
ρk=0.870358441095261
ρk=-0.2738479513406835
ρk=-0.2738479513406835
ρk=-0.2738479513406835
ρk=-0.2738479513406835
ρk=-0.2738479513406835
```

mateuszbaran:
That looks like it basically converged as much as it could though? f(x) and gradient norm are very small. I'd suggest modifying stopping criteria (the default is `StopWhenGradientNormLess(1e-12)` but for your problem `StopWhenGradientNormLess(1e-11)` should be fine).

Affie:
Thanks for the suggestion, you might be right. The "running away" from the start might just be because I initialized to the identity element of M which is very far from the solution. If I initialize closer I don't get it.

PR #298, opened 2023-10-01
Fix typos

DanielVandH:
PR created using https://github.com/tbroadley/github-spellcheck-cli.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -48,11 +48,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* The `AdaptiveWNGrad` stepsize is now availablbe as a new stepsize functor.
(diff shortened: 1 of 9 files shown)


DanielVandH:
This could probably just as well be applied to docstrings, but I didn't include .jl files in the tool.

kellertuer:
Thanks for running this. If you like, sure also apply it to the doc strings. I will check carefully that we do not have a false-positive in there later, though :)

DanielVandH:
Turns out that I don't know how to get it to properly work for .jl files, so I will leave it there (:. 

Note that the initial check from this tool is all automatic, but I have to manually confirm one-by-one (there are many false positives initially!) so I think there shouldn't be any issues hopefully.

kellertuer:
Thanks again, I checked the typos and there are no false-positives.
If you find the time to twiddle with the check to also include the docs, that would be neat. For this PR, we can just merge this and it is then part of the next release :)

Thanks again!

PR #299, opened 2023-10-01
Fix typos

DanielVandH:
PR created using https://github.com/tbroadley/github-spellcheck-cli.


DanielVandH:
Follow up to #298

kellertuer (approved):
Very nice PR – thanks for spotting all those and carefully using a tool here!
Sometimes we might have rushed a bit too much concerning the docs – while we did document hopefully most keywords for example – we maybe were not 100% careful.

A few seems also to be a bit of copy paste continuations...

If you want, feel free to also add yourself to
https://github.com/JuliaManifolds/Manopt.jl/blob/master/.zenodo.json
as an „Other” – anyone contributing a function or (at least a few) corrections may add themselves :)

kellertuer:
Could you maybe write a forem entry (https://forem.julialang.org) or similar about the spellchecker? I would maybe want to run this on Manifolds or ManifoldsBase as well.

DanielVandH:
Sure, I'll try write something up today. I'll also send you the whitelist after so you can (hopefully) avoid many false positives.

DanielVandH:
@kellertuer: See here https://forem.julialang.org/danielvandh/efficiently-fixing-typos-in-documentation-1j88. I hope it's clear enough.

DanielVandH:
[whitelist.txt](https://github.com/JuliaManifolds/Manopt.jl/files/12778774/whitelist.txt)

and here is the whitelist.

kellertuer:
Super detailed, thanks!

PR #300, opened 2023-10-06
Fix extra allocations in get_gradient!'s jacobian

Affie:
I'm not sure of the correct way to fix this, but here is the cause of the extra allocations in my example from: https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1546#issuecomment-1750536933

With this fix it shaves off 15 seconds of a 30-second solve.

### DIFF
--- a/src/plans/nonlinear_least_squares_plan.jl
+++ b/src/plans/nonlinear_least_squares_plan.jl
@@ -99,11 +99,11 @@ function get_gradient!(
 end
 
 function get_gradient!(
-    M::AbstractManifold, 
+    M::AbstractManifold,
     X,
     nlso::NonlinearLeastSquaresObjective{InplaceEvaluation},
     p,
-    Jval = zeros(nlso.num_components, manifold_dimension(M))
+    Jval=zeros(nlso.num_components, manifold_dimension(M)),
 )
     basis_p = _maybe_get_basis(M, p, nlso.jacobian_tangent_basis)
     nlso.jacobian!!(M, Jval, p; basis_domain=basis_p)
--- a/src/plans/nonlinear_least_squares_plan.jl
+++ b/src/plans/nonlinear_least_squares_plan.jl
@@ -99,10 +99,13 @@ function get_gradient!(
 end
 
 function get_gradient!(
-    M::AbstractManifold, X, nlso::NonlinearLeastSquaresObjective{InplaceEvaluation}, p
+    M::AbstractManifold, 
+    X,
+    nlso::NonlinearLeastSquaresObjective{InplaceEvaluation},
+    p,
+    Jval = zeros(nlso.num_components, manifold_dimension(M))
 )
     basis_p = _maybe_get_basis(M, p, nlso.jacobian_tangent_basis)
-    Jval = zeros(nlso.num_components, manifold_dimension(M))
     nlso.jacobian!!(M, Jval, p; basis_domain=basis_p)
     residual_values = zeros(nlso.num_components)
     nlso.f(M, residual_values, p)
--- a/src/solvers/LevenbergMarquardt.jl
+++ b/src/solvers/LevenbergMarquardt.jl
@@ -207,7 +207,7 @@ function initialize_solver!(
     lms::LevenbergMarquardtState,
 ) where {mT<:AbstractManifold}
     get_objective(dmp).f(get_manifold(dmp), lms.residual_values, lms.p)
-    lms.X = get_gradient(dmp, lms.p)
+    get_gradient!(dmp, lms.X, lms.p, lms.jacF)
     return lms
 end
 


src/plans/nonlinear_least_squares_plan.jl
```diff
@@ -99,10 +99,13 @@ function get_gradient!(
...
 function get_gradient!(
-    M::AbstractManifold, X, nlso::NonlinearLeastSquaresObjective{InplaceEvaluation}, p
+    M::AbstractManifold, 
+    X,
+    nlso::NonlinearLeastSquaresObjective{InplaceEvaluation},
+    p,
+    Jval = zeros(nlso.num_components, manifold_dimension(M))
 )
     basis_p = _maybe_get_basis(M, p, nlso.jacobian_tangent_basis)
-    Jval = zeros(nlso.num_components, manifold_dimension(M))
```
Affie, line 105: This is where all the allocations come from, in my case, I give a sparse jacobian as input.

src/plans/nonlinear_least_squares_plan.jl
```diff
@@ -99,10 +99,13 @@ function get_gradient!(
 end
 
 function get_gradient!(
-    M::AbstractManifold, X, nlso::NonlinearLeastSquaresObjective{InplaceEvaluation}, p
+    M::AbstractManifold,
+    X,
+    nlso::NonlinearLeastSquaresObjective{InplaceEvaluation},
+    p,
+    Jval=zeros(nlso.num_components, manifold_dimension(M)),
```
kellertuer, line 106: I am a bit confused by this new and even positional (not even keyword) argument.
It is not documented (yet), I am not 100% sure what it is for and would have more expected this to be an allocated field in `nlso` anyways – and an implementation detail of only this algorithm.

Is there a reason why it has to become part of the public interface of `get_gradient` where this positional argument is usually the index for the stochastic case and hence would cause ambiguities here?

src/plans/nonlinear_least_squares_plan.jl
```diff
@@ -99,10 +99,13 @@ function get_gradient!(
...
 )
     basis_p = _maybe_get_basis(M, p, nlso.jacobian_tangent_basis)
-    Jval = zeros(nlso.num_components, manifold_dimension(M))
     nlso.jacobian!!(M, Jval, p; basis_domain=basis_p)
```
kellertuer, line 109:
```suggestion
    nlso.jacobian!!(M, nlso.jacF, p; basis_domain=basis_p)
```

This would work as well since we anyways know that we have also available here?

src/solvers/LevenbergMarquardt.jl
```diff
@@ -207,7 +207,7 @@ function initialize_solver!(
     lms::LevenbergMarquardtState,
 ) where {mT<:AbstractManifold}
     get_objective(dmp).f(get_manifold(dmp), lms.residual_values, lms.p)
-    lms.X = get_gradient(dmp, lms.p)
+    get_gradient!(dmp, lms.X, lms.p, lms.jacF)
```
kellertuer, line 210:
```suggestion
    get_gradient!(dmp, lms.X, lms.p)
```

with the change above this should also be fine (again).

src/plans/nonlinear_least_squares_plan.jl
kellertuer: Ah no, I confused also and lms. Sorry for that (it's late and I had meetings all day); still, I am not much in favour of butchering the `get_gradient`/ `get_gradient!` interface with implementation details, and if so, then only as keyword arguments and properly documented.

src/solvers/LevenbergMarquardt.jl
kellertuer: (see above) please change this after adopting my remarks above.

src/plans/nonlinear_least_squares_plan.jl
mateuszbaran: Yes, that should be safe enough.

kellertuer: No sorry I confused `nlso` (the objective) and `lms` I think `jacF` is only a field of `lms`?

mateuszbaran: Ah, yes, `nlso` is not `lms`, you're right.

src/plans/nonlinear_least_squares_plan.jl
mateuszbaran: Yes, turning it into a keyword argument is a good idea.

mateuszbaran: It's for storing Jacobian computed as a temporary.

src/plans/nonlinear_least_squares_plan.jl
kellertuer: I see two possibilities here

1) (which I do not like so much) we introduce a keyword argument `jacobian=` to the `get_gradient!`/ `get_gradient` function, which we properly document. I would still feel this overcrowds the gradient interface with technical implementation details

2) we could do a cache at the objective level instead? That is we move the `jacF` field to `nlso` that might be used if possible (inplace and such)?

3) We could also in general do this in the cached scenario only anyways. That would be the Cleanest way I think, it would neither overcrowd the interface nor introduce cache fields to this specific objective.

src/plans/nonlinear_least_squares_plan.jl
kellertuer: I do see the idea here and what you try to avoid, sure. I am mainly trying to introduce this properly into the structure of Manopt without introducing too much of a mix between a nice interface and the (user hidden) implementation technical foo.

src/plans/nonlinear_least_squares_plan.jl
mateuszbaran: I don't think we have figured out caching Jacobians yet and storing `jacF` in objective just for this purpose wouldn't be ideal (initialization).

mateuszbaran: Anyway, Manopt would be much easier to optimize if we had the state more widely available in different functions.

kellertuer: Currently the design is as follows:

1) There is a problem `mp`, which contains the manifold and the objective, so anything _static_ to an optimization problem (± cache and count decorators for the objective)

2) There is a state `ms` that stores all dynamic data necessary for the algorithm

I would prefer to keep those separate, since that way one can easily solve one problem with different solvers and I feel your proposal would mix that quite a bit.

One thing one could maybe do, is to have the accessors (like `get_gradient!`) already on a very abstract level where they act on `mp` and `ms`, but that would be a quite major redesign.

kellertuer: So for example for this https://github.com/JuliaManifolds/Manopt.jl/blob/48f5ab0a1de4593a774a3c33e72b7eb17c22bd05/src/plans/gradient_plan.jl#L144 instead of working in place of `X` it could also work in place of a state. But if we have both, that might be a bit boiler-platey? The state one would always call the `X` one with its corresponding field? But then we would have to do this for all these evaluation functions.

kellertuer: Ah and that idea would kill the current cache design I fear.

mateuszbaran: State could be an optional keyword argument to problem accessors. If you don't provide it, accessor just allocates as much as it needs. If you do provide it, accessor checks if it can be used for caching and either uses it or allocates. I think it sounds simple enough?

kellertuer: I am not so sure, since it would involve both allocating and inplace accessors and the interplay with cache and count,
and I think if so, a proper way would be to allow `X` to be the state instead of using a keyword. And then still it would be unclear (for a generic state) which field the gradient would write to in the state.

mateuszbaran: The interplay would be that we don't give the state if we are not sure it's OK to do so. Allowing `X` to be the state doesn't make much sense because we would still need a different argument for output of in-place `get_gradient!`.

kellertuer: Hm now you lost me with the output of in-place – but maybe this comment area is also not the best place to discuss a (in my current mind) humungous rework-and-keyword-cluttering of all problem accessors.

mateuszbaran: Count and cache can just pass the state during evaluation if they get it.

kellertuer: I am still not yet so much on track – so all accessors would get a `state=` keyword – and then? if it is the `lms` you could still not dispatch on that to access the Jacobian, since it is a keyword.

if we have `get_gradient!(mp, lms, p)` that by default calls `get_gradient!(mp, lms.X, p)` one could at least dispatch on `lms`.

kellertuer: (just that for a generic state `s.X` would be horrible to write I fear?)

kellertuer:
To summarise a lengthy chat.

The idea proposed here breaks with one central design idea, that the problem (manifold, cost function , gradient,...) is stored in one struct and the state (everything the solver needs between iterates) in another.
It also abstracts accessors to the problem from implementation details.

With these ideas the current (though short) PR breaks – quite a bit.
I do not yet see a good way to resolve this, sadly, but I acknowledge, that “passing state memory to the problem/objective” might be something useful to have.

mateuszbaran:
I've submitted an alternative fix that doesn't break Manopt's design: #301 -- it should also fix this issue.

kellertuer:
Superseeded by #301

PR #303, opened 2023-10-12
Only compute Jacobian in RLM if last step success

Affie:
The estimate `lms.p` is only updated on a successful step, therefore a Jacobian is not needed if the last step failed. 
This PR skips updating the Jacobian if the last step failed.

### DIFF
--- a/src/plans/nonlinear_least_squares_plan.jl
+++ b/src/plans/nonlinear_least_squares_plan.jl
@@ -192,6 +192,7 @@ mutable struct LevenbergMarquardtState{
     damping_term_min::Tparams
     β::Tparams
     expect_zero_residual::Bool
+    last_step_successful::Bool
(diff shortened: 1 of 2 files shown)


mateuszbaran (approved):
LGTM. We could avoid recomputing of a few more things in case of a rejected step (`grad_f_c` and `lms.X`) but that should be less important.

kellertuer (approved):
I am fine with this as well.
Is this urgent? Then we could register a version with this, otherwise I would just wait for the JuMP interface to also finish?

Affie:
No, tagging is not urgent as I can work on the master branch locally and CI will still pass as it's only a performance enhancement.

kellertuer:
Ok, nice, I'll merge it to master and it will be part of the next release.

PR #315, opened 2023-11-08
Minor changes to Readme.md

odow:
x-ref https://github.com/jump-dev/JuMP.jl/pull/3568

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -39,9 +39,9 @@ a Riemannian manifold.
 
 The following packages are related to `Manopt.jl`
 
-* [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/stable/)—a library of manifolds implemented using [`ManifoldsBase.jl`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/) :octocat: [GitHub repository](https://github.com/JuliaManifolds/Manifolds.jl)
-* [`ManifoldsDiff.jl`](https://juliamanifolds.github.io/ManifoldDiff.jl/stable/)—a package to use (Euclidean) AD tools on manifolds, that also provides several differentials and gradients. :octocat: [GitHub repository](https://github.com/JuliaManifolds/ManifoldDiff.jl)
-* [`JuMP.jl`](https://jump.dev/) can be used as interface to solve an optimization problem with Manopt. See [usage examples](https://manoptjl.org/stable/extensions/). :octocat: [GitHub repository](https://github.com/jump-dev/JuMP.jl)
+* [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/stable/): a library of manifolds implemented using [`ManifoldsBase.jl`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/) :octocat: [GitHub repository](https://github.com/JuliaManifolds/Manifolds.jl)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -39,8 +39,8 @@ a Riemannian manifold.
 
 The following packages are related to `Manopt.jl`
 
-* [`Manifolds.jl`](https://juliamanifolds.github.io/Manifolds.jl/stable/) – a library of manifolds implemented using [`ManifoldsBase.jl`](https://juliamanifolds.github.io/ManifoldsBase.jl/stable/) :octocat: [GitHub repository](https://github.com/JuliaManifolds/Manifolds.jl)
```
odow, line 42: This was an en-dash, but it should be an em-dash, and typically no spaces are used surrounding an em-dash.

Readme.md
```diff
@@ -12,7 +12,7 @@ Optimization Algorithm on Riemannian Manifolds.
 For a function $f: ℳ → ℝ$  that maps from a [Riemannian manifold](https://en.wikipedia.org/wiki/Riemannian_manifold)
 ℳ to the real line, we aim to solve
 
-> Find the minimizer p on ℳ, i.e. the (or a) point where f attains its minimum.
+> Find the minimizer p on ℳ, that is, the (or a) point where f attains its minimum.
```
odow, line 15: Style guide recommends avoiding Latin abbreviations to help non-native speakers.

Readme.md
```diff
@@ -98,16 +98,16 @@ Note that all citations are in [BibLaTeX](https://ctan.org/pkg/biblatex) format.
 
 `Manopt.jl` belongs to the Manopt family:
 
-*  [manopt.org](https://www.manopt.org) – The Matlab version of Manopt, see also their :octocat: [GitHub repository](https://github.com/NicolasBoumal/manopt)
-* [pymanopt.org](https://www.pymanopt.org/) – The Python version of Manopt – providing also several AD backends, see also their :octocat: [GitHub repository](https://github.com/pymanopt/pymanopt)
+*  [manopt.org](https://www.manopt.org)—The MATLAB version of Manopt, see also their :octocat: [GitHub repository](https://github.com/NicolasBoumal/manopt)
```
odow, line 101: MATLAB is capitalized.

Readme.md
kellertuer: I am very much used to spaces in the languages I use besides English, so this really looks horrible now I think. Then I would prefer just a space instead I think.

Readme.md
kellertuer: Again, does not look nice, but yes, it is how they write their name so it might have to be like that. To me I read that always in a screaming and loud voice if it is caps.

odow: The registered mark is "MATLAB" so we don't have much choice over what looks nice.

Readme.md
odow: It seems spaces are sometimes used
https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use#:~:text=Spacing%20around%20an%20em%20dash,right%20up%20next%20to%20it.

But perhaps an em-dash is the wrong choice. What about just a comma?

kellertuer: a comma would also be ok, I think.

kellertuer:
You can ignore the changelog error failing here, that is just if we release a new version with a PR the changelog should have an entry. I will add this to another version anyways. You can still add a small (unreleased) entry with a change of the readme as a comment if you want.

kellertuer:
A : works as well, sure – do you need this registered or is it enough to have this on master/main?

odow:
It's enough to just have this on `main`. I can use a commit hash as the reference.

kellertuer:
Great. Then we can merge this already and it will be part of the next (but a bit later) release.

Issue #297, opened 2023-09-29
Manopt.jl Roadmap to 1.0


kellertuer:
I feel that by now the overall architecture of Manopt.jl seems reasonably stable. In this issue I would like to keep track of (probably breaking) changes and improvements that would need to be done before we release a 1.0. I would currently assume there is at least one, if not two breaking releases before we go for 1.0.
This is also not meant as something that should be rushed – since any breaking change has effects on the eco system.

Feel free to recommend further points as comments below to be added to this list. 

* [x] `set_parameter` was designed a bit ad-hoc and should have a more consistent set of symbols to address semantic an not syntactic elements of sub-problems and sub-states. E.g. `:p` should become `:Iterate`.
* [x] `get_parameter` should be used for accessing sub-solver states besides the result.
* [x] Integration into Optimization.jl (https://github.com/SciML/Optimization.jl/pull/712)
* [x] All functions from `src/functions` should either be moved to `ManifoldDiff` (basic functions) or `ManoptExamples` (TV, Bézier) (#331).
* [x] #381 
* [x] Consider reducing dependencies here by writing an extra package for visualisation, cf #575 .
* [x] #610 
* [x] #611
* [x] #634 

Feel free to open more issues of what you would like to see as features within Manopt.jl 1.0

oscardssmith:
Can I request some help on the Optimization.jl size with https://github.com/SciML/Optimization.jl/pull/906 to get updated to Manopt@0.5?

kellertuer:
Hi,

I can try to find some time, but for now my experience in getting help there was not the best of experiences. While the support is approximately half there, Manopt is also still not mentioned in the Readme, so for the last few months I did not follow up too much on that.

kellertuer:
All points on the list are now completed. So one could now consider to work towards a version 1.0 with final ideas.

Especially feedback would be nice on – what would be things to “reasonably break” to make Manopt
* easier to use
*  more unified in naming
* nicer in code
* ...whatever reason else

Feedback welcome!

One thing I am considering is:
1. Currently several abstract types have an `Abstract` upfront like the `AbstractManifoldProblem` or the `AbstractManifoldObjective` while other object “groups” like `Stepsize` do not. One could unify that, though not 100% sure to which of the cases, probably with `Abstract` upfront?

Feel free to add further points.

PR #331, opened 2023-11-28
Move Examples.

kellertuer:
This PR moves functions that formerly were used in examples to `ManoptExamples.jl` to reintroduce these old examples over there as well.

While formally this might be considered breaking, I would assume that these functions were actually unused since we abandoned the old examples a year ago when releasing Manopt 0.4.
Still all removed functions will be carefully documented in the changelog and the test suite cases / files will also move over.

This resolves one of the points in #297.

This PR is accompanied by the corresponding PRs in `ManifoldDiff`(https://github.com/JuliaManifolds/ManifoldDiff.jl/pull/37) and `ManoptExamples.jl` (https://github.com/JuliaManifolds/ManoptExamples.jl/pull/12).


kellertuer:
The reduction in global code coverage is unavoidable, since we reduced the code base but did not cover any of the 19 not yet covered lines. Besides this remark all is working and ready-to-merge.

Issue #333, opened 2023-12-15
Improving performance of L-BFGS


mateuszbaran:
I'm currently working on some benchmarks aimed at identifying Manopt.jl's performance bottlenecks.

Bottlenecks for quick-to-evaluate functions:

1. https://github.com/JuliaManifolds/Manopt.jl/blob/3d38cc559bf29e030ab5b24877c06131e679622a/src/plans/stepsize.jl#L795 should be in-place
2. `get_gradient` should be turned into in-place: https://github.com/JuliaManifolds/Manopt.jl/blob/3d38cc559bf29e030ab5b24877c06131e679622a/src/plans/stepsize.jl#L812
3. [resolved in #334] ~~https://github.com/JuliaManifolds/Manopt.jl/blob/3d38cc559bf29e030ab5b24877c06131e679622a/src/plans/quasi_newton_plan.jl#L495 is slow~~
4. https://github.com/JuliaManifolds/Manopt.jl/blob/3d38cc559bf29e030ab5b24877c06131e679622a/src/plans/quasi_newton_plan.jl#L474 should be in-place
5. https://github.com/JuliaManifolds/Manopt.jl/blob/3d38cc559bf29e030ab5b24877c06131e679622a/src/plans/stepsize.jl#L828 should be in-place
6. https://github.com/JuliaManifolds/Manopt.jl/blob/3d38cc559bf29e030ab5b24877c06131e679622a/src/plans/stepsize.jl#L368 should be in-place
7. https://github.com/JuliaManifolds/Manopt.jl/blob/3d38cc559bf29e030ab5b24877c06131e679622a/src/solvers/quasi_Newton.jl#L339 could we somehow multiply `η` by `α` in-place?

Maybe more but these look like the main ones

Bottlenecks for more expensive functions:

- TBD

The code:

```julia

using Optim, Manopt
using Manifolds

f(x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2
f_manopt(::Euclidean, x) = (1.0 - x[1])^2 + 100.0 * (x[2] - x[1]^2)^2

optimize(f, [0.0, 0.0], Optim.NelderMead())

function g!(storage, x)
    storage[1] = -2.0 * (1.0 - x[1]) - 400.0 * (x[2] - x[1]^2) * x[1]
    storage[2] = 200.0 * (x[2] - x[1]^2)
end

optimize(f, g!, [0.0, 0.0], LBFGS())

[...]

PR #334, opened 2023-12-15
Faster `safe_indices`

mateuszbaran:
Resolves (3) from #333 .


kellertuer (approved):
Looks good to me. We could mark this as resolved in the corresponding issue (maybe with a symbol?) – and we should collect a few more before registering a new version for sure.

PR #339, opened 2023-12-28
New benchmark

mateuszbaran:
To compare performance of Manopt.jl and Optim.jl. TODO:

- [x] Why does Manopt perform more iteration? I'm trying to match parameters of Optim.jl and Manopt.jl as much as possible. Line search is different but for now `StrongWolfe` errors for Manopt.
- [x] Are we fine with `StopWhenGradientInfNormLess`?


kellertuer:
Instead of `StopWhenGradientInfNormLess` one could also consider a `norm=` keyword for `StopWhenGradientNormLess` – but that should still default to the 2-norm, since

* basis would introduce more complexity for users with individual manifolds
*  we would change the default which I would consider breaking.

I am not sure why strong Wolfe errors, that would be interesting to know.

We could also remove the two existing benchmarks since I never ran them – in the long run we could set up a benchmark CI, but I have not yet fully understood how they work and which packages one would use for that.

ext/ManoptLineSearchesExt.jl
```diff
@@ -59,6 +59,7 @@ function (cs::Manopt.LineSearchesStepsize)(
         return α
     catch ex
         if isa(ex, LineSearches.LineSearchException)
+            println(ex)
```
kellertuer, line 62: I think I would prefer a re-throw of the error here; and maybe we should introduce such errors in our line searches as well.

src/solvers/quasi_Newton.jl
```diff
@@ -617,8 +617,8 @@ function update_hessian!(
     end
 
     # add newest
-    push!(d.memory_s, st.sk)
-    push!(d.memory_y, st.yk)
+    push!(d.memory_s, copy(M, st.sk))
```
mateuszbaran, line 620: This was causing a very nasty bug in L-BFGS. No copy = old values are overwritten in memory. So we need one value per memory entry in a buffer. @kellertuer could you rework this so that these memory buffer is preallocated instead of making a copy here? It fixes the bug but is slow.

ext/ManoptLineSearchesExt.jl
mateuszbaran: OK, then the easiest solution is to just not catch the error. We could have that in our line searches too.

src/solvers/quasi_Newton.jl
mateuszbaran: This fix doesn't completely remove the discrepancy in the number of iterations (for memory size = 2) but it makes the gap much smaller (in my test example I get 118 iterations in Manopt vs 90 in Optim). So there might be another issue but I'd prefer to look for it after the buffer allocation is fixed.

mateuszbaran:
The test failure seems unrelated to my changes (it's the ALM solver).

src/solvers/quasi_Newton.jl
kellertuer: I do not understand your first comment sadly, so _previously_ before we started optimising this, there was an allocation for sk and yk before so there was no need to copy here. That is what I find _so_ exhausting in optimising code with allocations that removing allocations has side effects (that I honestly do not have all in mind 5 years after writing some code).

The reason this is currently not preallocated is, that this way it grows for the first until it is filled, so if you never need the full memorz´y, it is also not filled. So I would prefer to keep it that way.

ext/ManoptLineSearchesExt.jl
kellertuer: Yes.

kellertuer:
Well, ALM has a subsolver that often is L-BFGS. So if you ALM breaks, Quasi Newton changed (maybe an error, but sometimes ALM is also a bit unstable). But it _is_ an effect of changing L-BFGS.

src/solvers/quasi_Newton.jl
mateuszbaran: I see, it's really unfortunate that it was so easy to miss that this allocation is necessary.

It doesn't have to be preallocated all at once though, the first `m` elements for memory could be copied like here and next ones could be overwritten using `copyto!`.

kellertuer: That sounds reasonable. I had thought we did it that way – but maybe we do not since we have the `deleteat!`? I am also not sure how to realise such a reuse (directly now, out of my head).

mateuszbaran: I've figured out how to do the reuse in a simple way :slightly_smiling_face: .

kellertuer: Ah, I See, because it is already a circular buffer, the push would throw away for the m+1st case. Nice solution!

mateuszbaran:
OK, then I will pay attention to it.

mateuszbaran:
Ref. https://github.com/JuliaNLSolvers/LineSearches.jl/issues/173 .

src/plans/hager_zhang.jl
kellertuer, line 1: Is this file meant to stay here or just an interims thing?

It looks too different in code style compared to the rest of the packages, is not really well-documented, sp that I would prefer if you did fix HZ in this file, maybe do a PR to linesearches.jl?

mateuszbaran: It's meant to stay in some package, and it would be easiest if it would be Manopt.jl. There is sadly no chance of my fix being reviewed in LineSearches.jl, and I would prefer to use fixed HZ linesearch for benchmarks as it appears to be much better than what Manopt.jl currently offers.

kellertuer: Why is there sadly no chance?

Hm, on the one hand I really do not like other peoples code to live in this package, neither in style nor with the idea it _could_ really just be in LineSeaches.jl. But what I do find worse is that I do not see any way to understand that code in its current form, since it is not documented. Sure the algorithm is quite involved, I understand that. But if we have a user coming by because there are further bugs in, what do we do? Ask Patrick? Fix in in both copies?

mateuszbaran: There was no response to my issue, neither on Github nor on Slack. I've never seen a change from "no response to issue" to "timely review of a PR". If you don't like it here, I'll just make `FixedHagerZhangLinesearch.jl`. If someone reports a bug, it will just be my problem :slightly_smiling_face: .

kellertuer: We could first try to open a PR with them and if that does not work out I would actually prefer your proposal of a small LineSearch add-on – also because that might be useful to other – non-manifold – people as well.

mateuszbaran: I will open a PR as soon as a maintainer writes that they would review it. Feel free to ping Patrick.

Regarding making this a LineSearches.jl add-on, actually the Manopt.jl interface allows me to optimize allocation of temporaries which is not possible in the LineSearches.jl interface.

kellertuer: Ah, that is also nice to know, that Manopt.jl allows for a bit more flexibility here :) I would still prefer the add-on package if that is possible.

mateuszbaran: OK, but I think it would be better as a Manopt.jl add-on than LineSearches.jl add-on.

kellertuer: The alternative I see is that we properly document all the new functions involved and check a bit that the general philosophy of the code maybe a bit closer in style to the Manopt.jl one – for seeing that better I would (however) fist have to understand it better. Then we could also keep it here , but I fear that such an improvement might also take a while.

mateuszbaran: We can move it later to Manopt, and deprecate the add-on.

kellertuer: Sure, that would also be a good solution.

mateuszbaran:
I've created the package for HZ: https://github.com/mateuszbaran/ImprovedHagerZhangLinesearch.jl .

kellertuer:
Thanks for the discussion and this nice solution then :)

kellertuer:
I also think it is not the optimal solution, but I also do agree, that thoroughly working through that code, making it (1) more Manopt.jl-like and (b) documenting it thoroughly might take more time indeed.

src/solvers/quasi_Newton.jl
```diff
@@ -345,6 +345,11 @@ function step_solver!(mp::AbstractManoptProblem, qns::QuasiNewtonState, iter)
     M = get_manifold(mp)
     get_gradient!(mp, qns.X, qns.p)
     qns.direction_update(qns.η, mp, qns)
+    if real(inner(M, qns.p, qns.η, qns.X)) > 0
```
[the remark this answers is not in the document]
mateuszbaran: Much more than a rounding error. You can try running the `test_case_manopt()` function from the PR to observe the issue. The direction gets wrong by the third iteration. I'd prefer a better fix too but it's fairly difficult to figure it out.

It never appears in my tests on flat Euclidean where Manopt's iterations are very close to Optim.jl's iterations. So, I don't know, but maybe it has something to do with vector transports. Both projection transport and parallel transport give this error though.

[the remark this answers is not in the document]
mateuszbaran: With this fix, both Manopt and Optim converge well on Rosenbrock with spherical constraints. Sometimes Manopt is faster, and sometimes Optim. For Euclidean Rosenbrock, Manopt is consistently a bit faster.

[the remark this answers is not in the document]
kellertuer: Well, this “fix” is till of the kind “if it is wrong reset, restart” instead of finding out what is wrong. For projection transport I think this _can_ happen, since that only approximates, it should not happen for parallel transport. But with now 3 master students, I will not have much time beyond my own small PRs I have planned (or want to finish). 


But I would still prefer to not have such duck-typing fixes in the code. If it is wrong, let's better find the error than just saying “Error appeared? Restart!”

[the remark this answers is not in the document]
kellertuer: Or more precisely, you are fixing a symptom, but not its origin. Sure, fixing the origin takes more time, but I see no reason to rush here?
(If your car breaks a tire, an actual working new tire would also be better than a large spool of duct tape wrapped around the old tire and saying: Look! Nearly a tire again – and on even grounds, still reasonably fast!)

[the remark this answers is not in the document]
mateuszbaran: Is the search direction in Riemannian L-BFGS actually guaranteed to be a descent direction from theory?

> Sure, fixing the origin takes more time, but I see no reason to rush here?

People are actually driving a car with a flat tire right now by using Manopt, so duct-taping it seems better than not doing anything?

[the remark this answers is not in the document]
kellertuer: For the first, I will check but I am quite sure (among others because the Matlab version has, to the best of my knowledge not such a duct tape fix).

You are right duct taping is maybe better than not fixing it, but we should also be careful not to forget about it then, which I see as a danger here if the leave it as you propose here. I would then _at least_ add a warning.

But I can try to find time to take a look at this as well, just not this week, since it is already now super-packed.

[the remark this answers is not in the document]
mateuszbaran: Sure, a warning and for example an open issue to remind us to look into it later would be fine. Anyway, I'm not sure we can 100% guarantee that there won't be an incorrect direction there, and having that check even after fixing the underlying issue might still be a good precaution. Whether it should restart or throw an error, I'm not sure. Restarting means getting an answer but more slowly, throwing means no answer at all.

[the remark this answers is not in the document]
kellertuer: I do not want to throw an error, that is for sure ;)

If you want to keep the check, then please with a warning that states what occurred and what is done to fix it.

And we should also open an issue with the smallest example you can generate :)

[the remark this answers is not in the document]
mateuszbaran: Hm, it seems to only happen with my modification of HZ. I will keep an eye for a more stable reproducer.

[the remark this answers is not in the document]
kellertuer: Oh, that sounds super interesting, but would maybe mean, you have a bug then? If it is not for the reason of qN itself, I would again prefer not to have this warning here;
but one could indeed come up with a “VerificationState” (similar to the same manifold) where in each iteration one allows for a certain test(s) to be performed.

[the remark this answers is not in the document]
mateuszbaran: I'm trying to figure it out but a bug in linesearch doesn't sound like a plausible explanation to me.

`VerificationState` sounds like something that would be nightmarishly difficult to implement. I believe such optional sanity checks could be useful in the future, even if they are off by default.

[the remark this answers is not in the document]
mateuszbaran: Ha, I've found an independent reproducer: #346 .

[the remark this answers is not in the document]
kellertuer: Oh I did not want to imply _all_ linesearch is wrong but maybe your improvements had a slight flaw in one specific case. But great that you found an example and good that it even works with PT.
Then we could indeed keep the warning here and keep that (and the duct tape fix) until we fix the bug.

mateuszbaran:
I'm trying to slowly wrap this up. How would you organize the hyperparameter tuner? It's a mix of somewhat generic code and example-specific code. `ObjectiveData` is technically generic but I'd imagine most people using it would have to tweak it anyway -- there is simply too many fine details to cover every possible use case through an API. Probably ManoptExamples.jl would be the right place, at least for now?

kellertuer:
Hi,
I would first have to check what that function does, but will try to find time for that the next few days then, but yeah, ManoptExamples would probably ne a good place. Note that in there we also already have Rosenbrock.

benchmarks/benchmark_comparison.jl
```diff
@@ -0,0 +1,280 @@
...
+        $g!,
+        $x0;
+        stepsize=$(stepsize),
+        evaluation=$(InplaceEvaluation()),
+        memory_size=$mem_len,
+        stopping_criterion=$(manopt_sc),
+        debug=[],
+        $kwargs...,
+    )
+    manopt_state = quasi_Newton(
```
kellertuer, line 109: Would it maybe make sense to do this call before the benchmark (Line 97)?

benchmarks/benchmark_comparison.jl
```diff
@@ -0,0 +1,280 @@
...
+function test_case_optim()
+    N = 4
+    mem_len = 2
+    ls_hz = LineSearches.HagerZhang()
+    method_optim = LBFGS(; m=mem_len, linesearch=ls_hz, manifold=Optim.Flat())
+    options_optim = Optim.Options(; g_tol=1e-6)
+
+    x0 = zeros(N)
+    x0[1] = 0
+    optim_state = optimize(f_rosenbrock, g_rosenbrock!, x0, method_optim, options_optim)
```
kellertuer, line 278: Similar to the previous all, this could also be one line just the `optimize` call whose return value is returned?

benchmarks/manoptuna.jl
```diff
@@ -0,0 +1,215 @@
+using Manifolds
+using PythonCall
+include("benchmark_comparison.jl")
+
+# This script requires optuna to be available through PythonCall
```
kellertuer, line 5: I have no experience with optuna, could this be documented a bit what the goal of this is?

benchmarks/manoptuna.jl
```diff
@@ -0,0 +1,215 @@
...
+    report::TTreport
+    should_prune::TTshould_prune
+end
+
+"""
+    ObjectiveData
+
+
+"""
+mutable struct ObjectiveData{TObj,TGrad}
```
kellertuer, line 51: This is still undocumented and the functor below _quite_ technical. Could we maybe document that a bit more to also make it understandable for either me or also you in some future time ;) ?

benchmarks/manoptuna.jl
```diff
@@ -0,0 +1,215 @@
...
+
+"""
+    lbfgs_study(; pruning_coeff::Float64=0.95)
+
+ensure `pruning_losses` is actually somewhat realistic,
+otherwise there is too little pruning (if values here are too high)
+or too much pruning (if values here are too low)
+regenerate using `lbfgs_compute_pruning_losses`
+"""
+function lbfgs_study(; pruning_coeff::Float64=0.95)
```
kellertuer, line 157: This also looks quite technical, so.I am not so sure what it does and what its purpose in the end is?
Should this later be a tutorial?

Changelog.md
```diff
@@ -5,6 +5,12 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.53] unreleased
```
kellertuer, line 8: This number has to be adapted.

src/solvers/quasi_Newton.jl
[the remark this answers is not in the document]
kellertuer: I think we discussed this enough in Zulip, and dressed this in tutorial mode. I hope I even added it here? We could add a tutorial debug for this if you feel that helps.

kellertuer (commented):
This looks like a very great start to a thorough Benchmark.

My maybe most central remark or question is

How would we add this in a consistent way?
The `benchmark_comparison.jl` for now is a script;
should that become something we can run every now and then on a CI?
Should the results be part of the docs?
They could then be updated on a branch whenever the benchmark is run (either on CI and committed or when run manually).

then one could for example turn the current benchmark into a Quarto notebook.

Does the benchmark run several solvers _and_ several examples? This could maybe be modularized

kellertuer:
If you would prefer having scripts that can be run easier (than Quarto Notebooks) we could also look into the new [Quarto Scripts](https://quarto.org/docs/computations/render-scripts.html) which I wanted to try on something anyways.

mateuszbaran:
> This looks like a very great start to a thorough Benchmark.
> 
> My maybe most central remark or question is
> 
> How would we add this in a consistent way? The `benchmark_comparison.jl` for now is a script; should that become something we can run every now and then on a CI? Should the results be part of the docs? They could then be updated on a branch whenever the benchmark is run (either on CI and committed or when run manually).

Currently I'm leaning towards turning it into a "how to choose the right solver (and its options) for your problem" tutorial. I'm not sure how (and for what problems) run it on CI. Note that the optimization script is quite demanding computationally despite all that advanced machinery.

> Does the benchmark run several solvers _and_ several examples? This could maybe be modularized

I've experimented with several examples but I haven't decided which one to use for an example. Most likely not Rosenbrock on sphere but I will decide when we select the right format.

> If you would prefer having scripts that can be run easier (than Quarto Notebooks) we could also look into the new [Quarto Scripts](https://quarto.org/docs/computations/render-scripts.html) which I wanted to try on something anyways.

I really liked Quarto notebooks until updating Julia broke my settings and I could not get it back to work for a couple of hours (it still doesn't work, I just gave up; it makes the Jupyter notebook but the Julia kernel refuses to run). That might be the main problem with turning it into a tutorial.

benchmarks/benchmark_comparison.jl
```diff
@@ -0,0 +1,280 @@
...
+
+function g_rosenbrock!(storage, x)
+    for i in 1:2:length(x)
+        storage[i] = -2.0 * (1.0 - x[i]) - 400.0 * (x[i + 1] - x[i]^2) * x[i]
+        storage[i + 1] = 200.0 * (x[i + 1] - x[i]^2)
+    end
+    return storage
+end
+
+optimize(f_rosenbrock, g_rosenbrock!, [0.0, 0.0], LBFGS())
```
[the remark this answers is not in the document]
mateuszbaran: Yes; this will be removed or commented when preparing final version.

benchmarks/benchmark_comparison.jl
mateuszbaran: It shouldn't really matter, benchmarking macro already takes care of compilation time.

benchmarks/manoptuna.jl
mateuszbaran: I will document it, I'm just not sure what style of describing to use here. I will add a tutorial-style commentary.

src/solvers/quasi_Newton.jl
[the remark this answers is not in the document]
mateuszbaran: Hm, good that you remind me. I thought we already had this check merged in Manopt.jl? This is 100% not a tutorial-only issue because some combinations of line search/manifold/vector transport will lead to non-descent directions here which we need to catch (or the user gets a very hard to debug problem). Those combinations are still useful because they sometimes converge faster than "always descent direction" alternatives.

kellertuer:
> I really liked Quarto notebooks until updating Julia broke my settings and I could not get it back to work for a couple of hours (it still doesn't work, I just gave up; it makes the Jupyter notebook but the Julia kernel refuses to run). That might be the main problem with turning it into a tutorial.

Remember to recompile `IJulia`. That is one of the main reasons I did so much `Pkg.activate()`... stuff in the documentation. If it happens that a new Julia version comes along, at least recompiling `IJulia` is crucial.

I would also be fine with these being benchmarks and a bit less tutorial focussed.

mateuszbaran:
> Remember to recompile `IJulia`. That is one of the main reasons I did so much `Pkg.activate()`... stuff in the documentation. If it happens that a new Julia version comes along, at least recompiling `IJulia` is crucial.

I did recompile `IJulia`. This script actually needs to run with the `Conda.jl` Python but I'm not sure how to run the notebook in its Jupyter. Or I did something wrong trying.

mateuszbaran:
I just tried again, activating the Conda.jl environment. `quarto render` fails throwing "No module named yaml" despite me being able to import it when I run `python` REPL. Directly running the browser notebook complains about `IJulia` not being installed despite me just doing `build IJulia` a moment ago.

mateuszbaran:
I've moved the interesting parts of this PR to separate PRs so this one can be closed I think.

mateuszbaran:
BTW, I tried to address some of your comments in the version submitted to ManoptExamples.

PR #341, opened 2024-01-01
Fixes from benchmark

mateuszbaran:
Since the benchmark in #339 is not ready yet, I've extracted the bugfixes to a separate branch.


Changelog.md
```diff
@@ -5,6 +5,20 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.46] January 1, 2024
+
+### Changed
+
+* A message it printed when a line search from `LineSearches.jl` reports search failure.
```
kellertuer, line 12: This is not so accurate since now we throw that error (thanks for changing that).

mateuszbaran: Right, that was changed. Thanks for noticing!

kellertuer (approved):
I am not sure where we lose the subsolver gradient access test but the Wolfe-conidtion one is one that is definelty unrelated to the changes here. If you have a moment, taking a look would be great but otherwise this is also fine.

mateuszbaran:
I think we can leave that coverage issue for later.

PR #344, opened 2024-01-13
Differentiate more precisely bewteen swarm and iterate in PSO more stricly

kellertuer:
We noticed an inconsistency yesterday, where (probably for historic reasons) the swarm was sometimes returned when calling `get_iterate` on the `ParticleSwarmState` – probably since it made a few things easer.

This PR fixes that and introduces `:Swarm` to access the swarm in general within the (much newer than  PSO) `get_manopt_parameter` and updates its occurrences accordingly.
Furthermore `get_manopt_parameter` is now as tolerant as `set_manopt_parameter!` and returns nothing (but does not error) if a state/problem/... does not have a certain parameter.


mateuszbaran:
For a generic interface maybe `:Population` would be better as it can refer to any evolutionary algorithm. Adding CMA-ES to Manopt is on my TODO list.

kellertuer:
That sounds like a very good idea, to call that generic name `: Population` – will do that :)

hajg-ijk (approved):

PR #345, opened 2024-01-16
Introduced an entry change stopping criterion and generalise the gradient norm SC.

kellertuer:
This PR

* [x]  introduces a `StopWhenEntryChangeLess` stopping criterion
* [x]  generalises the `StopWhenGradientNormLess` to accept arbitrary norm functions. Can you check, @mateuszbaran,  whether that includes your InfNorm one? I think this way it is much nicer to introduce different Norm-possibilities.
* [x] the entry change should be used in the PSO to check the population check (instead of “misusing” the iterate one)
* [x] needs test coverage after clarification of functionality.


kellertuer:
The decrease per patch is just the 10 lines of Strong Wolfe again that depending on the randomness in the tests sometimes are not reached. So this is ready to go.

src/plans/stopping_criterion.jl
```diff
@@ -419,30 +504,48 @@ end
...
 """
-mutable struct StopWhenGradientNormLess <: StoppingCriterion
+mutable struct StopWhenGradientNormLess{F,TF} <: StoppingCriterion
+    norm::F
     threshold::Float64
     reason::String
     at_iteration::Int
-    StopWhenGradientNormLess(ε::Float64) = new(ε, "", 0)
+    function StopWhenGradientNormLess(
+        ε::TF; norm::F=(M, p, X) -> norm(M, p, X)
```
mateuszbaran, line 532:
```suggestion
        ε::TF; norm::F=norm
```
There is no point in making a new function here.

src/plans/stopping_criterion.jl
```diff
@@ -419,30 +504,48 @@ end
 
 A stopping criterion based on the current gradient norm.
 
+# Fields
+
+* `norm`  – a function `(M, p, X) -> R` that computes a norm of the gradient `X` in the tangent space at `p` on `M``
```
mateuszbaran, line 509:
```suggestion
* `norm`  – a function `(M::AbstractManifold, p, X) -> ℝ` that computes a norm of the gradient `X` in the tangent space at `p` on `M``
```

src/plans/stopping_criterion.jl
```diff
@@ -419,30 +504,48 @@ end
...
+        sc.at_iteration = 0
     end
-    if (norm(M, get_iterate(s), get_gradient(s)) < c.threshold) && (i > 0)
-        c.reason = "The algorithm reached approximately critical point after $i iterations; the gradient norm ($(norm(M,get_iterate(s),get_gradient(s)))) is less than $(c.threshold).\n"
-        c.at_iteration = i
+    if (i > 0) && (sc.norm(M, get_iterate(s), get_gradient(s)) < sc.threshold)
+        sc.reason = "The algorithm reached approximately critical point after $i iterations; the gradient norm ($(sc.norm(M,get_iterate(s),get_gradient(s)))) is less than $(sc.threshold).\n"
+        sc.at_iteration = i
         return true
     end
```
mateuszbaran, line 550:
```suggestion
    if (i > 0) 
        grad_norm = sc.norm(M, get_iterate(s), get_gradient(s))
        if grad_norm < sc.threshold)
            sc.reason = "The algorithm reached approximately critical point after $i iterations; the gradient norm ($(grad_norm)) is less than $(sc.threshold).\n"
            sc.at_iteration = i
            return true
        end
    end
```
To avoid computing the norm twice.

src/solvers/particle_swarm.jl
```diff
@@ -297,7 +297,15 @@ function particle_swarm!(
...
-                                          StopWhenChangeLess(1e-4),
+                                          StopWhenEntryChangeLess(
+        :swarm,
+        (p, st, old_swarm, swarm) -> distance(
+            PowerManifold(get_manifold(p), NestedPowerRepresentation(), length(swarm)),
+            old_swarm,
+            swarm,
+        ),
+        1e-4,
+    ),
```
mateuszbaran, line 308: This looks like a very odd criterion. Why can't you just stop when the change in optimal point is less than 1e-4?

mateuszbaran (commented):
A few minor comments.

mateuszbaran:
BTW, yes, this is enough to represent the inf norm criterion :+1:

src/plans/stopping_criterion.jl
kellertuer: well-spotted.

src/solvers/particle_swarm.jl
kellertuer: That is actually the same as before, just that before we (sometimes) handled the swarm as the iterate and had an iterate change.

The reason to not check the (now) iterates change is, that even in the first iteration, this one might not change, while the swarm moves, maybe even all other points, depending on the social and velocity parameters.
Only if they also settle to move less it is very likely to indicate a reason to stop.

kellertuer:
> BTW, yes, this is enough to represent the inf norm criterion 👍

Nice! I had hoped so – and feel this is nicer than coming up with a new criterion (which might mean coming up with new crieria for other norms later as well, and then it gets too crowded). Just took me a while to see this relatively simple solution.

src/solvers/particle_swarm.jl
mateuszbaran: Then it would be easier to calculate the norm of velocities? There would be no need for (sometimes expensive) distance calculation and remembering the old swarm.

kellertuer: The reference we provide even just has a maxiter thing, so I would have to check how we came up with that one. But the iterate change one will indicate to stop too early, even an adapted one to check for the populations best ones.

kellertuer: Velocities sounds reasonable as well; but if we change this, then I would like to check literature before again a bit.

mateuszbaran: Sure, if this just replicates the old behavior in a nicer way, we can consider changing this stopping criterion in another PR.

kellertuer: Yes, this is exactly the same behaviour as before. To provide some detail:

Previously the swarm was sometimes considered to the the iterate (changed that in #344 ), what I did not yet change there is the one place where we still to some extend considered the swarm the iterate, since the `StopWhenChangeLess` was still redefined (until today) to check the change of the swarm instead.

With the new criterion this redefinition was changed to really use the Entry Change stopping criterion.

So for that reason these lines reproduce the behaviour from before, just in a better fitting stopping criterion name and with a nicer/shorter form.

kellertuer:
HM I just saw that you had one fix to just compute some norm only once, but it seems to have broken formatting _and_ tests. Will check this only after my lecture and some meetings somewhen in the late afternoon.

mateuszbaran:
I've fixed tests and formatting.

kellertuer:
Great thanks. Is the discussion above clarified and with that this PR ready to merge?

mateuszbaran (approved):
Yes, everything is fine.

PR #353, opened 2024-01-31
Introduce text checks (vale.sh) on Docstrings and comments

kellertuer:
This PR slowly works through all doc strings to fulfil the [vale.sh](https://vale.sh) rules the documentation already follows, because since the 3.0 version vale.sh works on Julia files and recognises its doc-strings.


kellertuer:
Vale does not yet fully work, so I do not want to activate a CI on this yet. The main challenges are

* maybe I was a bit to eager declaring strings to be vale-d as well (especially when they are formats)
* `qmd` files and their rendered results are not handled correctly, code in those markdown files seems to be treated as text, which of course is not what it is and complains about `_` that should not be used in words or such.

Still. These first steps
* improve readability
* have a large scale change, so I do not want to keep this PR open longer.

Since this does not add new features, but just rewrites docs, this will not be a new version, but be on master until a next feature justifying a new version comes along.

.codecov.yml
```diff
@@ -1,4 +1,4 @@
 ignore:
   - "src/examples/.*"
   - "src/tutorials/.*"
-  - "src/helpers/exports/.*" # until we figure out how to check asy runs - exclude.
\ No newline at end of file
+  - "src/helpers/exports/.*" # exclode exports, since there are no CI tests for Asy-exports
```
mateuszbaran, line 4:
```suggestion
  - "src/helpers/exports/.*" # exclude exports, since there are no CI tests for Asy-exports
```
Maybe it's time for Vale checks of .yml comments :smile:

mateuszbaran:
> * maybe I was a bit to eager declaring strings to be vale-d as well (especially when they are formats)
> * `qmd` files and their rendered results are not handled correctly, code in those markdown files seems to be treated as text, which of course is not what it is and complains about `_` that should not be used in words or such.

Would it be possible to enable Vale on CI only for comments but not for strings and qmd files?

BTW, I most likely won't look over all these changes so merge when you think it's ready.

kellertuer:
Sure it is a lot of changes, no worries – and most of them are even mainly unifying formatting of Keyword arguments and such.

For comments and not strings – in theory yes, in practice, I can not find how to do that.
For dmd files I for now deactivated that, but locally it still `vale`s the resulting `*.md` files as well and I am not able to say “please do not check this folder, those are autogenerated anyways”. As far as I could read about it, vale is not meant to ignore folders by design or something.

Issue #346, opened 2024-01-17
Add Warning if gradient larger than injectivity radius.


mateuszbaran:
This example generates an incorrect descent direction in L-BFGS:
```julia
using Manopt, Manifolds, LineSearches

function f_rosenbrock_manopt(::AbstractManifold, x)
    result = 0.0
    for i in 1:2:length(x)
        result += (1.0 - x[i])^2 + 100.0 * (x[i + 1] - x[i]^2)^2
    end
    return result
end

function g_rosenbrock_manopt!(M::AbstractManifold, storage, x)
    for i in 1:2:length(x)
        storage[i] = -2.0 * (1.0 - x[i]) - 400.0 * (x[i + 1] - x[i]^2) * x[i]
        storage[i + 1] = 200.0 * (x[i + 1] - x[i]^2)
    end
    riemannian_gradient!(M, storage, x, storage)
    return storage
end


function test_case_manopt()
    N = 128
    mem_len = 1
    M = Manifolds.Sphere(N - 1)
    ls_hz = LineSearches.HagerZhang{Float64}(alphamax=3.0)

    x0 = zeros(N)
    x0[1] = 1
    manopt_sc = StopWhenGradientNormLess(1e-6) | StopAfterIteration(1000)

    return quasi_Newton(
        M,
        f_rosenbrock_manopt,
        g_rosenbrock_manopt!,
        x0;
        stepsize=Manopt.LineSearchesStepsize(ls_hz),
        evaluation=InplaceEvaluation(),
        vector_transport_method=ParallelTransport(),
        return_state=true,
        memory_size=mem_len,
        stopping_criterion=manopt_sc,
    )
end

test_case_manopt()
```

EDIT:

Second example; longest gradient is about 0.12 long:

```julia
using Revise
using Manopt, Manifolds, LineSearches
using Random, ManoptExamples, LinearAlgebra

Random.seed!(30)

N = 1024

A = Symmetric(2 * randn(N, N) / N)
f_manopt = ManoptExamples.RayleighQuotientCost(A)
g_manopt! = ManoptExamples.RayleighQuotientGrad!!(A)

function test_case_manopt()
    mem_len = 2
    M = Manifolds.Sphere(N - 1)
    ls_hz = LineSearches.HagerZhang{Float64}(alphamax=3.0)

    x0 = zeros(N)
    x0[1] = 1
    manopt_sc = StopWhenGradientNormLess(1e-6) | StopAfterIteration(1000)

[...]

PR #354, opened 2024-01-31
Add a persistent layer of setting options and introduce a Tutorial mode

kellertuer:
This adresses a point raised in #346 and is for now realised in the `quasi_Newton` solver.

* [x] check which other solvers benefit from the “Tutorial” and the new warning
* [x] check test coverage
* [x] document the new mode
* [x] illustrate / mention the new mode in the intro tutorial.


kellertuer:
So the tutorial already has a section about this, I just have to figure out how to convince Quarto to display warnings :)

https://manoptjl.org/previews/PR354/tutorials/Optimize/#Using-the-“Tutorial”-mode

src/plans/debug.jl
```diff
@@ -1028,6 +1082,7 @@ Note that the Shortcut symbols should all start with a capital letter.
 
 * `:Cost` creates a [`DebugCost`](@ref)
 * `:Change` creates a [`DebugChange`](@ref)
+* `:Gradient` creattes a [`DebugGradient`](@ref)
```
mateuszbaran, line 1085:
```suggestion
* `:Gradient` creates a [`DebugGradient`](@ref)
```

src/plans/plan.jl
```diff
@@ -25,6 +25,8 @@ end
 """
     get_manopt_parameter(f, element::Symbol, args...)
 
+Access arbitrary parameters from `f` adressedby a symbol `element`.
```
mateuszbaran, line 28:
```suggestion
Access arbitrary parameters from `f` addressed by a symbol `element`.
```

src/plans/plan.jl
```diff
@@ -25,6 +25,8 @@ end
...
+
 For any `f` and a `Symbol` `e` we dispatch on its value so by default, to
 get some element from `f` potentially further qulalified by `args...`.
```
mateuszbaran, line 31:
```suggestion
get some element from `f` potentially further qualified by `args...`.
```

src/plans/plan.jl
```diff
@@ -35,6 +37,57 @@ function get_manopt_parameter(f, e::Symbol, args...)
...
+
+The parameters are queried from the global settings using [`Preferences.jl`](https://github.com/JuliaPackaging/Preferences.jl),
+so they are persistent within your activated Environment.
+
+# Currenlty used settings
+
+`:Mode`
+the mode can be set to `"Tutorial"` to get several hints especially in scenarios, where
+the optimisation on manifolds is different from the usual “experience” in
+(classicall, Euclidean) optimization.
```
mateuszbaran, line 56:
```suggestion
(classical, Euclidean) optimization.
```

src/plans/plan.jl
```diff
@@ -35,6 +37,57 @@ function get_manopt_parameter(f, e::Symbol, args...)
...
+
+Access global [`Manopt`](@ref) parametersadressed by a symbol `element`.
+We first dispatch on the value of `element`.
+
+if the value is not set, `default` is returned
+
+The parameters are queried from the global settings using [`Preferences.jl`](https://github.com/JuliaPackaging/Preferences.jl),
+so they are persistent within your activated Environment.
+
+# Currenlty used settings
```
mateuszbaran, line 51:
```suggestion
# Currently used settings
```

src/plans/plan.jl
```diff
@@ -35,6 +37,57 @@ function get_manopt_parameter(f, e::Symbol, args...)
...
+    e::Symbol, args...; default=get_manopt_parameter(Val(e), Val(:default))
+)
+    return @load_preference("$(e)", default)
+end
+# Handle empty defaults
+get_manopt_parameter(::Symbol, ::Val{:default}) = nothing
+get_manopt_parameter(::Val{:Mode}, v::Val{:default}) = ""
+
+"""
+    set_manopt_parameter!(element::Symbol, value::Union{String,Bool})
```
mateuszbaran, line 69:
```suggestion
    set_manopt_parameter!(element::Symbol, value::Union{String,Bool,<:Number})
```

src/plans/plan.jl
```diff
@@ -35,6 +37,57 @@ function get_manopt_parameter(f, e::Symbol, args...)
...
+
+Set global [`Manopt`](@ref) parameters adressed by a symbol `element`.
+We first dispatch on the value of `element`.
+
+The parameters are stored to the global settings using [`Preferences.jl`](https://github.com/JuliaPackaging/Preferences.jl).
+
+Passing a `value` of `""` deletes an entry from the preferences.
```
mateuszbaran, line 76:
```suggestion
Passing a `value` of `""` deletes the corresponding entry from the preferences.
```

src/plans/plan.jl
```diff
@@ -35,6 +37,57 @@ function get_manopt_parameter(f, e::Symbol, args...)
...
+Whenever the `LocalPreferences.toml` is modified, this is also `@info`rmed about.
+"""
+function set_manopt_parameter!(e::Symbol, value::Union{String,Bool,<:Number})
+    if length(value) == 0
+        @delete_preferences!("$(e)")
+        v = get_manopt_parameter(e, Val(:default))
+        default = isnothing(v) ? "" : (length(v) == 0 ? "" : " ($(get_manopt_parameter))")
```
mateuszbaran, line 83: Is this really supposed to just print `get_manopt_parameter` without any arguments?

src/plans/plan.jl
```diff
@@ -35,6 +37,57 @@ function get_manopt_parameter(f, e::Symbol, args...)
...
+We first dispatch on the value of `element`.
+
+The parameters are stored to the global settings using [`Preferences.jl`](https://github.com/JuliaPackaging/Preferences.jl).
+
+Passing a `value` of `""` deletes an entry from the preferences.
+Whenever the `LocalPreferences.toml` is modified, this is also `@info`rmed about.
+"""
+function set_manopt_parameter!(e::Symbol, value::Union{String,Bool,<:Number})
+    if length(value) == 0
+        @delete_preferences!("$(e)")
```
mateuszbaran, line 81:
```suggestion
        @delete_preferences!(string(e))
```
This is, I think, a more standard way of converting symbols to strings.

src/plans/stopping_criterion.jl
```diff
@@ -685,7 +634,7 @@ A functor for an stopping criterion, where the algorithm if stopped when a varia
 # Fields
 * `value` – stores the variable which has to fall under a threshold for the algorithm to stop
 * `minValue` – stores the threshold where, if the value is smaller or equal to this threshold, the algorithm stops
-* `reason` – stores a reason of stopping if the stopping criterion has one be
+* `reason` – stores a reason of stopping if the stopping criteri on has one be
```
mateuszbaran, line 637: ?

src/solvers/gradient_descent.jl
```diff
@@ -240,7 +240,15 @@ function gradient_descent!(
     ),
     stopping_criterion::StoppingCriterion=StopAfterIteration(200) |
                                           StopWhenGradientNormLess(1e-8),
-    debug=stepsize isa ConstantStepsize ? [DebugWarnIfCostIncreases()] : [],
+    debug=if (get_manopt_parameter(:Mode) == "Tutorial")
```
mateuszbaran, line 243: This is so much logic that I would suggest making a separate function for computing this default.

mateuszbaran: BTW, I'd also suggest making a helper `is_tutorial_mode` function.

tutorials/Optimize.qmd
```diff
@@ -156,9 +155,62 @@ Then we reach approximately the same point as in the previous run, but in far le
...
+grad_f2(M, p) = sum(grad_distance.(Ref(M), data, Ref(p)));
+```
+
+And we consider the gradient at the starting point in norm
+
+```{julia}
+norm(M, data[1], grad_f2(M, data[1]))
+```
+
+On the sphere, when we follow a geodesic, we “return” to the start point after length $2π$. If we “lang” short before the starting point due to a gradient of length just shy of $2π$, line search would not make much sense probably.
```
mateuszbaran, line 184: What do you mean by "lang"?

mateuszbaran (commented):
The suggested usage of Preferences.jl is that you restart your Julia session to take any preference changes into account. This ensures that all preference checks can be done very fast. Are you sure you want dynamic checks here?

src/plans/stopping_criterion.jl
kellertuer: Oh, not sure how that happened ;)

tutorials/Optimize.qmd
kellertuer: lang -> land

src/solvers/gradient_descent.jl
kellertuer: Yes, I am not yet sure we might have more modes, but if it basically remains this one, then that function sounds fair.

src/plans/plan.jl
kellertuer: I wanted to avoid to have printed `"nothing"` in the string. That is what this avoids.

kellertuer:
> The suggested usage of Preferences.jl is that you restart your Julia session to take any preference changes into account. This ensures that all preference checks can be done very fast. Are you sure you want dynamic checks here?

My honest answer is: I would prefer to be able to set this dynamically – otherwise I for example have no clue how to test that? How do I _restart Julia_ in testing to test the other mode?
For now this is persistent after starts, yes. But dynamic.

Besides the testing, I would also have not much clue how to implement what you refer to? So one would set the preference but that is never accessed but something different, because only on start (load of Manopt) the local preferences of preferences.jl are read? To me (but maybe I am just not that clever here) that reads like probably 100 more lines of code to split that into two layers, one that we load and one that we have during runtime?

src/plans/plan.jl
kellertuer: ah and if the value is length 0, eg for a string, this also avoids () and does not add brackets, like the last case does.

mateuszbaran: I see. Though maybe it would be better to print strings in `"`, so that empty string is `""`?

kellertuer: Hm, maybe, maybe we are also just overcomplicating things. In the very most cases this indeed will only be strings and `resetting to default ""` also does not read so nice.

mateuszbaran:
I thought about it again and you are most likely right to keep preferences dynamic. Let's not over-complicate it.

src/solvers/gradient_descent.jl
kellertuer: Will add the easy access method tomorrow. It will make a few places more readable.

docs/src/plans/index.md
```diff
@@ -47,4 +47,10 @@ The column “generic” refers to a short hand that might be used for readabil
 Since the iterate is often stored in the states fields `s.p` one _could_ access the iterate
 often also with `:p` and similarly the gradient with `:X`.
 This is discouraged for both readability as well as to star more generic, and it is recommended
-to use `:Iterate` and `:Gradient` instead in generic settings.
\ No newline at end of file
+to use `:Iterate` and `:Gradient` instead in generic settings.
+
+YOu can further activate a “Tutorial” mode by `set_manopt_parameter!(:Mode, "Tutorial"). Internally, the following convenience function is available.
```
mateuszbaran, line 52:
```suggestion
You can further activate a “Tutorial” mode by `set_manopt_parameter!(:Mode, "Tutorial")`. Internally, the following convenience function is available.
```

src/plans/debug.jl
```diff
@@ -946,7 +948,59 @@ function (d::DebugWarnIfFieldNotFinite)(
...
+end
+function (d::DebugWarnIfGradientNormTooLarge)(
+    mp::AbstractManoptProblem, st::AbstractManoptSolverState, i::Int
+)
+    if d.status !== :No
+        M = get_manifold(mp)
+        p = get_iterate(st)
+        X = get_gradient(st)
+        Xn = norm(M, p, X)
+        p_inj = d.factor * injectivity_radius(M, p)
```
mateuszbaran, line 989: Given that tangent bundle has 0 injectivity radius, I'd suggest using `max_stepsize` here instead of `injectivity_radius`.

tutorials/Optimize.qmd
```diff
@@ -156,9 +155,62 @@ Then we reach approximately the same point as in the previous run, but in far le
...
+grad_f2(M, p) = sum(grad_distance.(Ref(M), data, Ref(p)));
+```
+
+And we consider the gradient at the starting point in norm
+
+```{julia}
+norm(M, data[1], grad_f2(M, data[1]))
+```
+
+On the sphere, when we follow a geodesic, we “return” to the start point after length $2π$. If we “land” short before the starting point due to a gradient of length just shy of $2π$, line search would not make much sense probably.
```
mateuszbaran, line 184: Line search will always return something but it won't take into account that gradient is just a local approximation of the objective.

I think it would be better to rephrase it so that you don't have to use the word "probably".

src/plans/plan.jl
```diff
@@ -35,6 +37,64 @@ function get_manopt_parameter(f, e::Symbol, args...)
 end
 get_manopt_parameter(f, args...) = nothing
 
+"""
+    get_manopt_parameter(element::Symbol; default=nothing)
+
+Access global [`Manopt`](@ref) parametersadressed by a symbol `element`.
+We first dispatch on the value of `element`.
+
+if the value is not set, `default` is returned
```
mateuszbaran, line 46:
```suggestion
If the value is not set, `default` is returned.
```

mateuszbaran (commented):
Just a few last minor things.

docs/src/plans/index.md
kellertuer: Good that I am starting to use vale, that also reports on these. Somehow my VS Code Spellchecker does not want to work as I think it should.

src/plans/debug.jl
kellertuer: I think I do not understand what you mean here – I think the tangent bundle has even infectivity of infinite, since in every tangent space we have a vector space?

Besides that, you are right, maybe max_stepsize fits a bit better.

tutorials/Optimize.qmd
kellertuer: You are right, it will always return something, but in the situation described, the step size might actually just be 0, since for such a long vector -grad is basically a very scaled down +grad .

I will try to rephrase this to avoid the probably though.

src/plans/debug.jl
kellertuer: Changed it to `max_stepsize`.

mateuszbaran: Tangent *space* has infinite injectivity radius but tangent *bundle* (with the Sasaki metric) has injectivity radius 0 if the underlying manifold is not flat: https://mathoverflow.net/a/94348 .

mateuszbaran (approved):

src/plans/debug.jl
kellertuer: Sasaki Metric was the thing I missed here.

kellertuer:
GitHub has some Hickups today, first not being able to pull from gh-pages, now this is merged but this PR not closed?! Strange.

mateuszbaran:
I don't see this as merged, only closed.

mateuszbaran:
Wait, master branch history shows it's merged but his PR does not. Weird.

kellertuer:
I see a commit that refers to this on master, but also this not marked as merged. But if I open it, the changelog check fails, since the changelog is identical with master – so it _is_ on master. I am confused.

PR #357, opened 2024-02-23
Implementation of two bundle methods

hajg-ijk:
Implement the convex bundle method and the proximal bundle method.

* [x] Check code coverage.
* [x] Carefully recheck the documentation.


kellertuer:
@mateuszbaran  do you see why codecov does not want to commit the code coverage? We have a repository token, but it seems to not be used here by now.

mateuszbaran:
> @mateuszbaran do you see why codecov does not want to commit the code coverage? We have a repository token, but it seems to not be used here by now.

I've looked at it but I'm not sure, let's see if your attempt at fixing it is successful.

kellertuer:
Ah I had to update the CI to explicitly include the token now it works again. Will do that again with the other repositories when we have new PRs. It is something that changed to v4 of the codecov action.

benchmarks/benchmark_bundle_method.jl
```diff
@@ -0,0 +1,31 @@
+using Manopt, Manifolds, Random, QuadraticModels, RipQP, BenchmarkTools, Test
```
kellertuer, line 1: Do we want to keep this file here? I am not yet sure how to provide benchmarks.

examples/bundle_method_examples/L2TVTV2.jl
```diff
@@ -0,0 +1,68 @@
+using ColorSchemes, Random, QuadraticModels, RipQP, Manifolds, Manopt
```
kellertuer, line 1: This file should be moved to ManoptExamples.

hajg-ijk: Actually, all of that folder should be removed as it contains outdated examples.

benchmarks/benchmark_bundle_method.jl
hajg-ijk: This folder should be removed, too.

kellertuer: Did that already :)

hajg-ijk: These files are so old I completely forgot they existed.

src/plans/bundle_plan.jl
```diff
@@ -0,0 +1,126 @@
+#
+# Common files for bunlde-based solvers
+#
+
+function bundle_method_subsolver end
+@doc raw"""
+    bundle_method_subsolver(M, bms<:Union{ConvexBundleMethodState, ProximalBundleMethodState})
```
kellertuer, line 7: Could we change the second argument to be a little less dependent on the actual state?

I imagine, this could also be something one could implemented given a bundle of points and vectors. So that it would be maybe something like `bundle_method_subsolver(M, bundle)`

Sure for the two different bundle methods we would no longer dispatch on the state but implement two functions with different names.

src/plans/bundle_plan.jl
```diff
@@ -0,0 +1,126 @@
...
+    \\
+    \text{s. t.} \quad &
+    \sum_{j ∈ L_l} λ_j = 1,
+    \quad λ_j ≥ 0
+    \quad \text{for all } j ∈ L_l,
+\end{align*}
+```
+where ``L_l = \{k\}`` if ``q_k`` is a serious iterate, and ``L_l = L_{l-1} \cup \{k\}`` otherwise.
+"""
+bundle_method_subsolver(M, s) #change to problem state?
```
kellertuer, line 43: `#change to problem state?` is probably obsolete, but an inlace variant might be nice.
While `\lambda` in general does change from one iteration t the next in size – within the subsolver it should be the same in-place array we can use.

kellertuer (commented):
I checked the code carefully and here are a few small comments we should check/discuss.

src/plans/stopping_criterion.jl
```diff
@@ -623,6 +623,84 @@ function update_stopping_criterion!(c::StopWhenStepsizeLess, ::Val{:MinStepsize}
...
+    has_stopped = length(c.reason) > 0
+    s = has_stopped ? "reached" : "not reached"
+    return "f(x) is NaN:\t$s"
+end
+function show(io::IO, c::StopWhenCostNan)
+    return print(io, "StopWhenCostNan()\n    $(status_summary(c))")
+end
+
+"""
+    StopWhenIterNan <: StoppingCriterion
```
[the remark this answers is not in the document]
hajg-ijk: I remember implementing it because, iirc, in using the ALM for some estimations I was carrying out this would sometimes happen, but I wouldn't be able to reproduce it immediately. I will get back to you.

src/solvers/convex_bundle_method.jl
```diff
@@ -0,0 +1,586 @@
...
+    Y = Y - (inner(M, p, X, Y) / norm(M, p, X)^2 * X)
+    return sectional_curvature(M, p, X, Y)
+end
+
+# Comment ζ_1 out, as it is not needed for computations
+# ! Think about potentially adding a check like isfinite(ζ_1(k_min, diam))
+# function ζ_1(k_min, diam)
+#     (k_min < zero(k_min)) && return sqrt(-k_min) * diam * coth(sqrt(-k_min) * diam)
+#     (k_min ≥ zero(k_min)) && return one(k_min)
+# end
```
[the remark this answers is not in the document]
hajg-ijk: Sure, just that I wouldn't know what its use might be right now.

src/plans/bundle_plan.jl
hajg-ijk: I'm not exactly sure how to treat pre-allocated memory then.

kellertuer: I would have to check for the parameters, but something like `bundle_method_subsolver!(M, \lambda, bundle)` would be the the signature.

src/solvers/convex_bundle_method.jl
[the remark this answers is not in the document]
kellertuer: Me neither, but we use it in the paper and it might be nice to see it is available here as well.

src/solvers/convex_bundle_method.jl
```diff
@@ -0,0 +1,586 @@
...
+    return nothing
+end
+
+function (d::DebugStepsize)(
+    ::P, bms::ConvexBundleMethodState, i::Int
+) where {P<:AbstractManoptProblem}
+    (i < 1) && return nothing
+    Printf.format(d.io, Printf.Format(d.format), get_last_stepsize(bms))
+    return nothing
+end
```
kellertuer, line 586: I would have thought this should be the default fallback for any state. Is that not the case? Then we should maybe introduce it as a generic default for states.

src/solvers/convex_bundle_method.jl
```diff
@@ -0,0 +1,586 @@
...
+    atol_λ::R=eps(),
+    atol_errors::R=eps(),
+    bundle_cap::Int=25,
+    diam::R=π / 3,# k_max -> k_max === nothing ? π/2 : (k_max ≤ zero(R) ? typemax(R) : π/3),
+    domain=(M, p) -> isfinite(f(M, p)),
+    m::R=1e-3,
+    k_min=nothing,
+    k_max=nothing,
+    k_size::Int=100,
+    p_estimate=nothing,
```
kellertuer, line 408: This seems to be unused. there is a `p_estimate` passed to the state but that uses `p`. Can this be removed?

src/plans/bundle_plan.jl
```diff
@@ -0,0 +1,126 @@
...
+
+    StopWhenLagrangeMultiplierLess(tol_errors=1e-6, tol_search_dir=1e-3)
+
+    StopWhenLagrangeMultiplierLess(tol_lag_mult=1e-6)
+
+"""
+mutable struct StopWhenLagrangeMultiplierLess{T,R} <: StoppingCriterion
+    tol_errors::T
+    tol_search_dir::T
+    tol_lag_mult::R
```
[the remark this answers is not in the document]
kellertuer: I went for `mode` instead of type, but the rework was relatively easy, I think the code is nicer now and would allow for modes different from these (maybe checking the Lagrange multipliers even directly) later as well.

src/plans/stopping_criterion.jl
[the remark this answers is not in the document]
kellertuer: Ah! Looking at the code closer (before I wanted to delete it), the name was a bit ambiguous.

I thought you were speaking about the `Iteration` ($k=1,2,3,4,...$) until that hits NaN the algorithms have to run very long.
But what is. actually meant here is the `Iterate` (usually `state.p`). And sure for debugging reasons we can keep that as well.

src/solvers/convex_bundle_method.jl
kellertuer: I now see why we need that :)

kellertuer:
besides three small cases
* no solution returned from the subsolver in CBM (I think?)
* One criterion that I do not fully understand in PBM
* bundle size is hit im PBM

the main thing to implement tests for still are the increase warnings. Maybe by setting up a debug state, triggering debug once, modifying $\xi$ manually and triggering debug again and check for a warning appearing.
Yes 1.6 currently even does run into a warning for CBM, but the cases should be tested thoroughly anyways.

So I think we are on a relatively good way by now. For the first two I do not have much of an idea how to trigger these. I could work on the rest (also the points from above) tomorrow or Tuesday evening.

PR #361, opened 2024-03-01
Checking for non-descent direction in qN

mateuszbaran:
A slightly generalized check originally proposed in #339 . Ideally we would also have an option that purges the memory but that's fairly difficult to do in Manopt. This change may be a little difficult to test.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,12 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.55] unreleased
+
+### Added
+
+* Option `nondescent_direction_behavior` for quasi-Newton solvers. By default it checks for non-descent direction which may not be handled well by some stepsize selection algorithms.
+
 ## [0.4.54] February 28, 2024
 
 ### Added
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.4.54"
+version = "0.4.55"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"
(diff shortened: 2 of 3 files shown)


mateuszbaran:
I'm trying to add a test by defining a custom `struct QuasiNewtonGradientDirectionUpdate <: AbstractQuasiNewtonDirectionUpdate end` that returns gradient as the direction.

test/solvers/test_quasi_Newton.jl
```diff
@@ -357,6 +369,22 @@ using LinearAlgebra: I, eigvecs, tr, Diagonal
...
+        grad_f(M, p) = 2 * sum(p)
+        gmp = ManifoldGradientObjective(f, grad_f)
+        mp = DefaultManoptProblem(M, gmp)
+        qns = QuasiNewtonState(
+            M, p; direction_update=QuasiNewtonGradientDirectionUpdate(ParallelTransport())
+        )
+        @test_logs (
+            :warn,
+            "Computed direction is not a descent direction; resetting to negative gradient",
+        ) match_mode = :any solve!(mp, qns)
```
kellertuer, line 385: I usually leave out the string, though that might be a bit vague.

What does the `match_mode = :any` do? I have not yet seen that before.

mateuszbaran: `match_mode = :any` checks if the string is somewhere in the warning output of the `solve!` call (IIUC without it it needs to be an exact match). How would you prefer to describe it?

kellertuer: Ah if you leave the string empty it just checks that it is a warning but not the string. With containing – maybe just a key part not the full string is enough?

mateuszbaran: I don't really have any preference here, feel free to modify this test as you see fit.

kellertuer: We can also keep it as is for now.

kellertuer (approved):
I think my only remark would be that this changes the default. (`:ignore` was the default before), but this could also be seen as an enhancement.

We could merge this the next few days together with the typo PR as a new version then :)

mateuszbaran:
> I think my only remark would be that this changes the default. (`:ignore` was the default before), but this could also be seen as an enhancement.

Yes, I think this is an enhancement. It adds a bit of additional cost for checking but I think it's worth it.

> We could merge this the next few days together with the typo PR as a new version then :)

Sure, that's fine.

kellertuer (approved):

Issue #347, opened 2024-01-19
Change default stopping criterion in particle swarm


mateuszbaran:
Currently it is distance between old and new swarm but IMO norm of velocities would make more sense. Ref #345 .

kellertuer:
That does sounds reasonable, but it would be great to back that up with (Riemannian or Euclidean) literature on that.

mateuszbaran:
The closest thing I could find is move forcing: https://arxiv.org/pdf/1906.08867.pdf . Do you have any literature that uses the current stopping criterion?

mateuszbaran:
OK, then it's not any better in this regard :wink:

kellertuer:
That is exactly where my careful request about the citation comes from. If we improve it then with a reasonable source for that

mateuszbaran:
It's effectively very similar in the idea, very easy to implement and much faster, I don't see any drawbacks. Standard stopping criteria for PSO are usually more simple, see https://web2.qatar.cmu.edu/~gdicaro/15382/additional/CompIntelligence-Engelbrecht-ch16.pdf . Here is a practical example: https://www.mathworks.com/help/gads/particleswarm.html .

kellertuer:
I see, then sure, let's go for the one that's easier to compute :)

Compared to your idea of the change of the iterate / global best, Matlab at least has something like “if it has not changed for the last X iterations”, which for me also sounds reasonable.

PR #369, opened 2024-03-16
Change PSO SC to a more efficient way of computing.

kellertuer:
I thought a bit about #347. The velocities are actually equivalent to the change (the same if the steps are computed with exp). So this PR changes the stopping criterion of SC to this – a bit nicer – form.

Still to check

* [x] documentation
* [x] test coverage
* [x] changelog

### DIFF
--- a/test/solvers/test_particle_swarm.jl
+++ b/test/solvers/test_particle_swarm.jl
@@ -67,7 +67,7 @@ using Random
         s = particle_swarm(M, f, swarm)
         @test s ≈ 0.0
     end
-    @testset "Specific Stoping criteria" begin
+    @testset "Specific Stopping criteria" begin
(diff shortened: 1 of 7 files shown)


test/solvers/test_particle_swarm.jl
```diff
@@ -67,4 +67,8 @@ using Random
         s = particle_swarm(M, f, swarm)
         @test s ≈ 0.0
     end
+    @testset "Specific Stoping criteria" begin
```
mateuszbaran, line 70:
```suggestion
    @testset "Specific Stopping criteria" begin
```

mateuszbaran (approved):
LGTM

kellertuer:
Will merge this but only register it together with #370 once that is also merged

PR #376, opened 2024-04-05
Update `StopWhenAny` constructor for `Vector`

Vaibhavdixit02:
The element type should be subtype of `StoppingCriterion`


kellertuer (approved):
Thanks for the quick adaptions and fixing that bug :)

PR #377, opened 2024-04-07
Introduce support of subsolver records.

### LINKED ISSUE #371: Provide a collection of records from a subsolver
For sub solvers `record= ... ` passed to `sub_kwargs =` should already correctly enable recordings for the subsolver (similar and fully analogous to debug). Though for now, there is no way to “collect” these recordings.

This could be done nearly-automatic, probably with a `SubSolverRecordAction` in the main solver, that after every iteration collects all (or some) of the recorded values after every step of the main solver.
this should not be too complicated, but might be a bit technical.

kellertuer:
This is a PR to resolve #371.

* generalise the record factories to work more modular (similar to debug), this allows to easier add elements to `:Stop`
* add a `RecordStoppingReason` action to record the stopping reason (dah!)
* add a `RecordWhenActive` (analogue to debug), though not 100% necessary, might be nice to reduce allocations when only requiring every $n$-th subsolver run to actually be recorded
* add a `RecordSubsolverRecords` action that stores the sub solvers record (or parts of it) in the parents record entry for that iteration

# ToDo
* [x] verify and check that the idea works
* [x] extend docuemntation
* [x] write tests
* [x] extend tutorial
* [x] check test coverage

# Open Points

Currently the symbol `:Subsolver` is used _in_ subsolvers to declare them beding dependent (of the parent), while it might be breaking `:WhenActive` might be a nicer symbol. The `:Subsolver` could be used to indicate subsolver recordings. I have to think about this a bit and whether this is seriously breaking, since for now `:Subsolver`.  On the other hand it is not breaking code to the extend it does now work, just that since the last release to this then some might get a bit more subsolver debug.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
         os: [ubuntu-latest, macOS-latest, windows-latest]
     steps:
       - uses: actions/checkout@v4
-      - uses: julia-actions/setup-julia@v1
+      - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.julia-version }}
           arch: x64
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: julia-actions/setup-julia@v1
+      - uses: julia-actions/setup-julia@v2
         with:
           version: 1
       - name: Install JuliaFormatter and format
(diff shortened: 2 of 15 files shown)


kellertuer:
The design is nice and seems to work in general. I went for `:WhenActive` that is much nicer to use in writing

The main thing that seems to bother a bit still is, that though there is a lot of resets implemented, it seems that the recording of subsolver recordings is still incremental, this needs to be fixed still.
So though recorded values are reset to the empty array the next push still pushes to the previous recordings array – no clue yet why.

kellertuer:
I think in the rework I destroyed one Pair case, will check tomorrow, but the tutorial is extended (just failing now).

PR #379, opened 2024-04-12
Introduce Aqua.jl-based checks

kellertuer:
I added tests with `Aqua.jl`.

While we have a few ambiguity warnings that I for now ignored, they appear in different variants of the high-level solvers, that usually should not be a problem.

But there is one case of unbounded parameters that I could not resolve. Maybe @mateuszbaran you can take a look? The problem with 

https://github.com/JuliaManifolds/Manopt.jl/blob/f5461838e7b6d03107522fcfe3a544aba306d222/src/plans/solver_state.jl#L415-L437

is that in case it is a vector of symbols, the two parameters `TTS, TPS` are declared but not use; I tried to localise them to but them into the union (the trick that otherwise helped), but then the code uses `TPS` and `TTS` and that is not possible if they are local to the union. Is there any other good way to access the parameters you need?

I will keep this open for a while and maybe work a bit on the ambiguities, but maybe that would require a breaking change as well. At least for the last step size I will try to resolve that.


kellertuer:
I think I found something. This was fun and I removed a few ambiguities already. The remaining ones I might think about, but they are a bit tricky, when we allow both `solver(M, f, grad_f, rest...; kwargs...)` as well as `solver(M, objective, rest...; kwargs...)`
especially when `rest` also contains the two forms for sub solvers we currently support (sub-problem/state and closed-form variant as a function).

mateuszbaran:
> But there is one case of unbounded parameters that I could not resolve. Maybe @mateuszbaran you can take a look? The problem with

I will rewrite it 

> This was fun and I removed a few ambiguities already. The remaining ones I might think about, but they are a bit tricky, when we allow both `solver(M, f, grad_f, rest...; kwargs...)` as well as `solver(M, objective, rest...; kwargs...)`

This may be a bit hard to resolve.

kellertuer:
You would also usually not mix them up, because that would basically require to provide ab objective, a problem and another problem or such, so I am even fine with keeping the high-level interfaces on ignore.
It is also just a few ones that have different formats to be called: difference of convex have a few different formats that “conflict” in an ambiguity you would in reality never hit.

kellertuer:
> I will rewrite it 

Oh a very nice rewrite dispatching on that again :) Well-done.

kellertuer:
I checked both the `get_last_stepsize` and the higher level ones.
the last step size would need a bit of a (breaking) rework, that I could do before a next breaking release.

For the high level ones, I neither feel that would be easy to do nor that it might be a problem that there is an ambiguity. One would basically have to really mix two of the documented forms _and_ provide two sub problem solvers, one in closed form , one as a problem, and not provide a state.
The reason I do not want to narrow down those parameters in type (which could also resolve the ambiguity) is that I do not want to restrict `f` or `grad_f` to for example `Function`.

mateuszbaran:
`particle_swarm` should be actually fixed IMO, see this call:
```julia
        M = Circle()
        f(::Circle, p) = p * 2 * p
        swarm = [-π / 4, 0, π / 4]
        s = particle_swarm(M, ManifoldCostObjective(f), swarm)
        @test s ≈ 0.0
```
I will take a look at other ones too.

mateuszbaran:
`truncated_conjugate_gradient_descent` ambiguity should be harmless, and `get_last_stepsize` should probably be redesigned to fix those ambiguities.

kellertuer:
Ah I See the point with particle swarm, the main tricky thing seems really to be that we aim to cover single-number-cases as well, but it should be fine with just one extra case each hopefully then.

for the last stepsize I agree; I would consider that function more internal though, but currently its signatures are a bitt all-obver-the.place and should be minimised / unified, I can check whether I see a way without breaking (too much).

kellertuer:
I fixed last stepsize. There were a few too many interims dispatches that caused trouble, so I removed an unneccesary interims one, that was also only explicitly called in a test but nowhere else used. I documented the remaining ones more carefully and added an entry to the changelog about this

ext/ManoptManifoldsExt/manifold_functions.jl
```diff
@@ -58,3 +58,73 @@ function mid_point!(M::Sphere, y, p, q, x)
...
+
+````math
+    \operatorname{refl}_f(x) = \operatorname{refl}_{f(x)}(x),
+````
+
+Compute the result in `q`.
+
+see also [`reflect`](@ref reflect(M::AbstractManifold, p, x))`(M,p,x)`, to which the keywords are also passed to.
+"""
+reflect(M::AbstractManifold, pr::Function, x; kwargs...) = reflect(M, pr(x), x; kwargs...)
```
mateuszbaran, line 77: Technically it's type piracy so maybe it would be better to introduce a new function for this functionality?

kellertuer: Hm, complicated, because the reflection at the prox – which is what this is – does still bear the name reflection.

Yes I moved it here because Aqua said in the Manopt file it is type piracy, but in there? Hm, I am not sure.
But the worst is that I neither have a good name _and_ it would really be breaking to rename that, in the sense that it breaks all my Douglas-Rachford experiments.

mateuszbaran: It's still piracy in the extension, Aqua just doesn't test for that yet: https://github.com/JuliaTesting/Aqua.jl/issues/160 . This is not a high priority thing but I'd consider renaming for the next breaking release.

mateuszbaran: So you haven't fixed the problem, only swept it under the rug :sweat_smile:

kellertuer: hm, yeah, maybe. 
I would maybe move this to Manifolds, that would resolve it,
just if one wants to be precise the “original” reflect is the manifold one while this is the Manopt one – so I am actually a bit unsure how to resolve this without piracy

kellertuer: The only real solution I see here is, to move this to ManifoldsBase.jl as well.

mateuszbaran: Reflecting at point like fine to me and we can move that to ManifoldsBase but the methods that dispatch on `Function` are questionable. What if someone wanted to represent points on a manifold by functions? We would have an ambiguity.

kellertuer: The only answer I have to that is, that I have no clue about that. But for now we do not have that and I think representing points by functions would be a major rework anyways. In all 3 packages.

mateuszbaran: I did that some time ago in FunManifolds.jl. That's a very niche application but I don't want to introduce new issues that could affect it if "let's make a new function with a different name" is all it takes to avoid it.

kellertuer: But we could wrap that function in a subtype of ManifoldPoint?
The problem with points being functions is that _all_ which-leverl interfaces of Manopt.jl would not be able to handle them anyways. That is how I currently distinguish whether to create a random starting point, that a point is not a function.

My main problem with a new name is, that mathematically both have the same name here, so introducing a new name would confuse everyone reading the papers. So that takes all to confuse everyone, sure.

mateuszbaran: That's fine in limited scenarios but note that the confusion may also work the other way: someone makes a functor object, doesn't derive from `Function` for whatever reason, and is surprised with a weird error. You can perhaps call the function variant `reflect_by`?

mateuszbaran: So "something is function-like only if its type is `<:Function` is also a flawed assumption.

kellertuer: Sure, but in some places I did not yet see any other way for the “convenience” layer to distinguish between a Hessian being provided (and not needed to be approximated) or a point (and no random is needed). And maybe this is not a good design, but I am not clever enough to come up with a better one that is simple to use.

As I said, I do not have a good idea for a name. the `_by` would sound strange, for both a point and a function one would say you reflect _at_, but then this does not help, since it would refer to both.

kellertuer: > So "something is function-like only if its type is <:Function is also a flawed assumption.

Yes, sure.

kellertuer: But I also only do that on the high-level interfaces `solver(M, f, ...)`not on problem/state.

The idea is (also maybe not good) that as long as you are in the simple regime, the high level stuff should be easy to use.

mateuszbaran: Dispatching on functions is, as I said, fine in limited scenarios. So high-level solver interface doing it is not a big issue. But I don't want to propagate that design to ManifoldsBase.jl, especially when the alternative is just splitting it into two functions.

kellertuer: I will think about a name. But me and thinking about names is something random between tomorrow and next year :/

mateuszbaran: So let's keep `reflect` in Manopt until you figure it out :slightly_smiling_face:

kellertuer: Yes, I think that for now is the best. And I for now also feel that the piracy here is also only a mild one, I define a function based on a ManifoldsBase type, that in principle I “own” as well, but sure with `reflect!` I pirate then a bit since the function also comes from LinearAlgreba.

But yes I will keep that in mind and think about it to hopefully find a good alternative, maybe something keyword based `at_point=` and `at_function` or such (just the function name `reflect_at_point` seems to long for me again).

mateuszbaran: Sure, it can also be handled by keywords.

kellertuer:
I've taken a look at `particle_swarm`. The main problem are the two cases

* `particle_swarm(M, f, swarm)` versus
* `particle_swarm(M, mco, swarm)`

where for the first we have a method that also works if the swarm is a vector of numbers. For the second we do not do this, which leads to an ambiguity. the problem is here, that while `mco` does cary a cost inside and we could use `get_cost_function` to “extract” it and make it suitable to numbers, we can not “exchange” it in `mco`; at least I have no clue how to do that if the `mco` is for example “wrapped” in a cache.

Long story short, I can only resolve this, by writing a `particle_swarm(M, mco, swarm::AbstractVector{T}) where {T<:Number}` that errors. The only alternative would be to remove the variant that works on numbers, but that would even break the tests we have on that.

kellertuer:
`difference_of_convex_proximal_point` has exactly the same problem, namely that the only thing mandatory is a single function, just that there it is `grad_h` (instead of `f`), but the effect is completely the same as for PSO.

mateuszbaran:
I see, that's indeed rather unfixable. I'd suggest adding an error hint like here:
https://github.com/JuliaManifolds/Manifolds.jl/blob/4cbf30f44654f0be462193fdea7a164a3d746264/src/Manifolds.jl#L581-L588 and keeping it an ambiguity.

kellertuer:
`truncated_conjugate_gradient_descent` does not only cover a few deprecated cases, but also one where it is hard to distinguish the (optional) hessian and the optional initial value (wich unfortunately is both a point `p` and the tangent vector `X`), such that `X` and hessian get into conflict.
That leads to 3 ambiguities (the other 3 are resolved by a breaking change, removing the deprecated ones).

I am not sure what to do about these, since we usually do not type `p` nor `X` and while the hessian is currently “reconized” by being a function, we already discussed that we maybe do not want that either.
Though one could do functors to subtype `Function` that is maybe also restrictive. But then maybe the general question is:

If both a function am the initial point are optional, how to best distinguish the? For best of cases I would prefer to keep both positional, since that is “nice to use”, but hard to get unambiguous, probably.
So an alternative (but breaking change) could be to move all optional functions (like here the hessian) into (very unified named) keyword arguments.
(This would not resolve the two problems above, but well).

kellertuer:
> I see, that's indeed rather unfixable. I'd suggest adding an error hint like here:
> https://github.com/JuliaManifolds/Manifolds.jl/blob/4cbf30f44654f0be462193fdea7a164a3d746264/src/Manifolds.jl#L581-L588 and keeping it an ambiguity.

I have no clue how to write such a hint, but something like “doing PSO with a objective and a swarm consisting of numbers is not supported” – one could of course also just start such an algorithm, assuming that “A user providing an objective” knows what they do and it will be the right one already, so one would just generate the “array swarm”.

kellertuer:
Ah, one thing that might help (but maybe that is also a bit silly interface-wise) is to for example make the initial point always a tuple, so one would call

`difference_of_convex_proximal_point(M, grad_h, (p0,))`

which would remove the ambiguity with the Hessian, though not with the f-vs-objective (or grad-h-vs-objective) one.
I would prefer keeping p0 as the (always last) positional argument though, I think – because we dispatch on that for number types at least; otherwise one could consider moving p0 to a `initial_value` keyword or so (but that would again be quite breaking as well).

mateuszbaran:
> I have no clue how to write such a hint, but something like “doing PSO with a objective and a swarm consisting of numbers is not supported” – one could of course also just start such an algorithm, assuming that “A user providing an objective” knows what they do and it will be the right one already, so one would just generate the “array swarm”.

I will write that hint then.

> `truncated_conjugate_gradient_descent` does not only cover a few deprecated cases, but also one where it is hard to distinguish the (optional) hessian and the optional initial value (wich unfortunately is both a point `p` and the tangent vector `X`), such that `X` and hessian get into conflict.

OK, could you mention it in a comment in the test file for Aqua?

> If both a function am the initial point are optional, how to best distinguish the? For best of cases I would prefer to keep both positional, since that is “nice to use”, but hard to get unambiguous, probably.
> So an alternative (but breaking change) could be to move all optional functions (like here the hessian) into (very unified named) keyword arguments.

 I would suggest making Hessian a keyword argument, and keeping the rest positional. 7 different calls signatures in a docstring is a bit much, and it doesn't have the "no Hessian, no p" variant there which should probably also work?

kellertuer:
> I would suggest making Hessian a keyword argument, and keeping the rest positional. 7 different calls signatures in a docstring is a bit much, and it doesn't have the "no Hessian, no p" variant there which should probably also work?

That we should check carefully and introduce then for all solvers that need a hessian but are fine with an approximation (or maybe even more general: all solvers that require objective-parts that are optional / can be “filled” with a good default).

mateuszbaran:
> Ah, one thing that might help (but maybe that is also a bit silly interface-wise) is to for example make the initial point always a tuple, so one would call
> 
> `difference_of_convex_proximal_point(M, grad_h, (p0,))`
> 
> which would remove the ambiguity with the Hessian, though not with the f-vs-objective (or grad-h-vs-objective) one. I would prefer keeping p0 as the (always last) positional argument though, I think – because we dispatch on that for number types at least; otherwise one could consider moving p0 to a `initial_value` keyword or so (but that would again be quite breaking as well).

Making initial point a tuple looks awful to me, and since it's I think the most important optional parameter of every solver, I would try to avoid making it a keyword argument.

> That we should check carefully and introduce then for all solvers that need a hessian but are fine with an approximation (or maybe even more general: all solvers that require objective-parts that are optional / can be “filled” with a good default).

Sure, that sounds like a good idea.

kellertuer:
Ah to be precise `p` for TCG can never be optional, since it specifies the tangent space, the iterate is `X` which is optional, I am not sure a random tangent space as a domain is something reasonable ;) 

But I do agree 7 signatures might be a few too much.

mateuszbaran:
> Ah to be precise `p` for TCG can never be optional, since it specifies the tangent space, the iterate is `X` which is optional, I am not sure a random tangent space as a domain is something reasonable ;)

There is such signature there (the third one):
```julia
truncated_conjugate_gradient_descent(M, f, grad_f, Hess_f; kwargs...)
```

kellertuer:
> Making initial point a tuple looks awful to me, and since it's I think the most important optional parameter of every solver, I would try to avoid making it a keyword argument.

I fully agree that that would be quite ugly, and I feel the idea “`p=rand(M)` is always the last positional argument” (for TCG: a random tangent vector ;)) is a very good design idea.

Then we should check for the next breaking release to make things like an optional hessian a keyword argument. Some solvers already have `f` or such as keywords, where they can be filled reasonably or are only necessary for more debug opportunities.

kellertuer:
I first thought this might just be a reasonable check to add, turns out, it even improves stability / reliability rethinking a few choices that lead to the ambiguities.

kellertuer:
Is it ok if we do the rework of the interface (and their unification) in a new PR? I can also open an issue (and add that to the 1.0 roadmap) that sketches how to unify the high level interfaces and how they accept functions

mateuszbaran (approved):
Yes, sure, that's fine.

PR #384, opened 2024-04-26
add Polyak's Step size.

### LINKED ISSUE #224: Polyak's stepsize
Example reference: https://web.stanford.edu/class/ee364b/lectures/subgrad_method_slides.pdf . This looks relatively simple :).

kellertuer:
Resolves #224. Also yields a good reason to release 0.4.61.

What would be the best literature to reference here, @mateuszbaran?

### DIFF
--- a/src/plans/gradient_plan.jl
+++ b/src/plans/gradient_plan.jl
@@ -203,6 +203,24 @@ function get_gradient!(M::AbstractManifold, X, mcgo::ManifoldCostGradientObjecti
     return X
 end
 
+"""
+    X = get_subgradient(M;;AbstractManifold, sgo::AbstractManifoldGradientObjective, p)
+    get_subgradient!(M;;AbstractManifold, X, sgo::AbstractManifoldGradientObjective, p)
+
+Evaluate the subgradient, which for the case of a objective having a gradient, menas evaluating the
+gradient itself.
+
+While in general, the result might not be deterministic, for this case it is.
+"""
+function get_subgradient(M::AbstractManifold, agmo::AbstractManifoldGradientObjective, p)
+    return get_gradient(M, agmo, p)
+end
+function get_subgradient!(
+    M::AbstractManifold, X, agmo::AbstractManifoldGradientObjective, p
+)
+    return get_gradient!(M, X, agmo, p)
+end
+
 function _to_mutating_gradient(grad_f, evaluation::AllocatingEvaluation)
     return grad_f_(M, p) = [grad_f(M, p[])]
 end
--- a/src/plans/stepsize.jl
+++ b/src/plans/stepsize.jl
@@ -764,7 +764,7 @@ get_message(a::NonmonotoneLinesearch) = a.message
 Compute a step size
 
 ````math
-α_i = \frac{f(p^{(i)}) - f_{\text{best}}+γ_k}{\lVert{ ∂f(p^{(i)})} \rVert^2},
+α_i = \frac{f(p^{(i)}) - f_{\text{best}}+γ_i}{\lVert{ ∂f(p^{(i)})} \rVert^2},
 ````
 
 where ``f_{\text{best}}`` is the best cost value seen until the current iteration,
--- a/src/plans/stepsize.jl
+++ b/src/plans/stepsize.jl
@@ -764,11 +764,15 @@ get_message(a::NonmonotoneLinesearch) = a.message
 Compute a step size
 
 ````math
-α_i = \frac{f(p^{(k)}) - f_{\text{best}}+\gamma_k}{\lVert{ \operatorname{grad} f(p^{(k)})} \rVert^2},
+α_i = \frac{f(p^{(k)}) - f_{\text{best}}+γ_k}{\lVert{ ∂f(p^{(k)})} \rVert^2},
 ````
 
 where ``f_{\text{best}}`` is the (up to now) best cost function seen until now,
-and ``\gamma_i`` is a sequence of numbers that is square summable, but not summable.
+and ``γ_i`` is a sequence of numbers that is square summable, but not summable.
+Furthermore ``∂f`` denotes a subgradient of ``f`` at the current iterate ``p^{(k)}``.
+
+The step size is an adaption from the Dynamic step size discussed in Section 3.2 of [Bertsekas:2015](@cite),
+both to the Riemannian case and to approximate the minimum cost value.
 
 # Fields
 
@@ -797,11 +801,11 @@ function (ps::PolyakStepsize)(
     # We get these by reference, so that should not allocate in general
     M = get_manifold(amp)
     p = get_iterate(ams)
-    X = get_gradient(amp, p)
+    X = get_subgradient(amp, p)
     # Evaluate the cost
     c = get_cost(M, get_objective(amp), p)
     (c < ps.best_cost_value) && (ps.best_cost_value = c)
-    α = (c - ps.best_cost_value + ps.γ(i)) / norm(M, p, X)
+    α = (c - ps.best_cost_value + ps.γ(i)) / (norm(M, p, X)^2)
     return α
 end
 function show(io::IO, ps::PolyakStepsize)
--- a/src/plans/stepsize.jl
+++ b/src/plans/stepsize.jl
@@ -503,6 +503,7 @@ Then find the new stepsize by
 ```
 
 # Fields
+
 * `initial_stepsize`:        (`1.0`) the step size to start the search with
 * `memory_size`:             (`10`) number of iterations after which the cost value needs to be lower than the current one
 * `bb_min_stepsize`:         (`1e-3`) lower bound for the Barzilai-Borwein step size greater than zero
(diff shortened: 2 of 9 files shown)


src/plans/stepsize.jl
```diff
@@ -756,6 +757,60 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
 get_message(a::NonmonotoneLinesearch) = a.message
 
+@doc raw"""
+    PolyakStepsize <: Linesearch
+
+
+Compute a step size
+
+````math
+α_i = \frac{f(p^{(k)} - f_{\text{best}}+\gamma_k}{\lVert \operatorname{grad} f(p^{(k)})} \rVert},
```
mateuszbaran, line 767: α_i = \frac{f(p^{(k)}) - f_{\text{best}}+\gamma_k}{\lVert{ \operatorname{grad} f(p^{(k)})} \rVert},

src/plans/stepsize.jl
```diff
@@ -756,6 +757,60 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
+````
+
+where ``f_{\text{best}}`` is the (up to now) best cost function seen until now,
+and ``\gamma_i`` is a sequence of numbers that is square summable, but not summable.
```
mateuszbaran, line 771: I think you should mention `grad f` here, and also note that this stepsize also works for subgradients.

mateuszbaran:
> What would be the best literature to reference here, @mateuszbaran?

I like the explanation of this stepsize from this book: https://www.amazon.com/Convex-Optimization-Algorithms-Dimitri-Bertsekas/dp/1886529280 , section 3.2. It also has a few other variants.

src/plans/stepsize.jl
```diff
@@ -756,6 +757,60 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
+    amp::AbstractManoptProblem, ams::AbstractManoptSolverState, i::Int, args...; kwargs...
+)
+    # We get these by reference, so that should not allocate in general
+    M = get_manifold(amp)
+    p = get_iterate(ams)
+    X = get_gradient(amp, p)
+    # Evaluate the cost
+    c = get_cost(M, get_objective(amp), p)
+    (c < ps.best_cost_value) && (ps.best_cost_value = c)
+    α = (c - ps.best_cost_value + ps.γ(i)) / norm(M, p, X)
```
mateuszbaran, line 804:
```suggestion
    α = (c - ps.best_cost_value + ps.γ(i)) / norm(M, p, X)^2
```

mateuszbaran:
By the way, thanks for working on this :+1: .

src/plans/stepsize.jl
kellertuer: for now I just mentioned the two things stored in the struct, but you are right we need a second variant for sub gradients as well of course.

src/plans/stepsize.jl
kellertuer: Oh then I might have to fix the code as well here.

kellertuer:
> By the way, thanks for working on this 👍 .

No problem, I was looking for something easy to finish the day yesterday, and (besides missing the square) this actually was.

src/plans/gradient_plan.jl
```diff
@@ -203,6 +203,24 @@ function get_gradient!(M::AbstractManifold, X, mcgo::ManifoldCostGradientObjecti
     return X
 end
 
+"""
+    X = get_subgradient(M;;AbstractManifold, sgo::AbstractManifoldGradientObjective, p)
+    get_subgradient!(M;;AbstractManifold, X, sgo::AbstractManifoldGradientObjective, p)
```
mateuszbaran, line 208:
```suggestion
    X = get_subgradient(M::AbstractManifold, sgo::AbstractManifoldGradientObjective, p)
    get_subgradient!(M::AbstractManifold, X, sgo::AbstractManifoldGradientObjective, p)
```

src/plans/gradient_plan.jl
```diff
@@ -203,6 +203,24 @@ function get_gradient!(M::AbstractManifold, X, mcgo::ManifoldCostGradientObjecti
...
+
+Evaluate the subgradient, which for the case of a objective having a gradient, menas evaluating the
```
mateuszbaran, line 210:
```suggestion
Evaluate the subgradient, which for the case of a objective having a gradient, means evaluating the
```

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
+# Constructor
+
+    PolyakStepsize(;
+        γ = i -> 1/i,
+        initial_cost_estimate=0.0
+    )
+
+initialize the Polyak stepsize to a certain sequence and an initial estimate of ``f_{\text{best}}``
+"""
+mutable struct PolyakStepsize{F,R} <: Linesearch
```
mateuszbaran, line 791: Shouldn't it be a `Stepsize` instead of `Linesearch`?

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
+    PolyakStepsize <: Linesearch
+
+
+Compute a step size
+
+````math
+α_i = \frac{f(p^{(k)}) - f_{\text{best}}+γ_k}{\lVert{ ∂f(p^{(k)})} \rVert^2},
+````
+
+where ``f_{\text{best}}`` is the (up to now) best cost function seen until now,
```
mateuszbaran, line 770:
```suggestion
where ``f_{\text{best}}`` is the best cost value seen until the current iteration,
```

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
+and ``γ_i`` is a sequence of numbers that is square summable, but not summable.
```
mateuszbaran, line 771:
```suggestion
and ``γ_i`` is a sequence of numbers that is square summable, but not summable (the sum must diverge to infinity).
```

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
 get_message(a::NonmonotoneLinesearch) = a.message
 
+@doc raw"""
+    PolyakStepsize <: Linesearch
+
+
+Compute a step size
+
+````math
+α_i = \frac{f(p^{(k)}) - f_{\text{best}}+γ_k}{\lVert{ ∂f(p^{(k)})} \rVert^2},
```
mateuszbaran, line 767: We should consistently use either `i` or `k` for the current iteration.

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
+    # Evaluate the cost
+    c = get_cost(M, get_objective(amp), p)
+    (c < ps.best_cost_value) && (ps.best_cost_value = c)
+    α = (c - ps.best_cost_value + ps.γ(i)) / (norm(M, p, X)^2)
+    return α
+end
+function show(io::IO, ps::PolyakStepsize)
+    return print(
+        io,
+        "PolyakStepsize() with keyword parameters\n  * initial_cost_estimate = $(ps.best_cost_value)",
```
mateuszbaran, line 814: `best_cost_value` may be different from `initial_cost_estimate` so this could be misleading. Maybe let's either call it `best_cost_value` or not print it at all.

mateuszbaran (commented):
Just a few minor points, other than that it looks fine.

src/plans/stepsize.jl
kellertuer: ah Yes. Confused them.

src/plans/stepsize.jl
kellertuer: I somehow started the package with `i` but otherwise usually use `k`, so I sometimes confuse them.

src/plans/stepsize.jl
kellertuer: If you want to start from the time printing you have to set that number as the initial one, that's why I printed it as such (also to reflect the keyword).

kellertuer: But sure we could also not print it if you prefer, I would prefer to provide that information as the current state.

mateuszbaran: I see, then maybe let's keep it as is.

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
+
+Compute a step size
+
+````math
+α_i = \frac{f(p^{(i)}) - f_{\text{best}}+γ_k}{\lVert{ ∂f(p^{(i)})} \rVert^2},
+````
+
+where ``f_{\text{best}}`` is the best cost value seen until the current iteration,
+and ``γ_i`` is a sequence of numbers that is square summable, but not summable (the sum must diverge to infinity).
+Furthermore ``∂f`` denotes a subgradient of ``f`` at the current iterate ``p^{(k)}``.
```
mateuszbaran, line 772:
```suggestion
Furthermore ``∂f`` denotes a subgradient of ``f`` at the current iterate ``p^{(i)}``.
```

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
...
 get_message(a::NonmonotoneLinesearch) = a.message
 
+@doc raw"""
+    PolyakStepsize <: Linesearch
+
+
+Compute a step size
+
+````math
+α_i = \frac{f(p^{(i)}) - f_{\text{best}}+γ_k}{\lVert{ ∂f(p^{(i)})} \rVert^2},
```
mateuszbaran, line 767:
```suggestion
α_i = \frac{f(p^{(i)}) - f_{\text{best}}+γ_i}{\lVert{ ∂f(p^{(i)})} \rVert^2},
```

src/plans/stepsize.jl
```diff
@@ -756,6 +757,64 @@ function show(io::IO, a::NonmonotoneLinesearch)
 end
 get_message(a::NonmonotoneLinesearch) = a.message
 
+@doc raw"""
+    PolyakStepsize <: Linesearch
```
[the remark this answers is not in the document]
kellertuer: Eh? That should be in the last commit.

src/plans/stepsize.jl
kellertuer: Oh, one might notice, I am just on my first weekend after a stressful semester.

src/plans/stepsize.jl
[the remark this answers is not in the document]
mateuszbaran: It looks like you missed it in the docstring.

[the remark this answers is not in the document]
kellertuer: Ah this is the doctoring, I was confused. Maybe that by now becomes my general state.

mateuszbaran (approved):

src/plans/stepsize.jl
mateuszbaran: Don't worry, maybe a little bit of rest would be a good idea then :slightly_smiling_face:

kellertuer: Went on a longer bike ride this morning already, so I am on my way there :)

Issue #382, opened 2024-04-22
Quasi-Newton update rules issues


mateuszbaran:
I'm currently working on some updates to quasi-Newton direction update rules. Here are a few things to check/consider:
- [x] `QuasiNewtonCautiousDirectionUpdate` doesn't seem to use `θ` function anywhere.
- [ ] `QuasiNewtonLimitedMemoryDirectionUpdate` doesn't seem to use `scale`.
- [x] Some direction updates need a reset mechanism called by `initialize_solver!`, and maybe also when the direction stops being a descent direction. That shouldn't happen for a well-designed set of options but it's a bit hard to tell which ones are safe from that issue.
- [ ] Unify projections to a custom function instead of a boolean flag.

I'm also working on some variant of direction update for manifolds with corners inspired by L-BFGS-B. There are different approaches with elaborate line searches and update matrices but I'm aiming for something relatively simple and generic. I'm not aiming at competing with Fortran codes, my goal is making combined manifold-box constrained optimization work relatively well.

kellertuer:
I started looking a bit into these points.

Ad 1) The function is used when updating the Hessian at

https://github.com/JuliaManifolds/Manopt.jl/blob/1856a26d8bce827f7fa97176a9cd1b2effbf0863/src/solvers/quasi_Newton.jl#L677
or
https://github.com/JuliaManifolds/Manopt.jl/blob/1856a26d8bce827f7fa97176a9cd1b2effbf0863/src/solvers/quasi_Newton.jl#L738

Ad 2) That might indeed be the case, I could only find the scale in the update of the full-matrix variants.

https://github.com/JuliaManifolds/Manopt.jl/blob/1856a26d8bce827f7fa97176a9cd1b2effbf0863/src/solvers/quasi_Newton.jl#L514-L516

I am not sure where the scaling was “lost” in one of the reworks, but we can surely bring that back

Ad 3) Sure, I am not 100% sure how to realise that, but when we have the iterate available doing a reset when that is zero sounds reasonable. When we wrote this code, reusing the state (and hence the updates) was not so much thought given to, for sure.

* [ ] I also noticed that one of the recent updates that meant to introduce a stability idea did

https://github.com/JuliaManifolds/Manopt.jl/blob/1856a26d8bce827f7fa97176a9cd1b2effbf0863/src/plans/quasi_newton_plan.jl#L500
and
[...]

PR #388, opened 2024-05-09
`:reinitialize_direction_update` for quasi-Newton

mateuszbaran:
Addresses point 3 from #382. Also a minor cleanup for `QuasiNewtonCautiousDirectionUpdate`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,12 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.63] unreleased
+
+### Added
+
+* `:reinitialize_direction_update` option for quasi-Newton behavior when the direction is not a descent one.
+
 ## [0.4.62] May 3, 2024
 
 ### Changed
--- a/src/plans/quasi_newton_plan.jl
+++ b/src/plans/quasi_newton_plan.jl
@@ -10,6 +10,13 @@ abstract type AbstractQuasiNewtonDirectionUpdate end
 
 get_message(::AbstractQuasiNewtonDirectionUpdate) = ""
 
+"""
+    initialize_update!(s::AbstractQuasiNewtonDirectionUpdate)
+
+Initialize direction update. By default no change is made.
+"""
+initialize_update!(s::AbstractQuasiNewtonDirectionUpdate) = s
+
 @doc raw"""
     AbstractQuasiNewtonUpdateRule
 
@@ -413,6 +420,11 @@ function (d::QuasiNewtonMatrixDirectionUpdate{T})(
     get_vector!(M, r, p, -d.matrix \ get_coordinates(M, p, X, d.basis), d.basis)
     return r
 end
+function initialize_update!(d::QuasiNewtonMatrixDirectionUpdate)
+    copyto!(d.matrix, I)
+    return d
+end
+
 @doc raw"""
     QuasiNewtonLimitedMemoryDirectionUpdate <: AbstractQuasiNewtonDirectionUpdate
 
@@ -584,6 +596,17 @@ function (d::QuasiNewtonLimitedMemoryDirectionUpdate{InverseBFGS})(r, mp, st)
     return r
 end
 
+"""
+    initialize_update!(d::QuasiNewtonLimitedMemoryDirectionUpdate)
+
+Initialize the limited memory direction update by emptying the memory buffers.
+"""
+function initialize_update!(d::QuasiNewtonLimitedMemoryDirectionUpdate)
+    empty!(d.memory_s)
+    empty!(d.memory_y)
+    return d
+end
+
 @doc raw"""
     QuasiNewtonCautiousDirectionUpdate <: AbstractQuasiNewtonDirectionUpdate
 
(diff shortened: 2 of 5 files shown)


kellertuer:
Looks good at first glance (will take a closer look later); documenter fails, because of caching; it might use the cache from the other PR (of cached packages). Maybe that is a disadvantage of caching packages.

mateuszbaran:
Are ALM and EPM supposed to keep the state of `QuasiNewtonLimitedMemoryDirectionUpdate` between runs of the subsolver? I would guess they shouldn't but currently they do.

mateuszbaran:
Caching seems to have fixed itself. By the way, I think `:reinitialize_direction_update` would be a better default for `:nondescent_direction_behavior`, though it might be considered a breaking change.

kellertuer:
I think that sounds nice, how breaking would that be? The old one just took a gradient direction right? I would consider that mildly breaking and still ok

mateuszbaran:
Not that breaking, I will change it if you are OK with it.

kellertuer:
We also introduced the rule for how to handle non-descent directions as a non-braking improvement, so I am fine with this as well, but please write a detailed change entry about it.

mateuszbaran:
Is it sufficiently detailed in the latest commit?

kellertuer:
I prefer to end keywords with `=` but the rest is exactly the details I was hoping for.
Yes this is formerly breaking but I would consider such small improvements in defaults really more like improvements.
Yes we do break 100% reproducibility here, but we do not break existing code in the sense that it errors. So I think its fine to keep that nonbreaking

Changelog.md
```diff
@@ -5,6 +5,20 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.63] unreleased
+
+### Added
+
+* `:reinitialize_direction_update` option for quasi-Newton behavior when the direction is not a descent one. It is now the new default for `QuasiNewtonState`.
+
```
kellertuer, line 13: We could add a short entry here that the quasi Newton update rules are now initialised upon start of a solver with the new internal function `initialize_update!`; just to keep that complete in case someone want to add a new rule.

src/plans/quasi_newton_plan.jl
```diff
@@ -413,6 +420,11 @@ function (d::QuasiNewtonMatrixDirectionUpdate{T})(
     get_vector!(M, r, p, -d.matrix \ get_coordinates(M, p, X, d.basis), d.basis)
     return r
 end
+function initialize_update!(d::QuasiNewtonMatrixDirectionUpdate)
```
kellertuer, line 423: Would it be reasonable to put the matrix to copy over into the parameters here (positional, with default `I`)?

src/plans/quasi_newton_plan.jl
```diff
@@ -632,22 +657,17 @@ Generate a cautious update for either a matrix based or a limited memory based u
...
-mutable struct QuasiNewtonCautiousDirectionUpdate{U} <:
+mutable struct QuasiNewtonCautiousDirectionUpdate{U,Tθ} <:
                AbstractQuasiNewtonDirectionUpdate where {
-    U<:Union{QuasiNewtonMatrixDirectionUpdate,QuasiNewtonLimitedMemoryDirectionUpdate{T}}
-} where {T<:AbstractQuasiNewtonUpdateRule}
+    U<:Union{QuasiNewtonMatrixDirectionUpdate,QuasiNewtonLimitedMemoryDirectionUpdate}
+}
     update::U
-    θ::Function
+    θ::Tθ
```
kellertuer, line 665: Thanks for spotting, this is definetly a code line from before I was aware how all struct parameters work :)

kellertuer (commented):
LGTM, just a few  minor comments.

src/plans/quasi_newton_plan.jl
kellertuer: (I miss comments that are basically resolved by default but mentioned to provide a short comment)

src/plans/quasi_newton_plan.jl
mateuszbaran: That's somewhat unclear to me. It could probably be used for preconditioning but then maybe we would have to transport it between points upon reinitialization?

src/plans/quasi_newton_plan.jl
mateuszbaran: Non-concrete field types is something I sometimes look for and try to avoid or remove if possible :)

src/plans/quasi_newton_plan.jl
mateuszbaran: Anyway, I'd prefer to do it in a separate PR.

kellertuer: I am not sure what it could be used for, I just thought it might have been nice; but sure we can also keep that for a later PR, when we have a better idea what it could be used for.

Changelog.md
mateuszbaran: Done.

src/plans/quasi_newton_plan.jl
kellertuer: Yes very good; especially in older code of mine – or students of mine – where I (nor them) was not yet aware of that.

kellertuer (approved):

PR #391, opened 2024-06-04
Fix Initial condition on Lanczos

### LINKED ISSUE #390: Lanczos initial value shouldn't be zero
>I've quickly checked this, it looks like a bug. step_solver!(dmp::AbstractManoptProblem{<:TangentSpace}, ls::LanczosState, i) with i == 1 gets a zero-vector in ls.X and we have division by 0. I'm not familiar with this solver so I don't know where this case should be handled.

From https://github.com/SciML/Optimization.jl/pull/763#discussion_r1624513612

kellertuer:
If you start Lacnzos directly in a mnimizer, the first closed form solution yields NaNs. This is now fixed by checking this in the first order condition (which also does not make sense when we are exactly at a minimiser).

This fixes #390.

# 🛣️ Roadmap
* [x] maybe add a test for this.

### DIFF
--- a/src/solvers/Lanczos.jl
+++ b/src/solvers/Lanczos.jl
@@ -255,6 +255,10 @@ function (c::StopWhenFirstOrderProgress)(
     dmp::AbstractManoptProblem{<:TangentSpace}, ls::LanczosState, i::Int
 )
     if (i == 0)
+        if norm(ls.X) == 0
+            c.reason = "The gradient of the gradient is zero."
+            return true
+        end
         c.reason = ""
         return false
     end
--- a/src/solvers/Lanczos.jl
+++ b/src/solvers/Lanczos.jl
@@ -255,7 +255,7 @@ function (c::StopWhenFirstOrderProgress)(
     dmp::AbstractManoptProblem{<:TangentSpace}, ls::LanczosState, i::Int
 )
     if (i == 0)
-        if norm(ls.X) == 0
+        if norm(ls.X) == zero(eltype(ls.X))
             c.reason = "The gradient of the gradient is zero."
             return true
         end
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,7 +5,14 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.4.63] unreleased
+
+## [0.4.63] June 4, 2024
+
+### Changed
+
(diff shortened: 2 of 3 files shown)


src/solvers/Lanczos.jl
```diff
@@ -255,6 +255,10 @@ function (c::StopWhenFirstOrderProgress)(
     dmp::AbstractManoptProblem{<:TangentSpace}, ls::LanczosState, i::Int
 )
     if (i == 0)
+        if norm(ls.X) == 0
```
Vaibhavdixit02, line 258:
```suggestion
        if norm(ls.X) == zero(eltype(ls.X))
```

Do you expect an exact zero btw, this might be more robust with an \approx 0 check

Vaibhavdixit02: Do you expect an exact zero btw, this might be more robust with an \approx 0 check?

kellertuer: Fo now I mean exact zero. We can also introduce a tolerance, but the two cases where that appeared (and I know for both why) it was _exact_ zero.

1. Mateusz example started at a minimiser and the vector was literally `[0.0, 0.0]`.
2. your example had _just_ maxiter as stopping criterion, so it would have continued 10k iterations if you set that. At iteration 17 it ran below numerical precision and hence in tier 18 the norm of the gradient was again exactly 0.

kellertuer: The first I would consider a bug,
the second maybe half-a-bug; really just maxiter as a stopping criterion is not the best of choices.

kellertuer: Even when now trying to write a test, I just managed to get a first gradient of norm `1e-15` and that did not result in `NaN`s. So one really needs an example with _exactly_ zero gradient for Lanczos to fail.

mateuszbaran: > this might be more robust with an \approx 0 check?

This is actually a common misconception: for Float64 `x == 0` and `isapprox(0, x)` are equivalent, see:
```julia
julia> isapprox(0.0, nextfloat(0.0))
false
```
Maybe it would be better to have some positive tolerance but it's not obvious without looking carefully at the algorithm.

kellertuer: I can add more comments, but I was even with an hour of coming up with examples not able to find a gradient that was not _exactly_ zero and were Lanczos failed.
For example the new Test for ARC the gradient at the (numeric/from Eigenvectors) minimiser has only entries of size `1e-16` its norm is of size `1e-15`,
but still Lanczos runs fine and the special case introduced here is not needed (I mean it still stops after one step, but Lanczos is fine with not producing NaNs). So I feel _exact zero_ is the only problem here.

mateuszbaran (approved):

PR #392, opened 2024-06-05
Rework high level interfaces

### LINKED ISSUE #381: Unify passing of optional parts of the objective in high-level solver interfaces
As a continuation of #379, in order to reduce the ambiguities, we should unify how optional elements of the objective are passed.

For example TCG has 7 signatures – the Hessian, the base point p and the start point X are optional, but positional rguemnts, and any of these can be left out.

On the other hand the difference of convex algorithm there are two signatures

```
difference_of_convex_algorithm(M, f, g, ∂h, p=rand(M); kwargs...)
difference_of_convex_algorithm(M, mdco, p; kwargs...)
```

this can be enhanced by for example passing the gradient of `g` (to get the subproblem automatically generated) or the gradient of `f` (to get better debug), but all these are keyword arguments.
This should become the default, to reduce possibilities of ambiguity.

Second, there is often an ambiguity for the case where the points are not arrays but numbers and we provide a wrapper; this happens, when just the cost (or in general only one element like a gradient) is necessary for the objective (like in particle swarm).

And idea to resolve this is more like an internal redesign – we could introduce a keyword argument `iterate_type=:Number` that is set to number if the iterate is number and dispatch on the value internally. That way the dispatch does not happen on the high-level interface. This would even allow users with “non-`Number`-eltypes that are not arrays” to benefit from the idea of “wrapping into mutable arrays” automatically.
Hopefully this second idea can be done non-breaking; the first is breaking for sure.

## Edit

To be precise the ambiguity errors are

```
Ambiguity #1
kwcall(::NamedTuple, ::typeof(Manopt.difference_of_convex_proximal_point), M::ManifoldsBase.AbstractManifold, grad_h, p::Number) @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/difference-of-convex-proximal-point.jl:239
kwcall(::NamedTuple, ::typeof(Manopt.difference_of_convex_proximal_point), M::ManifoldsBase.AbstractManifold, mdcpo::O, p) where O<:Union{AbstractDecoratedManifoldObjective, ManifoldDifferenceOfConvexProximalObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/difference-of-convex-proximal-point.jl:273

Possible fix, define
  kwcall(::NamedTuple, ::typeof(Manopt.difference_of_convex_proximal_point), ::ManifoldsBase.AbstractManifold, ::O, ::Number) where O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.ManifoldDifferenceOfConvexProximalObjective}
```

* [x] Problem: Ob the objective level we basically can not handle a `Number` p, on the other hand we also do not want to type `grad_h`.

```
Ambiguity #2
difference_of_convex_proximal_point(M::ManifoldsBase.AbstractManifold, grad_h, p::Number; evaluation, cost, gradient, g, grad_g, prox_g, kwargs...) @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/difference-of-convex-proximal-point.jl:239
difference_of_convex_proximal_point(M::ManifoldsBase.AbstractManifold, mdcpo::O, p; kwargs...) where O<:Union{AbstractDecoratedManifoldObjective, ManifoldDifferenceOfConvexProximalObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/difference-of-convex-proximal-point.jl:273

Possible fix, define
  difference_of_convex_proximal_point(::ManifoldsBase.AbstractManifold, ::O, ::Number) where O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.ManifoldDifferenceOfConvexProximalObjective}
```

* [x] (the same as ambiguity 1, just for DCPPA instead of DCA)

```
Ambiguity #3
kwcall(::NamedTuple, ::typeof(Manopt.particle_swarm), M::ManifoldsBase.AbstractManifold, f, swarm::AbstractVector{T}) where T<:Number @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/particle_swarm.jl:249
kwcall(::NamedTuple, ::typeof(Manopt.particle_swarm), M::ManifoldsBase.AbstractManifold, mco::O, swarm::AbstractVector) where O<:Union{AbstractDecoratedManifoldObjective, AbstractManifoldCostObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/particle_swarm.jl:264

Possible fix, define
  kwcall(::NamedTuple, ::typeof(Manopt.particle_swarm), ::ManifoldsBase.AbstractManifold, ::O, ::AbstractVector{T}) where {O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.AbstractManifoldCostObjective}, T<:Number}
```

* [x] nearly the same as 1&2, just that here it is `f` being untyped and/or we do not want to specialise a swarm of numbers on a already generated objective.

```
Ambiguity #4
particle_swarm(M::ManifoldsBase.AbstractManifold, f, swarm::AbstractVector{T}; velocity, kwargs...) where T<:Number @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/particle_swarm.jl:249
particle_swarm(M::ManifoldsBase.AbstractManifold, mco::O, swarm::AbstractVector; kwargs...) where O<:Union{AbstractDecoratedManifoldObjective, AbstractManifoldCostObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/particle_swarm.jl:264

Possible fix, define
  particle_swarm(::ManifoldsBase.AbstractManifold, ::O, ::AbstractVector{T}) where {O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.AbstractManifoldCostObjective}, T<:Number}
```

* [x] quite similar to 3

```
Ambiguity #5
kwcall(::NamedTuple, ::typeof(Manopt.stochastic_gradient_descent), M::ManifoldsBase.AbstractManifold, grad_f, p::Number) @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/stochastic_gradient_descent.jl:176
kwcall(::NamedTuple, ::typeof(Manopt.stochastic_gradient_descent), M::ManifoldsBase.AbstractManifold, msgo::O, p) where O<:Union{AbstractDecoratedManifoldObjective, ManifoldStochasticGradientObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/stochastic_gradient_descent.jl:202

Possible fix, define
  kwcall(::NamedTuple, ::typeof(Manopt.stochastic_gradient_descent), ::ManifoldsBase.AbstractManifold, ::O, ::Number) where O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.ManifoldStochasticGradientObjective}
```

* [x] same as 1&2 just for stochastic gradient descent

```
Ambiguity #6
stochastic_gradient_descent(M::ManifoldsBase.AbstractManifold, grad_f, p::Number; cost, evaluation, kwargs...) @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/stochastic_gradient_descent.jl:176
stochastic_gradient_descent(M::ManifoldsBase.AbstractManifold, msgo::O, p; kwargs...) where O<:Union{AbstractDecoratedManifoldObjective, ManifoldStochasticGradientObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/stochastic_gradient_descent.jl:202

Possible fix, define
  stochastic_gradient_descent(::ManifoldsBase.AbstractManifold, ::O, ::Number) where O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.ManifoldStochasticGradientObjective}
```

* [x] same as 5

```
Ambiguity #7
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:451
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), M::ManifoldsBase.AbstractManifold, mho::O, p, X) where O<:Union{AbstractDecoratedManifoldObjective{E, <:AbstractManifoldSubObjective} where E, AbstractManifoldSubObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:534

Possible fix, define
  kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), ::ManifoldsBase.AbstractManifold, ::O, ::Any, ::TH) where {O<:Union{Manopt.AbstractDecoratedManifoldObjective{E, <:Manopt.AbstractManifoldSubObjective} where E, Manopt.AbstractManifoldSubObjective}, TH<:Function}
```

* [x] nearly the same as above just with 2 parameters instead of 1 (`f` and `grad_f` since we also have `p` and `X`).

```
Ambiguity #8
truncated_conjugate_gradient_descent(M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH, p, X; evaluation, preconditioner, kwargs...) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:467
truncated_conjugate_gradient_descent(M::ManifoldsBase.AbstractManifold, F, gradF, x, Y, H::TH; kwargs...) where TH<:Function @ Manopt deprecated.jl:103

Possible fix, define
  truncated_conjugate_gradient_descent(::ManifoldsBase.AbstractManifold, ::Any, ::Any, ::TH, ::Any, ::TH) where {TH<:Function, TH<:Function}
```

* [x] similar to the above just that here the Hessian is typed, since the second is deprecated, this might be resolved in the next breaking change anyways

```
Ambiguity #9
truncated_conjugate_gradient_descent(M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH; kwargs...) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:451
truncated_conjugate_gradient_descent(M::ManifoldsBase.AbstractManifold, mho::O, p, X; kwargs...) where O<:Union{AbstractDecoratedManifoldObjective{E, <:AbstractManifoldSubObjective} where E, AbstractManifoldSubObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:534

Possible fix, define
  truncated_conjugate_gradient_descent(::ManifoldsBase.AbstractManifold, ::O, ::Any, ::TH) where {O<:Union{Manopt.AbstractDecoratedManifoldObjective{E, <:Manopt.AbstractManifoldSubObjective} where E, Manopt.AbstractManifoldSubObjective}, TH<:Function}
```

* [x] this is maybe a more serious one, since it is an interplay of providing the hessian _or_ initial point/vector.

```
Ambiguity #10
truncated_conjugate_gradient_descent(M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH; kwargs...) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:451
truncated_conjugate_gradient_descent(M::ManifoldsBase.AbstractManifold, mho::O, p, X; kwargs...) where O<:Union{AbstractDecoratedManifoldObjective, ManifoldHessianObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:525

Possible fix, define
  truncated_conjugate_gradient_descent(::ManifoldsBase.AbstractManifold, ::O, ::Any, ::TH) where {O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.ManifoldHessianObjective}, TH<:Function}
```

* [x] nearly the same as 9

```
Ambiguity #11
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:451
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), M::ManifoldsBase.AbstractManifold, mho::O, p, X) where O<:Union{AbstractDecoratedManifoldObjective, ManifoldHessianObjective} @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:525

Possible fix, define
  kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), ::ManifoldsBase.AbstractManifold, ::O, ::Any, ::TH) where {O<:Union{Manopt.AbstractDecoratedManifoldObjective, Manopt.ManifoldHessianObjective}, TH<:Function}
```

* [x] nearly the same as 10

```
Ambiguity #12
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH, p, X) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:467
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), M::ManifoldsBase.AbstractManifold, F, gradF, x, Y, H::TH) where TH<:Function @ Manopt deprecated.jl:103

Possible fix, define
  kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent), ::ManifoldsBase.AbstractManifold, ::Any, ::Any, ::TH, ::Any, ::TH) where {TH<:Function, TH<:Function}
```

* [x] also a combination of the previous 3, though a case that shall not be possible (with 2 hessians) like 9.

```
Ambiguity #13
truncated_conjugate_gradient_descent!(M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH, p, X; evaluation, preconditioner, kwargs...) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:591
truncated_conjugate_gradient_descent!(M::ManifoldsBase.AbstractManifold, F::TF, gradF::TG, x, Y, H::TH; kwargs...) where {TF<:Function, TG<:Function, TH<:Function} @ Manopt deprecated.jl:103

Possible fix, define
  truncated_conjugate_gradient_descent!(::ManifoldsBase.AbstractManifold, ::TF, ::TG, ::TH, ::Any, ::TH) where {TF<:Function, TG<:Function, TH<:Function, TH<:Function}
```

* [x] the same as 12.

```
Ambiguity #14
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent!), M::ManifoldsBase.AbstractManifold, f, grad_f, Hess_f::TH, p, X) where TH<:Function @ Manopt ~/Repositories/Julia/Manopt.jl/src/solvers/truncated_conjugate_gradient_descent.jl:591
kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent!), M::ManifoldsBase.AbstractManifold, F::TF, gradF::TG, x, Y, H::TH) where {TF<:Function, TG<:Function, TH<:Function} @ Manopt deprecated.jl:103

Possible fix, define
  kwcall(::NamedTuple, ::typeof(Manopt.truncated_conjugate_gradient_descent!), ::ManifoldsBase.AbstractManifold, ::TF, ::TG, ::TH, ::Any, ::TH) where {TF<:Function, TG<:Function, TH<:Function, TH<:Function}
```

* [x] nearly the same as 13.

### LINKED ISSUE #400: Initialise states based on point type already and not requiring an actual point
Several states and similar require a point to initialise points.
Itmight be a good idea to requite the constructors to also accept just the type (and default the point's type to fill that by default).

# Examples
* in the caches extension the constructor at https://github.com/JuliaManifolds/Manopt.jl/blob/4a52a82122b3054bb4369c3e7e391bb0cd8d197d/ext/ManoptLRUCacheExt.jl#L40C5-L40C6 allocates a point and a tangent vector but only uses their types.

kellertuer:
This is an approach to fix #381 

# 🛣️ Roadmap

* [x] move the “point is a number” case not on the highest level but when creating the replacement functions internally
* [x]  check what's left with TCG
* [x] 📚check that the high-level interfaces in the new format are updated as well
* [x] 🛠️ rename `get/set_manopt_paramater` to just `get/set_parameter` since it is no longer exported – and it makes the code a bit more readable
* [x] 🛠️change the keyword `staibilize=`keyword in Quasi newton to a `project!=` keyword as in the other solvers.
* [x] 🛠️unify naming and also call `update_stopping_criterion!` `set_parameter`, since the dispatch by the first type should be good enough.
* 🛠️ unify sub solver initialisation for solver states: accept both problem and state or accept a single function and the evaluation keyword; make p and X keywords
   * [x] ALM 
   * [x] ARC
   * [x] EPM
   * [x] DoC
   * [x] DCPPA
   * [x] FW
   * [x] PBM
   * [x] CBM
   * [x] TR
* Resolve #400 here as well
  an idea could be to put the semantic explanation with the “easy to use” part (like `AverageGradient()`) and put the details like fields of the structs as well as a reference to the interface into the `AverageGradientState` docs. 
  * [x] gradient update rules
  * step sizes (the following factories should be created)
    * [x] Armijo
    * [x] Constant
    * [x] Decreasing
    * [x] Polyak
    * [x] WolfePowellBinary
    * [x] WolfePowell
    * [x] Nonmonotone
    * [x] AdaptiveWNGradient
  * [x] stopping Criteria (will not be changed): the following have a manifold for defaults
     * `StopWhenChangeLess`
     * `StopWhenGradientChangeLess`
     * so the effort should not be that large, but I have to think about a naming that is easy.
     * ah maybe these are indeed to complicated, because we could not store them in the `StopWhenAny` as factories and “producing” them before is a bit too tough, so I will leave this as is for now.
* [x] 📈 test coverage.
* [x] rename `inequality_constrains` and `equality_constrains` keyword arguments to `inequality_constraints` and `equality_constraints` respectively. This ensures they are correctly passed to the `ConstrainedManifoldObjective` constructor


kellertuer:
The `::Number` ambiguities with types points are all removed.

TCG has some strange ambiguities, that I still have to look at. but all other high-level interfaces do not have ambiguities any longer.
The trick was to slightly redefine the methods that “generated” array-functions for number type data.

kellertuer:
Small update: For now any single of the remaining 6 ambiguities, when trying to fix them, the number only goes further up. I am a bit los on this by now, since trying to _reduce_ ambiguities only _increases_ them (with my limited knowledge).

Some are also due to a deprecated signature so we can maybe leave these out in the end.

kellertuer:
The code with mandatory Hessian is so much cleaner. So this will be part of 0.5.0 – so nothing urgent and I will wait for a few other issues to be resolved and check whether we have other breaking changes we should consider / include then.

kellertuer:
Oh tests might be a bit tricky, because we first have to bump version on ManoptExamples.jl it seems for the tests.

kellertuer:
This is mainly waiting for https://github.com/JuliaManifolds/ManoptExamples.jl/pull/16 and is (again) not urgent. So we can also think a bit about what else “to break” to improve.

kellertuer:
Oh ManoptExamples only works with Julia 1.8, so I would either have to raise the Julia version here as well or run the tests with manoptExamples only on Julia 1.8 but not on 1.6

mateuszbaran:
Maybe we should make ManoptExamples.jl compatible with Julia 1.6? What's the issue?

kellertuer:
That would maybe also be an idea, there was something with range.

mateuszbaran:
Yes, you just need to replace
```julia
range(interval[1], interval[2], n)
```
with
```julia
range(interval[1], interval[2]; length=n)
```

kellertuer:
Ah will check; probably the best way to loosen the constraints to 1.6 again once I fixed that (maybe I was a bit too eager to finish the weak deps). I am travelling tomorrow so I might only find time on Thursday for that then.

kellertuer:
Short status update here: With ManoptExamples 0.1.8 tests should be fine again. I also removed all deprecated keywords, but it seems I missed something on reworking the interfaces somewhere since now the cache (some `StorageRef`?) Fails since it suddenly gets `Floats` where it expects `Vector{Float}`s. No clue why and I thought I had that resolved (so I hope it is not due to removing deprecated stuff).

But well. This might hence still take quite a while I feel. but we also can wait until we have more breaking stuff.

kellertuer:
I think I at least narrowed down the error: now the `p` we store internally is always mutating (that is on circle and positive numbers it is internally also a `[p]` if `p` is a number, but the `point_storage` win the storage/cache gets initialised to `p` somethow, so that is numbers – and then conversion fails.

Since this is now consistent in all high level interfaces, we might have to check a few places where internally this is accidentally not the case (that points are always vectors in structures of Manopt)

src/solvers/interior_point_Newton.jl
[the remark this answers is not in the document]
kellertuer: Locally this does not fail and I do not understand what the problem here shall be?

mateuszbaran:
Since this is a breaking change then maybe it would be a good idea to also tackle #400 here? Especially implementation of the factory pattern.

mateuszbaran:
The adjustments necessary for step sizes don't look that complicated to me, especially if we don't have to make them non-breaking.

kellertuer:
No, also the other ones for the gradient rules should not be that complicated and doable nonbreaking I think.
That was why I thought we could to that after this PR, but I can check (when I have finished the points still open) how much we can address that here as well, sure.

kellertuer:
I looked at it and I agree – in a breaking way it should be easy to do and also much easier than trying to do it non-breaking. Will add it here, sure.
Just that this way the PR might take a bit longer, given that my semester starts next week (and I have resit-exams to administer this week).

mateuszbaran:
Great! I think it is fine to delay a breaking release to resolve that issue with line searches.

kellertuer:
I needed a small trick – the tutorials have to render for this version also on the dev version from the folder, otherwise they would run on 0.4; we could maybe return to normal after the breaking release.

kellertuer:
I fell into a small rabbit hole.

Now we have for states with sulbsolvers the constructors
* `State(M, sub_problem_closed; evaluation=[...], kwargs...)` that wraps the evaluation and calls
* `State(M, sub_problem, sub_state; kwargs...)` which now has a single type allowed for the state.

To make this easier, `p=` was moved to a keyword. So then I did the same with all other states.

But with this, the constructors are very clear now, though it was a bit of a refactoring.
I am thinking about not exporting the states to declare them a bit more internal.

This finishes the main intention of this PR. so only the factory-idea is left here (and code coverage).

kellertuer:
I just worked on a first sketch for the factory idea. I would like some feedback on whether this is what you want @bvdmitri, and from @mateuszbaran whether this is still nice memory wise (I hope so).

The idea is as follows. The factory just wraps the direction update `T` all `args` and `kwargs`; I even added a manifold if one wants to, see

https://github.com/JuliaManifolds/Manopt.jl/blob/3147fb4a77ec64d5c658c45a1edd90e16a7ea50c/src/plans/gradient_plan.jl#L272-L323

(oh the docs are still outdated; the actual type is only a parameter no longer a field)

Then the actual call is either just `factory()` (works if the internal manifold from before is not nothing) or `factory(M)` that then passes all args/kwargs to the actual type.
There is also a `produce_rule` that accepts both rules and factories but makes sure the result is a rule

With that both

```
gradient_descent(SomeManifold(), f, g, p; direction = MomentumGradient())
gradient_descent(SomeManifold(), f, g, p; direction = MomentumGradientRule(M, p))
```

work, but since the factory also accepts the `p` as a positional argument also the old version

```
gradient_descent(SomeManifold(), f, g, p; direction = MomentumGradient(M,p))
```

works and even

```
gradient_descent(SomeManifold(), f, g, p; direction = MomentumGradient(p))
```
(so you can fill any args/kwargs. The last one I like because if you have memory/a point you want to pass you do not longer have to think of the manifold either.

besides that, I feel since `p` can always be filled with `rand(M)` I might move that even to a keyword argument in the (more internal) `XRule` types. The current X` types will then always create such a factory.

Let me know

a) what you think (before I rework all other rules and
b) wherelse we could provide such a factory pattern

If the factory is ok in its form, it might even work for nearly any other “thing” in Manopt, but my first idea would be

* `Stepsize`s
* `StoppingCrtierion`

maybe also Debug/Record things, but since they can already usually addressed with symbols (cf. https://manoptjl.org/stable/plans/debug/#Manopt.DebugActionFactory-Tuple{Symbol}), maybe they are already fine.

bvdmitri:
Nice work, I’m currently on vacation, but will be able to provide some feedback next week when return to the office :)

kellertuer:
Cool, enjoy your vacation!

kellertuer:
Since the fix was slightly breaking, I also added the fix for the remaining 2 points in #382 here. Now both the full matrix and the limited memory variant are equivalent in how they perform their updates.

src/plans/debug.jl
```diff
@@ -1171,20 +1164,26 @@ one are called with an `i=0` for reset.
...
+   ```
+   [:Iterate, " | ", :Cost, :Stop, 10]
+   ```
 
-Adds a group to :Iteration of three actions ([`DebugIteration`](@ref), [`DebugDivider`](@ref)`(" | "),  and[`DebugCost`](@ref))
-as a [`DebugGroup`](@ref) inside an [`DebugEvery`](@ref) to only be executed every 10th iteration.
-It also adds the [`DebugStoppingCriterion`](@ref) to the `:EndAlgorhtm` entry of the dictionary.
+   Adds a group to :Iteration of three actions ([`DebugIteration`](@ref), [`DebugDivider`](@ref)`(" | "),  and[`DebugCost`](@ref))
+   as a [`DebugGroup`](@ref) inside an [`DebugEvery`](@ref) to only be executed every 10th iteration.
+   It also adds the [`DebugStoppingCriterion`](@ref) to the `:EndAlgorhtm` entry of the dictionary.
```
bvdmitri, line 1173: ?
```suggestion
   It also adds the [`DebugStoppingCriterion`](@ref) to the `:EndAlgorithm` entry of the dictionary.
```

src/plans/gradient_plan.jl
```diff
@@ -276,20 +269,75 @@ individual one that provides these values.
...
+"""
+    DirectionUpdateRuleFactory{M,T,A,K} <: DirectionUpdateRule
+
+A generic factory to create direction update rules.
+Motivated by the fact that in some situations it is nicer to not being forced
+to specify a manifold, this factory allows to postpone that.
+It still stores all parameters one wants to provide, even those that would
+require the manifold to be filled with a reasonable default.
+Any optional parameter or keyword argument can hence still be overwritten.
+Similarly you can also pass a msnifold to the constructor, in case you
```
bvdmitri, line 281:
```suggestion
Similarly you can also pass a manifold to the constructor, in case you
```

src/plans/gradient_plan.jl
```diff
@@ -276,20 +269,75 @@ individual one that provides these values.
...
+
 """
     IdentityUpdateRule <: DirectionUpdateRule
 
 The default gradient direction update is the identity, usually it just evaluates the gradient.
+
+Use `Gradient()` to create the corresponding factory.
 """
 struct IdentityUpdateRule <: DirectionUpdateRule end
+IdentityUpdateRule(M) = IdentityUpdateRule() # Temp hack for the mockup
```
bvdmitri, line 335: This looks quite fishy, maybe its required, but maybe the comment should explain a bit more why its needed or what is needed to remove the hack

src/plans/gradient_plan.jl
```diff
@@ -350,7 +401,36 @@ function (mg::MomentumGradient)(
...
+* `direction=`[`IdentityUpdateRule`](@ref) preprocess the actual gradient before adding momentum
+* $(_kw_X_default)`
+* `momentum=0.2` amount of momentum to use
+* $(_kw_vector_transport_method_default): $(_kw_vector_transport_method)
+
+!!! info
+    This function generates a [`DirectionUpdateRuleFactory`](@ref) for this
+    for the [`MomentumGradientRule`](@ref). If you do not provide the manifold
+    or any other of the keywords that require the manifold for a default,
+    you can provide a manifold later to use that manifold's default's instead.
```
bvdmitri, line 427: Here the message says "...for this for the..." which is a bit broken English. It also repeats in all other docstings, maybe worth to make a variable for this note and interpolate it in the same way as e.g. `$(_kw_p_default)`

src/plans/gradient_plan.jl
```diff
@@ -376,104 +457,114 @@ them to the current iterates tangent space.
...
+
+@doc raw"""
+    NesterovRule <: DirectionUpdateRule
+
+Compute a Nesterov inspired direction update rule.
+See [`Nesterov`](@ref) for details
+
+# Fields
+
+* `γ`
```
bvdmitri, line 537: missing description after refactoring?

bvdmitri (commented):
Thats an amazing amount of work, I've checked some of the changes, but the PR looks big so I skipped some parts (especially related to the rename of setting the parameters). The factory pattern looks very nice and as long as tests are passing you should be good to go :) when PR is ready to merge we could also test our codebase in RxInfer with the new changes to see if there are some hidden bugs or regressions.

kellertuer:
Sure, I was just asking about feedback on the factory, the rest is quite a bit of rework because we noticed Aqua was not too happy with a few ambiguities.
The reason both are on this PR is, since that will be the next (breaking) version that. So no worries, I was not asking you to review the rest – just the idea of the factory.

Nice. that you like it, so I will continue introducing that to the other areas as well and also use that to review the docs.

src/plans/gradient_plan.jl
kellertuer: this is also one of the things I do not like too much with the factory: The factory for now _always_ “pushes” the manifold towards the `DirectionUpdateRule`

I did not yet have a goo idea what to do here. Maybe one could add a parameter to the factory, whether “on production” it hast to pass the manifold or not.
the other idea is of course to say: Any update rule _always_ has to have a constructor with at least `M::AbstractManifold` as first argument as per agreement, then this line would just have to be taped on M.

I will think about that which I like better (factory parameter or mandatory arguments in rule constructors).

src/plans/gradient_plan.jl
kellertuer: Oh yes, I must have missed this.

In general since we now have 2 doc strings (one for the `XRule` one for `X`) I will split most docs into: The math part and how to use it will be in `X`, the technical details will be in the rule then.

kellertuer:
Concerning hidden bugs, the Average gradient had one, where in some cases the history was not filled correctly and hence it was much closer to gradient descent than averaging. This already has been fixed.

For the rest, sure I'll let you know all the refactoring to the new factory has been finished; probably the main work is even the docs and not the code itself.

src/plans/gradient_plan.jl
kellertuer: I fixed this now when reworking the factory to be more general (also working for other elements in Manopt later) and added a `requires_manifold` argument, so this line will be removed :)

kellertuer:
@bvdmitri thanks for the feedback, I addressed the structural things already; I generalised the factory also in name to a `ManifoldDefaultsFactory` so it can be used with other things as well.
The docs are not yet perfect, but I also have to work through the remaining direction rules, stopping Criteria, and Stepsizes. and will do that during that as well.

An idea is that there is usually an internal type (like the `XRule`s now) that documents fields and internal parts and is also not exported. Compared to that the new `X` (exported) will document the mathematical part and have the new Note to the internal one with the factory.

Since this involves a few types to rework, this will still take a while, but I plan to maybe refactor a type per day (besides teaching that started today again).

bvdmitri:
Great amount work @kellertuer ! nicely done, the documentation is indeed important, but the implementation is solid IMO

kellertuer:
Thanks. As I said I will then work through the changes for some time, but I feel it improves usability quite a bit :)

kellertuer:
Small update, all direction rules should now work fine with the factory, that is one can now always write

`conjugate_gradient_descent(M, f, grad_f; coefficient=FletcherReevesCoefficient())`


but one can still overwrite any argument of keyword argument when calling these (still without using a manifold) for example the threshold here can be set `factory = ConjugateGradientBealeRestart(; threshold=0.2)` – but its vector transport it would need is still only filled with a default later when one passes a manifold to the `factory(M)` this call creates. This way one can easily still overwrite keywords and even arguments to coefficients or also for the

`AverageGradient(; n=20)`

this still works fine for all keywords of any of these rules. And test coverage should also by now work.
So the only thing on this PR left is to check whether it is reasonable to do this for step sizes (probably) and stopping criteria (not 100% sure, not many need a manifold I think), too.
So that we have `Armijo()` as a factory for `ArmijoLinesearch` and such; the factory is already generic enough for this, it's mainly a matter of slight adaption of code and a bit larger adaption of documentation.

kellertuer:
This now maybe also closes #406, #407, #408 (all Manifolds bumps for the package, docs and tutorials).

kellertuer:
@mateuszbaran I do not seem able here to get RecursiveArrayTools (now a weak dependency here as well) to work fine with Statistics. Any ideas?

Today was rally not my day. My progress over several hours can bee strictly upper bounded by zero.

mateuszbaran:
Ah, I see you've figured it out already.

kellertuer:
My main problem is, I neither know what I actually did wrong nor how I fixed it (in basically total confusion). But it works now and for this PR I am even only 2 stepsize rules aways from finishing all rework.

kellertuer:
I think this is good to go (was a bit more work than I thought) – and turned out quite nice I think

@bvdmitri could you maybe check whether the new docs that provide the new interface are understandable? These are
* Step size factories at https://manoptjl.org/previews/PR392/plans/stepsize/#Manopt.ArmijoLinesearch (and the following ones); I left the math now in those, and moved the internal fields and technical descriptions to the corresponding `XStepsize` structs below, see https://manoptjl.org/previews/PR392/plans/stepsize/#Manopt.AdaptiveWNGradientStepsize and following
* Similarly the gradient update rules at https://manoptjl.org/previews/PR392/solvers/gradient_descent/#Manopt.AverageGradient and following have also been updated
* ...and also all conjugate gradient descent coefficients at https://manoptjl.org/previews/PR392/solvers/conjugate_gradient_descent/#cg-coeffs

If you have time, maybe give this new version a test run – if not, that is of course also fine

@mateuszbaran if you have time – the high level interfaces have now been reworked, so there are no warnings from Aqua any longer, especially no ambiguities, start points `p` are set to `rand` as default last positional argument whenever reasonable. Throughout I hope this is now consistent – and at least Aqua is happy :)

I would like to merge this maybe towards the end of the week, so if you have time to comment on these changes, that would be great – I think this still works as intended; I hope I have documented all breaking changes. Overall this breaking change is mainly unification of documentation and keyword names and improvements on Usability, since `M` is less often needed.

bvdmitri:
I noticed some bad renders in the docs, but otherwise seems good to me :) 
I also noticed that math doesn't render nicely very often in many docstrings, not sure why is this the case...
Screenshot 2024-08-27 at 08 18 21
Screenshot 2024-08-27 at 08 19 54
Screenshot 2024-08-27 at 08 21 18
Screenshot 2024-08-27 at 08 21 49

Here the English is broken IMO
Screenshot 2024-08-27 at 08 19 16

How about 
```
This function generates a `ManifoldDefaultsFactory`for `ConstantStepsize`.
The manifold specification is optional; if not provided, the manifold will be automatically determined from the optimization procedure. This affects all arguments and keyword arguments with defaults that depend on the manifold, unless explicitly specified here.
```

Also sometimes its inconsistent as here

Screenshot 2024-08-27 at 08 28 13

(also in other places, e.g. `Polyak` and `PolyakStepsize`)

But just let me reiterate that this is a great amount of work you did and its quite impressive!

kellertuer:
Ah, the info box is a function (with that type as a variable, exactly because I want to just fix the English in one place ;) and maybe I set that in copy. paste not correctly. Will check – also your other findings; thanks for checking!

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: ["1.6", "1.9", "1.10"]
+        julia-version: ["1.8", "1.9", "1.10"]
```
mateuszbaran, line 14:
```suggestion
        julia-version: ["1.6", "1.9", "1.10"]
```
Wasn't the issue with Julia 1.6 resolved already?

kellertuer: I don't remember, let's ask CI ;)

kellertuer: Indeed! Thanks for noticing/reminding, and great to be back on 1.6 even.

kellertuer:
I might consider in a later PR to also introduce the factory at least to some stopping criteria (`StopWhenChangeLess` or `StopWhenGradientChangeLess` or such), but that would also require `StopWhenAny` and `StopWhenAll` to be able to handle those, so it is maybe a bit more work – should be doable nonbreaking still I think.

thanks again for all the ideas and contributions!

PR #393, opened 2024-06-07
Rework presentation of keywords

kellertuer:
This is heavily WIP (just had a bit on time on the train with not so much concentration).

This aims to

* find a format for keyword arguments that is better readable in both the rendered docs _and_ REPL
* -> for not I have a newline too much in the HTML format (hence rendering it here might be nice)
*  indicate defaults for keywords is for now done with an emoji in the last line (a bit too much offset for now).

# 🛣️ Roadmap
* [x] find a better way of presenting keyword arguments
* [x] start collecting common keywords and their defaults in a central “string collection” to reduce duplication in strings (and hence avoid possible multiple typos / indifferences when updating parts)
* rework solver interfaces
  * [x] `plans/`
  * `solvers/`
    * [x] ARC and its sub solvers, especially Lanczos
    * [x] ALM
    * [x] Chambolle-Pock
    * [x] CG
    * [x] CBM
    * [x] CPPA
    * [x] DoC
    * [x] DCPPA
    * [x] DR
    * [x] EPM
    * [x] `Frank_Wolfe_method` and its state
    * [x] `gradient_descent` and its state
    * [x] LM
    * [x] NM
    * [x] PSO
    * [x] PDRSNA
    * [x] PBM
    * [x] QN
    * [x] TCG
    * [x] SGM
    * [x] Stochastic Gradient
    * [x] TR
    * [x] conjugate residual (and its plan)
    * [x] interior point newton (and its plan)
* generally
   * [x] work through replaces `` `:(\s+)\(`(.*)`\)`` to ``=$2`:`` (maybe with a newline)
   * [x] similarly for `` `:(\s+)\([`(.*)`\)` `` (keyword with linking examples
   * Introduce Documenterinterlinks for the functions (after most doc strings are no longer raw)
     * [x] `default_retraction_method`
     * [x] `default_inverse_retraction_method`
     * [x] `zero_vector`
     * [x] `rand`
     * [x] `at_iteration` field should be unified
   * unify use of indices
      * [x] `k` for iterations
      * [x] `I` for inequality constraints `g_i`
      * [x] `j` for equality constraints `h_j`
   * [x] carefully check to add more links e.g. for 
     * [x] `get_gradient`
     * [x] `get_cost`
     * [x] `get_iterate`
   * [x] extend the Nelder-Mead documentation and write down the whole algorithm in the doc string.
   * [x] check and fix why it seems that `:WhenActive` is broken in https://manoptjl.org/stable/tutorials/HowToDebug/#Subsolver-debug


kellertuer:
I think I found a good format by now and with the snippets we also avoid “stringduplication” – that is for example the “first parameter is a manifold” text will soon be the same in all doc strings.

kellertuer:
I think I have converged to a nice new format, and with a bit of rewriting how the doc string is defined, this allows to reuse parts as well – for example prominent keywords like `evaluation=`

## Format

```
* `keyword=default`:
  description
```

is nice, since the `=` also indicates that what follows is the default.
All doc strings will be reformulated as “non-`raw`” to allow for interpolation. this means raw parts, like formulae, should be defined before.

Now that this is clear, this PR is really just a large work to check all doc strings and reform them. But the benefit is, that this reduces a lot of duplication in the strings as well, since common tens are now defined in `plans/docstring_snippets.jl`

Still a bit of work to do, where I will try to work on a doc string every day.

kellertuer:
So now this should be good to go. Not sure it is worth its own version, since I will not activate `vale` (basically text style check) for now, since they have too often breaking changes and though I implemented Julia code stuff for them the newest version reports, .jl files are not allowed in checks.

kellertuer:
Since this is really just involving doc string rewrites, I will merge this now and include it in the next release (the other PR still open). It will hence already reside on master for a while then (without a new version).

PR #395, opened 2024-06-10
Refactor get_reason

### LINKED ISSUE #389: Built-in stopping criterions allocate new error strings on each check, even when not required
During the optimization of some of my code, I noticed that the built-in stopping criteria in Manopt are less efficient than they could be. Specifically, they allocate memory every time they are called.

image

Notice the very large portion of allocations is being spend inside the `string` and `print_to_string` functions. And, no, I don't do any optimizations in the string manifold 😅 . The rest of the allocations are most probably related to my code.

Interestingly, the allocation profile shows that most of my allocations are from the error messages generated by the stopping criteria, as seen [here](https://github.com/JuliaManifolds/Manopt.jl/blob/e97d0c0b105feb365ab4450a7ce1f5919cdbfc6c/src/plans/stopping_criterion.jl#L151). The issue is that these error messages are allocated every time, even when they are not needed. This occurs because of the way Julia's string interpolation works.

I believe replacing:
```julia
c.reason = "The algorithm reached its maximal number of iterations ($(c.maxIter)).\n"
```
with
```julia
c.reason = lazy"The algorithm reached its maximal number of iterations ($(c.maxIter)).\n"
```
in all the places will address this inefficiency.

P.S. Great package, by the way!

kellertuer:
This resolves #389.
I had a bit of a boring train ride today and worked through all stopping criteria, removed `.reason` fields and when necessary introduced values to store obtained values being printed when generating the string.

There are two small things left

* [x] for the Lagrange Multiplier criterion in the bundle methods the symbols should maybe be strings since they might also be `-\xi` (and not just `\xi` like now)
* [x] somehow I have to check what happens with `JuMP` their 3 checks for reasons return now empty strings and I do no directly see why


src/plans/stopping_criterion.jl
```diff
@@ -1070,3 +1124,12 @@ end
 function update_stopping_criterion!(c::StoppingCriterion, ::Val, v)
     return c
 end
+
+@doc raw"""
+    get_reason(s)
```
mateuszbaran, line 1129:
```suggestion
    get_reason(s::AbstractManoptSolverState)
```

src/plans/bundle_plan.jl
```diff
@@ -100,32 +100,56 @@ but a vector of two values is required for the `:both` mode.
...
 """
-mutable struct StopWhenLagrangeMultiplierLess{T<:Real,A<:AbstractVector{<:T}} <:
-               StoppingCriterion
-    tolerance::A
+mutable struct StopWhenLagrangeMultiplierLess{
+    T<:Real,A<:AbstractVector{<:T},B<:Union{Nothing,<:AbstractVector{<:Symbol}}
+} <: StoppingCriterion
+    tolerances::A
+    values::A
+    sumbols::B
```
mateuszbaran, line 108:
```suggestion
    symbols::B
```

mateuszbaran:
> * for the Lagrange Multiplier criterion in the bundle methods the symbols should maybe be strings since they might also be `-\xi` (and not just `\xi` like now)

That can be saved in a separate field in the criterion but leaving the string in this one place would be fine too.

src/solvers/Lanczos.jl
```diff
@@ -244,22 +244,23 @@ m(X_k) \leq m(0)
...
+* `at_iteration`: indicates at which iteration (including `i=0`) the stopping criterion
+  was fulfilled and is `-1` while it is not fulfilled.
 """
-mutable struct StopWhenFirstOrderProgress <: StoppingCriterion
-    θ::Float64 #θ
-    reason::String
-    StopWhenFirstOrderProgress(θ::Float64) = new(θ, "")
+mutable struct StopWhenFirstOrderProgress{F} <: StoppingCriterion
+    θ::F
+    at_iteration::Integer
```
mateuszbaran, line 252:
```suggestion
    at_iteration::Int
```

src/plans/stopping_criterion.jl
```diff
@@ -825,20 +884,26 @@ reasons.
 """
 mutable struct StopWhenAll{TCriteria<:Tuple} <: StoppingCriterionSet
     criteria::TCriteria
-    reason::String
-    StopWhenAll(c::Vector{StoppingCriterion}) = new{typeof(tuple(c...))}(tuple(c...), "")
-    StopWhenAll(c...) = new{typeof(c)}(c, "")
+    at_iteration::Integer
```
mateuszbaran, line 887:
```suggestion
    at_iteration::Int
```

kellertuer:
Oh, I missed one solver, in CMA-ES, which has quite a few stopping criteria, they always returned a reason even if they were not active.

For now I think I did rewrite them, but testing them is still missing.
We might take the chance and change the `is_active` fields in these and change them to `at_iteration` in order to provide the iteration number in the reason (at which the stopping criterion indicated to stop).

mateuszbaran (approved):

mateuszbaran:
> We might take the chance and change the `is_active` fields in these and change them to `at_iteration` in order to provide the iteration number in the reason (at which the stopping criterion indicated to stop).

Yes, I think changing it to `at_iteration` would make sense.

kellertuer:
> > We might take the chance and change the `is_active` fields in these and change them to `at_iteration` in order to provide the iteration number in the reason (at which the stopping criterion indicated to stop).
> 
> Yes, I think changing it to `at_iteration` would make sense.

Did that. Also tried to use the same tests for `get_reason` as for others and noticed that on construction, the `StopWhenEvolutionStagnates`s show method errors (not affecting tests for now, but on REPL it did); since one of the means to compute is (on construction) over an empty array.

kellertuer:
Hopefully also fixed the last error I mentioned above, which was mainly that the median of an empty vector does not exist, so I introduced a check and tested that as well.

mateuszbaran (approved):

Issue #397, opened 2024-06-27
typos


spaette:
Was `dictionary's` the intention here?

```
$ sed -n '793p' Manopt.jl/src/plans/record.jl
is called for the `Vector` and the result is added to the debug dictionaries entry
$ sed -n '1078p' Manopt.jl/src/plans/debug.jl
is called for the `Vector` and the result is added to the debug dictonaries entry
$
```

Florida
Intelligence
Variant
address
algorithm
criterion
deactivate
documentation
evaluation
generations
generations
otherwise
otherwise
symbol
than
wrapped

```
$ grep -n Flordia Manopt.jl/docs/src/references.bib
472:    SCHOOL    = {Flordia State University},
$ grep -n INtelligence Manopt.jl/docs/src/references.bib
220:    BOOKTITLE = {7th International Conference on Swarm INtelligence}
$ grep -nr Varant Manopt.jl
Manopt.jl/src/solvers/difference-of-convex-proximal-point.jl:434:    Varant I: allocating closed form of the prox
Manopt.jl/src/solvers/difference-of-convex-proximal-point.jl:453:    Varant II: in-place closed form of the prox
Manopt.jl/src/solvers/difference-of-convex-proximal-point.jl:471:    Varant III: subsolver variant of the prox
$ grep -n adress Manopt.jl/.vale.ini
47:Google.We = false # For tutorials we want to adress the user directly.
$ grep -nr "start of the algorith" Manopt.jl/tutorials/HowToDebug.qmd
79:* `:Start` to print something at the start of the algorith. At this place all other (the following) places are “reset”, by triggering each of them with an iteration number `0`
$ grep -n crtierion Manopt.jl/tutorials/ImplementOwnManifold.qmd
186:    stopping_criterion = StopAfterIteration(10),  # A simple stopping crtierion
$ grep -n dactivate Manopt.jl/tutorials/ImplementOwnManifold.qmd
50:# to keep the output and usage simple let's dactivate tutorial mode here
$ grep -n documenation Manopt.jl/docs/make.jl
10:Render the `Manopt.jl` documenation with optional arguments
$ grep -nr evalutation Manopt.jl
Manopt.jl/docs/src/tutorials/GeodesicRegression.md:157:`evalutation` type accordingly.
Manopt.jl/tutorials/GeodesicRegression.qmd:191:`evalutation` type accordingly.
$ grep -n generatiosn Manopt.jl/src/solvers/cma_es.jl
[...]

PR #398, opened 2024-06-28
fix typos

spaette:
Closes https://github.com/JuliaManifolds/Manopt.jl/issues/397

src/plans/{debug.jl,record.jl}

awaiting clarification on whether the above files

should read `debug dictionaries entry` or `debug dictionary's entry`

docs/src/references.bib

skipping any modifications of the .bib file


kellertuer:
Thanks. Sure we can leave the bib for a future PR.

PR #399, opened 2024-06-28
Kellertuer/riemannian interior point

kellertuer:
This is my (third) summer project: This is initial code for a Riemannian Interior Point Newton method based on the master thesis of @mstokkenes, where of course documentation and testing are not part of master theses, but exploring how this algorithms works and illustrating that is works.

# 🛣️ Roadmap

* [ ] 🛠️ refactor code to `VectorialHessianFunction`s, which are newer than the original code base
* [ ] 📚extend all new documentation strings
* [ ] 📝 Maybe write/extend the tutorial on constrained optimization.
* [ ] 📈 ensure proper test coverage


kellertuer:
Uff, this turns out to be much more tricky than I thought. Somewhere in the KKTVectorFieldNormSq (used in the LInesearch) is a bug (the gradient on the large product manifold is not correct), but I am nearly out of ideas as to why (besides computing a complete example by hand). Spent many hours on that already.

So this sadly might still take quite a while.

kellertuer:
@Mateusz just if you have time, do you want to take a look?
I now even added a 1D-Euclidean example where the gradient of the merit function (=the one used for line search) produces a wrong gradient (see check at the end), but anything I computed by hand indicates that all formula are correct, so I am a bit out of ideas (after about 10 hours of checking handwritten computations against code).

Even found a small bug yesterday during that 10 hours, but it seems that was not the only one.
(The only other thing to check is the updates of sigma and rho, but the main thing is to find why the IPNewton does not do the right thing)

I currently suspect
* The linear system being wrong ( `CondensedKKTVectorFieldJacobian` (say A) is the matrix, `CondensedKKTVectorField` (say b) is minus the right hand side, so what we solve is AX+b = 0
* gradient checks of the KKTVEctorFieldNormSq (and its gradient) fail.

All formulae are documented both close to a style from the paper of Lai&Yoshise (adapted variable names) and one nicer to read (only docs missing: Parameters of IPN).
I personally am slowly turning both blind and dumb on this code, I just do not see through it anymore.

mateuszbaran:
Sure, I will take a look. I've tried running `test_interior_point_newton_simple.jl` (I think this is the right file?) and there are multiple errors but the last `
check_gradient(N, F, grad_F, q, qY; plot=true, error=:info)` actually runs fine, with the exception that the gradient is wrong.

kellertuer:
Well, that is the simple on (x°2 on the real line with constraint x >= 1) I can quickly check, I had hoped I always copied over the last versions I checked with. Yeah my main worry for now is the F (KKTVEctorFieldSquared – or merit function in the paper)

mateuszbaran:
I think I've fixed it by multiplying by 2?
image

kellertuer:
oh :D let me also check the larger example then. As I said, I have checked by now so many things by hand that I was sure, I am not able to see the last trick I am missing (but found two small bugs today still).

kellertuer:
The larger example (still not too large but at least Riemannian `test_interior_point_Newton.jl` from my student still has a wrong gradient at least after 5 steps, But I feel it can not be off by far if now the Euclidean one is correct

mateuszbaran:
I can give you a hint that the error is in the first component of the gradient:
```julia
julia> ForwardDiff.gradient(k -> F(N, k), q)
([-5.578785081155688, 0.8210125930525509, -1.1470345086837803], [3.80509088710519, 1.195579403962125, 2.567637759181358], Float64[], [3.850822689745532, 2.5016182248665264, 2.4737559598127277])

julia> grad_F(N, q)
([-5.569136798420136, 1.0900342547693727, -0.8725353489543635], [3.80509088710519, 1.195579403962125, 2.567637759181358], Float64[], [3.850822689745532, 2.5016182248665264, 2.4737559598127277])
```

kellertuer:
The Large or the small example? That looks more like the larger one on the sphere – are you sure forward diff then provides a Riemannian gradient?

mateuszbaran:
The large one, around line 80 in test_interior_point_Newton.jl. ForwardDiff's gradient passes the `check_gradient` at least.
image

kellertuer:
I am a bit surprised by that since that reads like the euclidean gradient but that component is actually on the sphere (there other 3 are on parameters and hence Euclidean.

I was hoping that is correct. but I will start tomorrow checking f and the components of g as well as F and its gradient again. Sadly grad_F is one of the most involved formulae (“fought” through the formula of the condensed KKT and its Jacobian today).

mateuszbaran:
Actually that also surprised me after some thinking and I've been digging further. When the first component is projected on the sphere to make it Riemannian, actually the same result as `KKTVectorFieldNormSqGradient` is returned and `check_gradient` of course fails.

kellertuer:
Thanks for checking, that seems quite strange, since the norm of the kit vector field is always and only norms of tangents so I am surprised that turns out to be Euclidean correct? For now I do not understand what that means. Maybe that one of the Hessians is wrong (f or g)?

mateuszbaran:
Ah, I've found it. The _check_ is wrong: the vector `X` is direction of which we check correctness of gradient is not a tangent vector.

kellertuer:
Oh, hehe! As I said in the first post, this is quite some lines of code, so I lost a bit the overview after these weeks working on that. thanks for finding that.

mateuszbaran:
You're welcome :slightly_smiling_face: . Sometimes looking at something from outside helps with finding simple issues.

kellertuer:
Nice! Now the gradient is correct _and_ the algorithm stays feasible for the first 400 iterations. So hopefully an Armijo and checking the rho/sigma updates fix the rest

kellertuer:
Yes, I usually try for quite a while myself and was sure, this can only be a strange bug. It was ;)

So, it even converges (starting on the boundary on the right at [0,1,1] (normed back) and the minimiser at north it only takes 5k iterations

Screenshot 2024-07-29 at 21 50 10

to reach that by now. since it does reach the point I would assume the main remaining problem is updating the parameters (barrier and such).

kellertuer:
Thanks for the help yesterday, I think I am now really just missing to improve the stopping criterion (probably by implementing the KKT residual) – it now seems to work quite well (besides stopping a bit late).

kellertuer:
For the example from above, I am surprised we need a very large KKTResidual threshold only to get to this after 65 iterations – but it looks like it now works.

Screenshot 2024-07-30 at 11 52 53

The main thing missing yesterday was, that in the linesearch there is a $\tau$ parameter in the centrality conditions (not yet fully documented) that should be updated during the iterations. That gave the speedup.

So besides documentation (writing the rest and reading everything carefully), this only needs test coverage. Happy to have this working finally.

src/Manopt.jl
```diff
@@ -149,6 +149,25 @@ using Requires
...
+Rn(args...; kwargs...) = Rn(Val(Rn_default()), args...; kwargs...)
+@doc """
+   Rn_default()
+
+Specify a default value to dispatch [`Rn`](@ref) on.
+This default is set to `Manifolds`, indicating, that when this package is loded,
+it is the preferred package to ask for a Euclidean space.
+"""
+Rn_default() = :Manifolds
+Rn(::Val{T}, args...; kwargs...) where {T} = DefaultManifold(args...; kwargs...)
```
mateuszbaran, line 169: Wouldn't it be more simple to just have `Euclidean` in ManifoldsBase.jl? Or just use `DefaultManifold` regardless of whether Manifolds.jl is loaded?

src/plans/conjugate_residual_plan.jl
```diff
@@ -0,0 +1,335 @@
...
+
+```math
+f(X) = \frac{1}{2} \lVert \mathcal A[X] + b \rVert_p^2,\qquad X ∈ T_p\mathcal M,
+```
+
+defined on the tangent space ``T_p\mathcal M`` at ``p`` on the manifold ``\mathcal M``.
+
+In other words this is an objective to solve ``\mathcal A(p)[X] = -b(p)``
+for some linear symmetric operator and a vector function.
+Note the minus on the right hand side, which makes this objective especially taylored
```
mateuszbaran, line 17:
```suggestion
Note the minus on the right hand side, which makes this objective especially tailored
```

src/Manopt.jl
kellertuer: Well we discussed the Euclidean part lengthy recently ;)
Another argument against moving `Euclidean` is, that it would split the implementation into two (the basic functions in base the newer / developing ones in Manifolds).

I would keep this function internal here, and it allows mainly a nicer display (Euclidean) when manifolds is loaded, while the algorithm still works if it its not loaded.
Actually this is the first algorithm that does need these product manifolds with Euclidean (3 times even).

Why do you not like this idea?

src/plans/conjugate_residual_plan.jl
kellertuer: I do Taylor expansions too much, there you ta_y_lor a function ;)

src/Manopt.jl
mateuszbaran: We did, and this right here is IMO the perfect example why `Euclidean` in ManifoldsBase would be useful -- something the recent discussion was somewhat lacking.

Anyway, if you think this is the right solution, maybe we should have this function in ManifoldsBase and exported? And if you don't think this is the right solution, then you have your question answered.

kellertuer: I think this it the right solution.

We could call it differently, when we move this to Base (maybe even `Euclidean` – ah that might not work when it should be overridden by Manifolds – maybe `vector_space(n)`?).

kellertuer: > I'm not necessarily for having Euclidean in ManifoldsBase but it would be nice to have an official solution for this exact problem.

You are right a function like `vector_space(n)` is probably a very good tradeoff, that allows to access Euclidean (or DefaultManifold) from ManifoldsBase.
The more I see that, the more I like `vector_space(n)` for this.

mateuszbaran: `vector_space` sounds like a good name (definitely better than `Rn`) but for consistency I think we can't have it return `DefaultManifold` from ManifoldsBase.jl by default. We can only tell users that if they need to call it without loading `Manifolds.jl`, they have to implement it themselves (possibly as returning `DefaultManifold`).

kellertuer: Can you elaborate as to why? I like in my solution here, that it does exactly that.
It always provides a vector space, `DefaultManifold` if nothing more precise given, `Euclidean` if available. That's what the `Rn_default` is for.

mateuszbaran: Mostly because it makes the value returned by `vector_space(n)` dependent on module load order, which can lead to surprising results where seemingly trivial changes change the type you get. It may have really weird side effects if for some reason IPM is used both before and after loading Manifolds.jl.

IMO the minor convenience is not worth trouble it can cause.

mateuszbaran: Either `DefaultManifold` is a good enough result of `vector_space` and there is not point in changing it to `Euclidean`, or it is not and should not be returned anyway.

kellertuer: Hm, but then there is no need to introduce `vector_spacee(n)` basically? If `DefaultManifold` is a good enough default we Cana also just write that anywhere. 
I functionality I even agree, it is. The main reason I did this here is that `Euclidean` looks nicer in prints.

You are right that there is a changing behaviour when loading a package – that might be misleading / confusing to a user. I would be fine with the convenience over the trouble, but I totally see that is a matter of taste.

We could turn the `Rn_default` into a `ManifoldsBase.DEFAULT_VECTORSPACE` setting or so? With my persona preference I could still set that to `:Euclidean` (or `:Manifolds`, depending on whether we consider that a manifold or package indicator). Though I am not sure a variable or a function is the best choice for that, this would allow to default it to `:ManifoldsBase`(or `:DefaultManifold`) to keep your preference by default.
On the other hand the bad-kellertuer-preference in Manopt.j could be to change that to `:Manifolds` ( or `:Euclidean`) for my convenience (and the users confusion 🤓)

mateuszbaran: > Hm, but then there is no need to introduce `vector_spacee(n)` basically? If `DefaultManifold` is a good enough default we Cana also just write that anywhere.
> I functionality I even agree, it is. The main reason I did this here is that `Euclidean` looks nicer in prints.

Yes, if `vector_space` returns `DefaultManifold`, then there is not much reason for it to exist. IMO it only makes sense if it's left unimplemented. And `Euclidean` looking nicer is a fairly weak reason IMO to introduce such surprising behavior.

> We could turn the `Rn_default` into a `ManifoldsBase.DEFAULT_VECTORSPACE` setting or so? With my persona preference I could still set that to `:Euclidean` (or `:Manifolds`, depending on whether we consider that a manifold or package indicator). Though I am not sure a variable or a function is the best choice for that, this would allow to default it to `:ManifoldsBase`(or `:DefaultManifold`) to keep your preference by default. On the other hand the bad-kellertuer-preference in Manopt.j could be to change that to `:Manifolds` ( or `:Euclidean`) for my convenience (and the users confusion 🤓)

That isn't much better because each time such global setting is used, type instability would be necessary. We do a similar thing for AD backends but that's a different story because we tell the users that it's only for convenience and for performance they should be explicit, and there are actual substantial differences between backends.

If you really want to see `Euclidean` instead of `DefaultManifold`, maybe just implement IPM in `ManoptManifoldsExt`? Then it wouldn't have to touch `DefaultManifold` at all. You've even said recently that `DefaultManifold` should only be used for testing. So it would make its use in IPM not valid, and so IPM should only be available with Manifolds.jl loaded. This would also be a valid solution.

kellertuer: Since this would also only happen on initialisation, I do not see much of a performance issue here. All manifolds (product manifolds) I need are stored.

So with that, something like `vector_space(n)` (even in ManifoldsBase) for people to specify their favourite vector space implementation is something I think is nice (however realised)

mateuszbaran: I can see a few different valid designs, just the one where we use either `DefaultManifold` or `Euclidean` based on whether `Manifolds.jl` is loaded or not is one that makes the least sense to me. These are IMO valid solutions:
1. IPM uses `vector_space` function, which is unimplemented in `ManifoldsBase.jl` and returns `Euclidean` when Manifolds.jl is loaded.
2. IPM has `vector_space` as an input argument.
3. IPM uses `DefaultManifold` explicitly.
4. IPM uses `Euclidean` explicitly and thus requires `Manifolds.jl`.

> To be honest IPM could be available as soon as anyone gives me a vector space (in addition to an arbitrary manifold) so I do not agree to put it into the extension.

Then `vector_space` should be an input argument of IPM, not a global function. It could have a default value based on whether Manifolds.jl is loaded or not, which would be somewhat acceptable because it would be limited in scope to IPM.

kellertuer: I think I like 2, but would really prefer what you wrote in your last part, that this default value is improved when loading Manifolds.jl (sure only in display actually). I am fine if that is only for IPM and hence limited scope, as it is now with `Rn`. So the main point here is, do we want to offer such thing in `ManifoldsBase`?

mateuszbaran: OK, then I think the conclusion is clear here and we should go with 2 but it means that `vector_space` isn't going to be in ManifoldsBase, and every algorithm that requires some kind of Euclidean space should have their own separate implementation of option 2. The last part is basically a small extension of option 2.

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,385 @@
...
+    X=get_gradient(M, cmo, p),
+    μ::Vector=ones(length(get_inequality_constraint(M, cmo, p, :))),
+    λ::Vector=zeros(length(get_equality_constraint(M, cmo, p, :))),
+    s=μ,
+    ρ=μ's / length(μ),
+    σ=calculate_σ(M, cmo, p, μ, λ, s),
+    γ=0.9,
+    retraction_method::AbstractRetractionMethod=default_retraction_method(M, typeof(p)),
+    sub_kwargs=(;),
+    _sub_M=M × Rn(length(λ)),
```
mateuszbaran, line 241:
```suggestion
    vector_space=Rn,
    _sub_M=M × vector_space(length(λ)),
```
This is more or less the key change for option 2 discussed above.

kellertuer: ...and Rn can stay as is? That would be neat and still fine with me :)

mateuszbaran: Yes, Rn could stay as is. I think this would be the best solution based on our discussion :slightly_smiling_face: .

kellertuer: And much easier than I already feared. Will do that tomorrow when I return to docs and tests for IPN (just finished the ConjugateRes part).

src/Manopt.jl
kellertuer: Let's close this since we have the solution remark below open still.

mateuszbaran: I think my comment below should illustrate the solution well enough :slightly_smiling_face: . I wouldn't call your idea "not possible", it's just more confusing than it needs to be with no way out. The `vector_space` argument gives a way out so it makes the design good enough.

kellertuer: As I said, I do not fully understand the difference, but I am fully fine with the `vector_space` argument idea and will follow that. Maybe I was not clear enough above, but the vector_space = thing makes my idea possible as long as I can keep my `Rn`.

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,385 @@
...
+* `grad_h=nothing`: the gradient of the equality constraints
+* `gradient_range=nothing`: specify how gradients are represented, where `nothing` is equivalent to [`NestedPowerRepresentation`](@extref)
+* `gradient_equality_range=gradient_range`: specify how the gradients of the equality constraints are represented
+* `gradient_inequality_range=gradient_range`: specify how the gradients of the inequality constraints are represented
+* `h=nothing`: the equality constraints
+* `Hess_g=nothing`: the Hessian of the inequality constraints
+* `Hess_h=nothing`: the Hessian of the equality constraints
+* `inequality_constraints=nothing`: the number ``m`` of inequality constraints.
+* `λ=ones(size(h(M,x),1))`: the Lagrange multiplier with respect to the equality constraints ``h``
+* `μ=ones(size(h(M,x),1))`: the Lagrange multiplier with respect to the inequality constraints ``g``
```
mateuszbaran, line 56:
```suggestion
* `μ=ones(length(g(M,x)))`: the Lagrange multiplier with respect to the inequality constraints ``g``
```
I changed `h` to `g` and `length` should be more clear than `size`.

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,385 @@
...
+* `s=μ`: initial value for the slack variables
+* `σ=μ's/length(μ)`: ? (TODO find details about barrier parameter)
+* `stopping_criterion::StoppingCriterion=`[`StopAfterIteration`](@ref)`(200)`[` | `](@ref StopWhenAny)[`StopWhenChangeLess`](@ref)`(1e-5)`: a stopping criterion
+* `retraction_method=[`default_retraction_method`](@extref)`(M, typeof(p))`: the retraction to use, defaults to the default set `M` with respect to the representation for `p` chosen.
+* `stepsize=` TODO
+* `sub_kwargs=(;)`: keyword arguments to decorate the sub options, for example debug, that automatically respects the main solvers debug options (like sub-sampling) as well
+* `sub_stopping_criterion=TODO`: specify a stopping criterion for the subsolver.
+* `sub_problem=TODO`: provide a problem for the subsolver, which is assumed to work on the tangent space of `\mathcal M \times ℝ^n`
```
mateuszbaran, line 64:
```suggestion
* `sub_problem=TODO`: provide a problem for the subsolver, which is assumed to work on the tangent space of ``\mathcal M \times ℝ^n``
```

src/plans/constrained_plan.jl
```diff
@@ -343,6 +362,182 @@ end
...
+end
+
+@doc raw"""
+    LagrangianHessian{CO, V, T}
+
+Implement the Hessian of the Lagrangian of a [`ConstrainedManifoldObjective`](@ref) `co`.
+
+```math
+\operatorname{Hess} \mathcal L(p, μ, λ)[X]
+= \operatorname{Hess} f(p) +  \sum_{i=1}^m μ_i \operatorname{Hess} g_i(p)[X] + \sum_{j=1}^n λ_j \operatorname{Hess} h_j(p)[X]
```
mateuszbaran, line 487: A small note: for Hessians and gradients of multi-argument functions like `L` I like to have an indication with respect to which variable it is taken (like in the original paper for IPM). So for example `\operatorname{Hess}_p` instead of just `\operatorname{Hess}`. I think it makes it easier to read.

mateuszbaran: Similarly the text could be more clear that it's the Hessian with respect to `p`.

src/plans/interior_point_Newton_plan.jl
```diff
@@ -0,0 +1,851 @@
...
+) where {R<:Real}
+    N = get_manifold(mp)
+    Y = get_gradient(N, get_objective(mp), ips.q)
+    grad_norm = norm(N, ips.q, Y)
+    max_step = max_stepsize(N, ips.q)
+    return ifelse(isfinite(max_step), min(l, max_step / grad_norm), l)
+end
+
+"""
+InteriorPointCentralityCondition{CO}
```
mateuszbaran, line 730:
```suggestion
    InteriorPointCentralityCondition{CO}
```

src/plans/interior_point_Newton_plan.jl
```diff
@@ -0,0 +1,851 @@
...
+function get_manopt_parameter(ipcc::InteriorPointCentralityCondition, ::Val{:γ})
+    return ipcc.γ
+end
+function set_manopt_parameter!(ipcc::InteriorPointCentralityCondition, ::Val{:γ}, γ)
+    ipcc.γ = γ
+    return ipcc
+end
+
+@doc raw"""
+StopWhenKKTResidualLess <: StoppingCriterion
```
mateuszbaran, line 763:
```suggestion
    StopWhenKKTResidualLess <: StoppingCriterion
```

src/solvers/augmented_Lagrangian_method.jl
```diff
@@ -42,8 +42,8 @@ manifold- or objective specific defaults.
 """
 mutable struct AugmentedLagrangianMethodState{
     P,
-    Pr<:AbstractManoptProblem,
-    St<:AbstractManoptSolverState,
+    Pr<:Union{F,AbstractManoptProblem} where {F},
```
mateuszbaran, line 45: This is equivalent to just `Pr` with no upper bound.

src/solvers/augmented_Lagrangian_method.jl
```diff
@@ -42,8 +42,8 @@ manifold- or objective specific defaults.
...
+    St<:Union{AbstractEvaluationType,AbstractManoptSolverState},
```
mateuszbaran, line 46: Are you sure this is right?

src/plans/constrained_plan.jl
kellertuer: Sure, that sounds good. I will carefully check that when working on this.

src/solvers/interior_point_Newton.jl
kellertuer: will change the line above to length as well then. Thank for spotting the typo.

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,385 @@
...
+* `h=nothing`: the equality constraints
+* `Hess_g=nothing`: the Hessian of the inequality constraints
+* `Hess_h=nothing`: the Hessian of the equality constraints
+* `inequality_constraints=nothing`: the number ``m`` of inequality constraints.
+* `λ=ones(size(h(M,x),1))`: the Lagrange multiplier with respect to the equality constraints ``h``
+* `μ=ones(size(h(M,x),1))`: the Lagrange multiplier with respect to the inequality constraints ``g``
+* `s=μ`: initial value for the slack variables
+* `σ=μ's/length(μ)`: ? (TODO find details about barrier parameter)
+* `stopping_criterion::StoppingCriterion=`[`StopAfterIteration`](@ref)`(200)`[` | `](@ref StopWhenAny)[`StopWhenChangeLess`](@ref)`(1e-5)`: a stopping criterion
+* `retraction_method=[`default_retraction_method`](@extref)`(M, typeof(p))`: the retraction to use, defaults to the default set `M` with respect to the representation for `p` chosen.
```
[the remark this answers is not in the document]
kellertuer: This will become a bit nicer with the other PR (that I will continue after this) that has this line as a default string to include, so we only have to type this once :)

src/solvers/augmented_Lagrangian_method.jl
kellertuer: We can also just write F, sure. The idea here is, that it is
a) an `AbstractManoptProblem` (then `St` is a state) to indicate a(n iterative) subsolver
b) `F` (in the sense of a function, but maybe a functor) for a closed form solution, then `St` is an `AbstractEvaluationType` to indicate whether the subsolver-closed-form function allocates or works in-place.

So both these are reflected in that writing maybe for clarity – but your are right just F would also work here.

src/solvers/augmented_Lagrangian_method.jl
kellertuer: See cases a) (first two types) and b) (second type) in the last comment. I think that is.

src/solvers/augmented_Lagrangian_method.jl
mateuszbaran: I see, though in the b) case I think it would be more clear to store the evaluation type in a dummy `AbstractManoptSolverState` instead of making this a union.

kellertuer: Well in the current form I can easily dispatch on both cases directly, but sure that change would not be too much of work (and could even be considered non-breaking since it is internal detail). However, this would affect all solvers with sub solvers, so would be a bit of work.

mateuszbaran: I just suggest defining
```julia
struct DummyManoptSolverState{E<:AbstractEvaluationType} <: AbstractManoptSolverState end
```
(up to name) so that `sub_state` remains an actual state for consistency. Also note that `sub_state` is also used in `get_message`.

kellertuer: Sounds fair.
`get_message` is mainly a summary of messages in a state (e.g. stopping criterion + stepsize and such). But yeah it might be nice to do that and sounds like a reasonable amount of work (after the current ones are finished). 

Maybe `ClosedFormSubSolverState`?

kellertuer: I still feel just passing the evaluation to the keyword is nice and easy to use, but I will check what best to do there.

mateuszbaran: Thanks! `ClosedFormSubSolverState` sounds fair.

kellertuer: Wrote an issue of this to keep track.

src/solvers/interior_point_Newton.jl
kellertuer: Added that in all places where reasonable (in the high-level interface maybe not yet documented, tests for that are still not yet complete).

kellertuer:
I now carefully documented all functions – found a small bug in one check function – that sadly makes the step sizes a bit more conservative than necessary. Will check that later.

Besides that, there is only 2 or 3 special cases of the high-level interface missing in test coverage (about 20 lines) that should not be much work left.

So I'll set this to ready-to-review already.

src/plans/debug.jl
```diff
@@ -382,6 +382,85 @@ function show(io::IO, di::DebugEntry)
...
+)
+
+"""
+mutable struct DebugFeasibility <: DebugAction
+    atol::Float64
+    format::Vector{Union{String,Symbol}}
+    io::IO
+    function DebugFeasibility(
+        format=["feasible: ", :Feasible]; io::IO=stdout, atol=1e-13
+    )
```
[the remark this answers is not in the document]
kellertuer: Whenever you think: This small change will not affect formatting anyways...

kellertuer:
While Markus did start an interface in manifolds with `p::Numer`, this seems to be a bit too involved for me at least, since the product manifolds might have to be adapted as well. I'll leave that out until there is a need for this.

So besides maybe also checking the default values for performance – this is finished.

kellertuer:
Frank-Wolfe is using it as well, the others not, but for consistency it would be nice to have that unified on all sub solvers.

For the number case – exactly, I also feel that is not so much a useful case.

The one thing still bothering me is performance, I can not figure out how to get the centrality condition (within linesearch) set up to produce okay-ish step sizes.
Currently they are super conservative (they are just mean to never leave the feasible set).
I can easily get them much smaller (`1e-430` or so) but not larger. And I think I follow for 100% the paper. Well.
Maybe I'll write the authors an email when this is published.

mateuszbaran:
OK, I will update the other solvers too then.

Did you compare the centrality condition with any existing implementation of IPM? It looks like Euclidean methods should have something very similar.

kellertuer:
No, I just stumbled upon that today after 4pm when I documented the last doc string (that of the centrality condition).

kellertuer:
Ah, interesting, just accepting all (`centrality_condition= (M,p) -> true`) works better than any other check. Very interesting. Will keep that in one test just to further investigate somewhen later.

kellertuer:
Oh, I have to work a bit more on codecov the next days (deactivated too much debug); but not today, maybe also neither tomorrow nor the day after (was a busy week with work, though I am on vacation actually).

mateuszbaran:
OK, interesting. I still have a few things I'd like to check here. I think I've finished the closed-form subsolver rework, and even found an issue with a test while doing that.

src/solvers/FrankWolfe.jl
```diff
@@ -60,18 +60,18 @@ mutable struct FrankWolfeState{
...
         ITM<:AbstractInverseRetractionMethod,
     }
-        return new{P,T,Pr,Op,TStep,TStop,TM,ITM}(
+        sub_state_storage = maybe_wrap_evaluation_type(sub_state)
+        return new{P,T,Pr,typeof(sub_state_storage),TStep,TStop,TM,ITM}(
             p,
             initial_vector,
             sub_problem,
-            sub_state,
+            sub_state_storage,
```
kellertuer, line 74: Why this renaming? The sub state is usually always the field `sub_state`, so this seems a bit unusual compared to the rest.

mateuszbaran: I've just made a new name for the temporary variable to not have to rely on compiler figuring out how to optimize a variable changing its type.

kellertuer: Ah, I misread that as being the new name of the field. If one reads that correctly this is of course fine.

mateuszbaran:
I've taken a look at line search approaches for Euclidean IPM and it turned out to be very involved. Take a look at, for example, section 17.4 or chapter 19 in https://link.springer.com/book/10.1007/978-3-031-08720-2 .

Anyway, if you want to check what performs better or worse, it would be good to test it on a larger set of problems.

kellertuer:
I actually do not want to check that. I would be fine with a reasonable default for now.

mateuszbaran:
That's perfectly understandable :smile: .

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+
+For that seach direction a line search is performed, that additionally ensures that
+the constraints are further fulfilled.
+
+# Input
+
+* `M`:      a manifold ``\mathcal M``
+* `f`:      a cost function ``f : \mathcal M → ℝ`` to minimize
+* `grad_f`: the gradient ``\operatorname{grad} f : \mathcal M → T \mathcal M`` of ``f``
+* `Hess_f`: the Hessian ``\operatorname{Hess}f(p): T_p\mathcal M → T_p\mathcal M``, ``X ↦ \operatorname{Hess}F(p)[X] = ∇_X\operatorname{grad}f(p)``
```
mateuszbaran, line 47:
```suggestion
* `Hess_f`: the Hessian ``\operatorname{Hess}f(p): T_p\mathcal M → T_p\mathcal M``, ``X ↦ \operatorname{Hess}f(p)[X] = ∇_X\operatorname{grad}f(p)``
```
shouldn't it be small `f`?

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+\operatorname{J} F(p, μ, λ, s)[X, Y, Z, W] = -F(p, μ, λ, s),
+\text{ where }
+X ∈ T_p\mathcal M, Y,W ∈ ℝ^m, Z ∈ ℝ^n,
+```
+
+see [`CondensedKKTVectorFieldJacobian`](@ref) and [`CondensedKKTVectorField`](@ref), respectively,
+for the reduced form, this is usually solved in.
+From the resulting `X` and `Z` in the reeuced form, the other two, ``Y``, ``W``, are then computed.
+
+From the gradient ``(X,y,Z,W)`` at the current iterate ``(p, μ, λ, s)``,
```
mateuszbaran, line 32:
```suggestion
From the gradient ``(X,Y,Z,W)`` at the current iterate ``(p, μ, λ, s)``,
```
And this I think should be capital `Y`.

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+a line search is performed using the [`KKTVectorFieldNormSq`](@ref) norm of the KKT vector field (squared)
+and its gradient [`KKTVectorFieldNormSqGradient`](@ref) together with the [`InteriorPointCentralityCondition`](@ref).
+
+Note that since the vector field ``F`` includes the gradients of the constraint
+functions ``g,h`, its gradient or Jacobian requires the Hessians of the constraints.
```
mateuszbaran, line 37:
```suggestion
functions ``g, h``, its gradient or Jacobian requires the Hessians of the constraints.
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+
+The keyword arguments related to the constraints (the first eleven) are ignored if you
+pass a [`ConstrainedManifoldObjective`](@ref) `cmo`
+
+* `centrality_condition=missing`; an additional condition when to accept a step size.
+  This can be used to ensure that the resulting iterate is still an interior point if you provide a check `(N,q) -> true/false`,
+  where `N` is the manifold of the `step_problem`.
+* `equality_constraints=nothing`: the number ``n`` of equality constraints.
+* `evaluation=`[`AllocatingEvaluation`](@ref)`()`:
+  specify whether the functions that return an array, for example a point or a tangent vector, work by allocating its result ([`AllocatingEvaluation`](@ref)) or whether they modify their input argument to return the result therein ([`InplaceEvaluation`](@ref)). Since usually the first argument is the manifold, the modified argument is the second."
```
mateuszbaran, line 62:
```suggestion
  specify whether the functions that return an array, for example a point or a tangent vector, work by allocating its result ([`AllocatingEvaluation`](@ref)) or whether they modify their input argument to return the result therein ([`InplaceEvaluation`](@ref)). Since usually the first argument is the manifold, the modified argument is the second.
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+        kwargs...,
+    )
+end
+function interior_point_Newton!(
+    M::AbstractManifold,
+    cmo::O,
+    p;
+    evaluation::AbstractEvaluationType=AllocatingEvaluation(),
+    X=get_gradient(M, cmo, p),
+    μ::Vector=ones(length(get_inequality_constraint(M, cmo, p, :))),
```
mateuszbaran, line 220:
```suggestion
    μ::AbstractVector=ones(inequality_constraints_length(cmo)),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    Y=zero(μ),
```
mateuszbaran, line 221:
```suggestion
    Y::AbstractVector=zero(μ),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    λ::Vector=zeros(length(get_equality_constraint(M, cmo, p, :))),
```
mateuszbaran, line 222:
```suggestion
    λ::AbstractVector=zeros(equality_constraints_length(cmo)),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    Z=zero(λ),
```
mateuszbaran, line 223:
```suggestion
    Z::AbstractVector=zero(λ),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    s=copy(μ),
```
mateuszbaran, line 224:
```suggestion
    s::AbstractVector=copy(μ),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    W=zero(s),
```
mateuszbaran, line 225:
```suggestion
    W::AbstractVector=zero(s),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    ρ=μ's / length(μ),
```
mateuszbaran, line 226:
```suggestion
    ρ::Real=μ's / length(μ),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    σ=calculate_σ(M, cmo, p, μ, λ, s),
```
mateuszbaran, line 227:
```suggestion
    σ::Real=calculate_σ(M, cmo, p, μ, λ, s),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+    retraction_method::AbstractRetractionMethod=default_retraction_method(M, typeof(p)),
+    sub_kwargs=(;),
+    vector_space=Rn,
+    #γ=0.9,
+    centrality_condition=missing, #InteriorPointCentralityCondition(cmo, γ, zero(γ), zero(γ)),
+    step_objective=ManifoldGradientObjective(
+        KKTVectorFieldNormSq(cmo), KKTVectorFieldNormSqGradient(cmo); evaluation=evaluation
+    ),
+    _step_M=M × vector_space(length(μ)) × vector_space(length(λ)) × vector_space(length(s)),
```
mateuszbaran, line 236:
```suggestion
    _step_M::AbstractManifold=M × vector_space(length(μ)) × vector_space(length(λ)) × vector_space(length(s)),
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+* `grad_g=nothing`: the gradient of the inequality constraints
+* `grad_h=nothing`: the gradient of the equality constraints
+* `gradient_range=nothing`: specify how gradients are represented, where `nothing` is equivalent to [`NestedPowerRepresentation`](@extref)
+* `gradient_equality_range=gradient_range`: specify how the gradients of the equality constraints are represented
+* `gradient_inequality_range=gradient_range`: specify how the gradients of the inequality constraints are represented
+* `h=nothing`: the equality constraints
+* `Hess_g=nothing`: the Hessian of the inequality constraints
+* `Hess_h=nothing`: the Hessian of the equality constraints
+* `inequality_constraints=nothing`: the number ``m`` of inequality constraints.
+* `λ=ones(length(h(M,x),1))`: the Lagrange multiplier with respect to the equality constraints ``h``
```
mateuszbaran, line 73:
```suggestion
* `λ=ones(length(h(M, p)))`: the Lagrange multiplier with respect to the equality constraints ``h``
```

src/solvers/interior_point_Newton.jl
```diff
@@ -0,0 +1,382 @@
...
+* `μ=ones(length(g(M,x)))`: the Lagrange multiplier with respect to the inequality constraints ``g``
```
mateuszbaran, line 74:
```suggestion
* `μ=ones(length(g(M, p)))`: the Lagrange multiplier with respect to the inequality constraints ``g``
```

mateuszbaran (commented):
A few small things

kellertuer:
I think I am finished with my part including code cov. I kept a new point for the Closed form solver part and its code coverage. That is the only part missing here.

src/solvers/interior_point_Newton.jl
kellertuer: With the `_` upfront I expect this to be an internal keyword only anyways (should not be set by the user, makes a few other keywords shorter to specify the default), but sure we can still type them.

kellertuer:
I think the only two sub solvers not yet following the new style are
* Convex Bundle Method
* Proximal Bundle Method
*  Adaptive Regularisation with Cubics (ARC)
* Difference of Convex Proximal Point
* Difference of Convex
* Trust regions

 (oh, we have quite a few ones with sub solvers)?

kellertuer:
Oh TR looks a bit more involved, the code is quite old and might also benefit from a bit of improvements anyways. Do you want to do that or shall I?

mateuszbaran:
I can adapt TR.

kellertuer:
We have all others _and_ proper code cov.

kellertuer:
Oh, I introduced an ambiguity? Why did that not happen for the other solvers? I do not directly se how to resolve that.

mateuszbaran:
I've adapted TR and resolved the ambiguity.

kellertuer:
I am not so sure where we lost the 2 lines on TR, but the rest seems to be fine.

kellertuer:
Yay! I'd call this finished, but would be happy if you could approve it as well.
We can also wait until tomorrow with merging.

mateuszbaran:
I would like to take one more look at this PR tomorrow, mostly to check if there are any major performance issues or some small documentation improvements.

kellertuer:
Sure, If you find something and do a commit, you can also adapt the date in the changelog.
I can check back here tomorrow afternoon then, since I will be on a hike tomorrow (one we can also do when you are here).

mateuszbaran (approved):
Performance is not great but to improve that I need to extend `PowerManifold` a bit first. I think this would be better suited for a separate PR.

Changelog.md
```diff
@@ -5,12 +5,35 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this
...
 
 ### Added
 
 * `max_stepsize` methods for `Hyperrectangle`.
 
+### Added
+
+* Introduces a Riemannian Interior Point Newton Method based on the paper by Lai & Yoshise.
+* Add a `is_feasible` function and a detailed `DebugFeasibility` debug action
+
```
mateuszbaran, line 36: Oh, and this I think shouldn't be in the "0.4.67" section.

kellertuer:
> Performance is not great but to improve that I need to extend `PowerManifold` a bit first. I think this would be better suited for a separate PR.

I think s well, this took quite a while to get done, so working is my first goal here, performance a next one.
I did work through most of the access functions used for constraints and used the single access ones for performance for example.

Changelog.md
kellertuer: Ah, yes, I  must have missed that on a merge, maybe this merge went too well at that point ;)

PR #402, opened 2024-07-24
max_stepsize improvements

mateuszbaran:
I've been debugging some issues with optimization on Hyperrectangle and noticed it lacks `max_stepsize`. Also, `WolfePowellLinesearch` by default tries calling `max_stepsize` with an invalid point. If it tries to get the global maximum stepsize, it should not provide the point, and if it wants local maximum stepsize (which would make some sense), it should call `max_stepsize` before each search separately.

### DIFF
--- a/ext/ManoptManifoldsExt/manifold_functions.jl
+++ b/ext/ManoptManifoldsExt/manifold_functions.jl
@@ -22,6 +22,12 @@ function max_stepsize(M::FixedRankMatrices, p)
 end
 max_stepsize(M::FixedRankMatrices) = manifold_dimension(M)
 
+"""
+    max_stepsize(M::Hyperrectangle, p)
+
+The default maximum stepsize for `Hyperrectangle` manifold with corners is maximum
+of distances from `p` to each boundary.
+"""
 function max_stepsize(M::Hyperrectangle, p)
     ms = 0.0
     for i in eachindex(M.lb, p)
--- a/ext/ManoptManifoldsExt/manifold_functions.jl
+++ b/ext/ManoptManifoldsExt/manifold_functions.jl
@@ -22,13 +22,26 @@ function max_stepsize(M::FixedRankMatrices, p)
 end
 max_stepsize(M::FixedRankMatrices) = manifold_dimension(M)
 
-max_stepsize(M::Hyperrectangle, p) = max_stepsize(M)
+function max_stepsize(M::Hyperrectangle, p)
+    ms = 0.0
+    for i in eachindex(M.lb, p)
+        dist_ub = M.ub[i] - p[i]
+        if dist_ub > 0
+            ms = max(ms, dist_ub)
+        end
+        dist_lb = p[i] - M.lb[i]
+        if dist_lb > 0
+            ms = max(ms, dist_lb)
+        end
+    end
+    return ms
+end
 function max_stepsize(M::Hyperrectangle)
     ms = 0.0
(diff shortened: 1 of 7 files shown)


src/plans/stepsize.jl
```diff
@@ -872,7 +872,7 @@ mutable struct WolfePowellLinesearch{
         candidate_point::P=allocate_result(M, rand),
         candidate_tangent::T=allocate_result(M, zero_vector, candidate_point),
         candidate_direction::T=allocate_result(M, zero_vector, candidate_point),
-        max_stepsize::Real=max_stepsize(M, candidate_point),
+        max_stepsize::Real=max_stepsize(M),
```
kellertuer, line 875: sure taking the global max stepsize here makes sense – however I would expect the candidate point to be a valid point so the local variant should in Principe be fine as well?

ext/ManoptManifoldsExt/manifold_functions.jl
```diff
@@ -18,7 +18,30 @@ the choice of typical distance in Matlab Manopt, the dimension of `M`. See
 [this note](https://github.com/NicolasBoumal/manopt/blob/97b6eb6b185334ab7b3991585ed2c044d69ee905/manopt/manifolds/fixedrank/fixedrankembeddedfactory.m#L76-L78)
 """
 function max_stepsize(M::FixedRankMatrices, p)
-    return manifold_dimension(M)
+    return max_stepsize(M)
+end
+max_stepsize(M::FixedRankMatrices) = manifold_dimension(M)
+
+function max_stepsize(M::Hyperrectangle, p)
```
kellertuer, line 25: I have (still) not yet understood the hyper rectangle too much, but I believe you that this is correct. maybe we should stat documenting the max stepsize functions with formulae as well (then put them into the extension page)?

src/plans/stepsize.jl
mateuszbaran: The default value for `candidate_point` is not valid, and even if it was valid, it would make an implicit assumption that the point doesn't really matter. If it doesn't matter, why would it be passed there?

kellertuer: ah because we just allocate it; I was having in mind that would be the start point or something. Then either using the start point or the current version of the global max are fine.

ext/ManoptManifoldsExt/manifold_functions.jl
mateuszbaran: Eventually we should have a custom line search for manifolds with corners that handles break points (positions when we hit a new boundary by following the descent direction with projection) but that's a somewhat bigger project. This `max_stepsize` is essentially a stop-gap solution that allows existing methods to sometimes work well enough. Right now `max_stepsize` defaults to 0 which makes line searches never work on `Hyperrectangle`. This is somewhat similar in spirit to `TangentBundle` where the global injectivity radius is also 0 but we have a specialized `max_stepsize`.

I can add a documentation about it somewhere.

kellertuer: Thanks for the explanation, yes a bit of docs would be great especially since this is an interims fix.

mateuszbaran: I've added docs regarding `max_stepsize`.

kellertuer (approved):
Looks good to me.

We can merge and register tomorrow for example?

mateuszbaran:
Sure, I will update the date in changelog.

PR #404, opened 2024-08-02
Improve performance of IPM

mateuszbaran:
This significantly improves performance of interior point Newton method. Together with some light improvements to ManifoldsBase, I got from 9ms to below 3ms on the problem from the test suite. The two biggest issues were:
1. Accessing product manifold components like `Y[N, 1]` is really slow. I recommend using things like `Y1, Y2 = submanifold_components(N, Y)` in performance-sensitive code. Maybe it would be possible to speed up `Y[N, 1]` but it would require tinkering with compiler heuristics which is quite difficult and depends on the version of Julia.
2. Broadcasting. It can eliminate vast majority of intermediate arrays.

Not that many potential gains remain as those 3ms are dominated by calls to Hessian and gradient of `g`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,12 @@ All notable Changes to the Julia package `Manopt.jl` will be documented in this 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.69] – unreleased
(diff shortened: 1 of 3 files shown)


kellertuer:
I was not aware of the first. For the second I was maybe more focused on the math being correct. Thanks for fixing this :)

kellertuer (approved):
at least the sub manifold components I would not have known to use. thanks for those improvements.

PR #409, opened 2024-08-29
Adjust text for ExponentialFamilyProjection.jl

bvdmitri:
This PR improves the text about ExponentialFamilyProjection and also notes RxInfer as a parent package.

Thanks @kellertuer 

Ref https://github.com/ReactiveBayes/ExponentialFamilyProjection.jl/issues/37


kellertuer:
Thanks that was exactly what I was asking for to make it more precise. Will merge this and it will be part of the next release then.

PR #412, opened 2024-09-20
update readme logo with text.

kellertuer:


mateuszbaran:
I just thought, maybe the lines in the logo could use Julia colors instead of being black? There are two groups of three lines, which would fit nicely.

kellertuer:
In the long run that is one of the plans why I started to redo the logo in Julia; for now this is the logo file generated with the Matlab code, where the contours are computed in angles (as if that were part of $\mathbb R^2$); in the Matlab code I am not able to choose individual colours. Also Julia colors might require the colormap to be adapted as well, so the logo still looks nice.

PR #428, opened 2024-12-06
Prefix InterLinks links with the logos of the corresponding packages

kellertuer:
Here is a solution based on the hints in
https://github.com/JuliaDocs/DocumenterInterLinks.jl/issues/13#issuecomment-2522479911

I feel this is nice and if of (@mateuszbaran ) I would to the same in Manifolds.jl.


mateuszbaran:
Sure, it looks nice :+1:

PR #432, opened 2024-12-27
Rework LevenbergMarquardt to use the vector function functionality

### LINKED ISSUE #332: Robust loss functions for outlier rejection
Is it possible to use loss functions such as Huber, Tukey, and adaptive in Manopt.jl (specifically RLM)?
Similar to ceres solver: http://ceres-solver.org/nnls_modeling.html#lossfunction

kellertuer:
This make the interface for Levenberg-Marquardt a bit more flexible.

# 🛣️🗺️
* [x] rework vector functions to also provide a proper jacobian interface
* [x] store the cost in the nonlinear least squares objective as a `VectorGradientFunction`
* [x] rework `get_jacobian` (formerly “get-gradient-from-jacobian” to “pass through to the new inner `vgf`” without smoothing
* [x] check that all new functions are properly documented
* [x] work on test coverage 📈 

The main thing to figure out is the new Jacobian to return in case of smoothing and how to adapt the regularisation parameter $\lambda_k$ for this case. This last point especially I have not yet understood, neither from the ceres documentation linked above or the nearly-the-same phrasing at the end of Section 4.3 in Triggs et al.


kellertuer:
This was faster than I thought. Tests should be up and running again for the rework.

For LM this is even non-breaking, since what mainly changed is the internal representation of the vector function and its jacobian as a `VectorGradientFunction`.
This even allows to now provide instead of a jacobian a function returning all gradients or a vector of gradient functions in LM.

Smoothing should be covered as well, though the $\lambda_k$ is probably still a “poor-mans-version” that for most cases should work but is not so much covered by theory.

For general vector functions there is one small (breaking) change, though that was more of a bug I think:
Previously specifying a vector function to be `InplaceEvaluation` only made the jacobian (and Hessian) mutating versions. The actual vectorial (cost) function was always assumed to be allocating. That was more by accident and also quite inconsistent.
Now also the actual vector function then has to be mutating. I might have to check that `get_value!` is consistent with this as well.

mateuszbaran:
I've taken a look and I think we need to generalize robustification a bit. Both Ceres and Triggs consider a more general objective, where the smoothing function $\rho_i$ is applied to a group of observations instead of a single observation like in your PR. See for example Eq. (3) in Triggs. I think restricting each group to 1 element makes this extension much less useful because the least squares Jacobians are computed on a per-smoothing block basis. Moreover, it's even a performance regression for the non-smoothing case because cost and Jacobian computations are much harder to vectorize.

I think for simplicity we can assume that each group has the same size but we need to let it be arbitrarily large.

kellertuer:
I thought about that a bit as well. My main problem is, that this current rework I could do in about a week. It still got quiiiite technical for the vectorial function, but nice to have Jacobians therein now covered nicely.

But for arbitrary block the current structure of the smoothing thingy is useless. that would need a complete rewrite of this current work. I am not sure how that can be done and how much time that would need.

So if this is useless without blocks – we should just abandon smoothies from this PR and only keep the vectorial rework.

mateuszbaran:
OK, then I think it's better to do the vectorial rework here and add smoothing when we have time to work on blocks.

kellertuer:
sounds fair. Sad for the whole factory I built, but then it is what it is, then I wasted my energy on that.

We are also missing the proper `lambda_k` for smoothing anyways, so then lets remove smoothing again and do that somewhen later. will add a comment on the original issue about this as well.

mateuszbaran:
We can finish smoothing later, it's certainly not a wasted effort. Maybe a little bit on the premature integration but finishing this will be a good next project IMO (after LieGroups.jl and GeometricKalman.jl).

kellertuer:
yeah will try to put it aside on another branch then.

kellertuer:
I reworked this PR again. There is now a “start-smoothing” branch to keep the current approach.

There was something I did not like anyways: Ceres does smoothing only _after_ absolute value and squaring the residuals. That way “Huber” is more like “Huber on the sqrt” to cancel the squaring.
When we redo this again, we should be more flexible here and have $s_i(x) = \lVert x \rVert_2^2$ as the default modifier/outer function (on the blocks) and then do a proper huber (and not one on the sqrt) to replace $s_i$.

Currently these modifiers (smoothing) can not have parameters, also one thing one could improve in a new PR.
I would then also prefer the term modifier, since smoothing is not so well-phrased.

mateuszbaran:
Hi! We need to derive efficient update rules to add smoothing to Manopt and it appears to be more work than initially anticipated. We already have two big ongoing projects so this will take some time. The rework in the PR makes nonlinear least squares in Manopt a bit more flexible but it probably won't be useful to you until smoothing is added.

kellertuer:
I would still be interested in the example, but the main problem for now is that the link in the issue to Ceres and the literature in there does the rescaling/updates for non smooth objectives for the __Euclidean__ case, but no one has ever done that on manifolds.

As Mateusz said, this is something we both find interesting and want to derive, but that is not something we have a time estimate for when this will happen, so we omit that from this PR for now. yes than this PR “only” reworks internals of LM, but that is still nice to have – just from a feature perspective, it is super boring.

src/documentation_glossary.jl
```diff
@@ -81,13 +82,23 @@ define!(:LaTeX, :log, raw"\log")
...
 define!(:LaTeX, :quad, raw"\quad")
 define!(:LaTeX, :reflect, raw"\operatorname{refl}")
 define!(:LaTeX, :retr, raw"\operatorname{retr}")
+define!(:LaTeX, :rm, (letter) -> raw"\mathrm{" * "$letter" * "}")
+define!(:LaTeX, :sqrt, (s) -> raw"\sqrt{" * "$s}")
 define!(:LaTeX, :subgrad, raw"∂")
 define!(:LaTeX, :sum, raw"\sum")
+define!(:LaTeX, :transp, (letter) -> raw"\text{" * "$letter" * "}")
 define!(:LaTeX, :text, (letter) -> raw"\text{" * "$letter" * "}")
+define!(:LaTeX, :transp, raw"\mathrm{T}")
```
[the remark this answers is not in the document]
kellertuer: For – eh – security reasons!

No, the first one is too much and overwritten by the second anyways.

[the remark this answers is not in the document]
mateuszbaran: I see, the [WET](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#WET) principle :smile:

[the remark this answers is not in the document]
kellertuer: Yeah.

Thanks for noticing, fixed it.

src/plans/nonlinear_least_squares_plan.jl
```diff
@@ -1,132 +1,178 @@
...
+    M::AbstractManifold,
+    nlso::NonlinearLeastSquaresObjective{
+        E,<:AbstractVectorFunction{E,<:ComponentVectorialType}
+    },
+    p;
+    kwargs...,
+) where {E<:AbstractEvaluationType}
+    v = 0.0
+    for i in 1:length(nlso.objective)
+        v += abs(get_value(M, nlso.objective, p, i))^2
```
mateuszbaran, line 80: Since the values are real, `abs` seems to be redundant here I think?

mateuszbaran: By the way, IMO the canonical implementation should be the component-free one for nonlinear least squares, not the one that accesses components one by one.

kellertuer: probably a bit, yes, depending on how you interpret the objective/residual. 

Might be work to keep this in case we generalise to complex residuals later.

kellertuer: > By the way, IMO the canonical implementation should be the component-free one for nonlinear least squares, not the one that accesses components one by one 

I do not understand what you mean, I think I did not change logic anywhere, I just wrapped f and jacobian into the vector function. The case here is where single functions are stored ( $f = (f_1, f_2, ... f_n)$) hence the component wise access. If one function is stored, this is not done.

mateuszbaran: > Might be work to keep this in case we generalise to complex residuals later.

OK.


> I do not understand what you mean, I think I did not change logic anywhere, I just wrapped f and jacobian into the vector function. The case here is where single functions are stored ( $f = (f_1, f_2, ... f_n)$) hence the component wise access. If one function is stored, this is not done.


Ah, right, the single function one is below.

src/plans/vectorial_plan.jl
```diff
@@ -262,54 +296,517 @@ function VectorHessianFunction(
...
+Evaluate the Hessians of the vector function `vgf` on the manifold `M` at `p` in direction `X`
+and the values given in `range`, specifying the representation of the gradients.
 
+Since `i` is assumed to be a linear index, you can provide
 * a single integer
 * a `UnitRange` to specify a range to be returned like `1:3`
 * a `BitVector` specifying a selection
 * a `AbstractVector{<:Integer}` to specify indices
-* `:` to return the vector of all gradients, which is also the default
+* `:` to return the vector of all gradients
```
mateuszbaran, line 321: Since it's `get_hessian`, it's probably not gradients?

kellertuer: Yes, that was a copy paste error before that I also did not see when unifying the doc strings here. It should read “vector of all Hessian evaluations” (it still returns a vector of tangent vectors).

src/documentation_glossary.jl
```diff
@@ -231,25 +242,39 @@ define!(
...
     :Problem,
-    :Default,
-    (; M="M", p="p") -> "\n```math\n$(_tex(:argmin))_{$p ∈ $(_tex(:Cal, M))} f($p)\n```\n",
+    :NonLinearLeastSquares,
+    (; M="M", p="p") -> """
+```math
+$(_tex(:argmin))_{$p ∈ $(_math(:M; M=M))} $(_tex(:frac,1,2)) $(_tex(:sum))_{i=1}^n $(_tex(:abs, "f_i($p)"))^2
+```
+
+where ``f: $(_math(:M; M=M)) → ℝ^n`` is written with component functions ``f_i: $(_math(:M; M=M)) → ℝ``,
```
mateuszbaran, line 273: The number of components in an NLS problem is denoted by `m` in most places I think?

mateuszbaran:
LGTM up to the `n`/`m` inconsistency :+1:

src/documentation_glossary.jl
kellertuer: Oh then I have to check again, yes I confused that to be `n` for some time but had thought I fixed all occurrences.
Letters. Nearly as complicated as Numbers.

Affie:
Hi, here is a draft pr with a little toy example triangulating a point on SE2 with bearing measurements to 4 points on TranslationGroup(2) with one outlier: https://github.com/JuliaRobotics/RoME.jl/pull/767

mateuszbaran:
Thanks! We will try that when working on robustness.

mateuszbaran (approved):

PR #447, opened 2025-04-08
Customizable subsolver for Levenberg-Marquardt

mateuszbaran:
The stuff we've just discussed during JuliaManifolds community call.

### DIFF
--- a/src/solvers/LevenbergMarquardt.jl
+++ b/src/solvers/LevenbergMarquardt.jl
@@ -210,14 +210,15 @@ end
 
 function default_lm_lin_solve!(sk, JJ, grad_f_c)
     try
-        sk .= cholesky(JJ) \ -grad_f_c
+        ldiv!(sk, cholesky(JJ), grad_f_c)
     catch e
         if e isa PosDefException
-            sk .= JJ \ -grad_f_c
+            sk .= Symmetric(JJ) \ grad_f_c
         else
             rethrow()
         end
     end
+    sk .*= -1
     return sk
 end
 
--- a/src/solvers/LevenbergMarquardt.jl
+++ b/src/solvers/LevenbergMarquardt.jl
@@ -215,6 +215,8 @@ Solve the system `JJ \\ grad_f_c` where JJ is (mathematically) a symmetric posit
 definite matrix and save the result to `sk`. In case of numerical errors the
 `PosDefException` is caught and the default symmetric solver `(Symmetric(JJ) \\ grad_f_c)`
 is used.
+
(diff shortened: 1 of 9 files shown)


mateuszbaran:
@Affie I've added support for the custom subsolver. It should be safer by default. If it keeps failing, could you add a print like `println(lms.damping_term, " | ", norm(lms.residual_values))` to `step_solver!` for LM and give me the values?

kellertuer:
You could also print those with `debug=[:iteration, " ", :damping_term, " | ", :residual_values]` as a keyword passed to LM, so that you do not have to clone and use Manopt in development mode.
I did not run this code but I hope I remember my tricks correctly ;)

Affie:
> You could also print those with `debug=[:iteration, " ", :damping_term, " | ", :residual_values]` as a keyword passed to LM, so that you do not have to clone and use Manopt in development mode. I did not run this code but I hope I remember my tricks correctly ;)

Yes, I discovered that before and used `debug = [:Iteration, :Change,  " | ", :Cost, " | ", :Stepsize," | ", :GradientNorm," | ", :damping_term, "\n", :Stop]` previously.

src/solvers/LevenbergMarquardt.jl
```diff
@@ -208,6 +208,19 @@ function initialize_solver!(
     return lms
 end
 
+function default_lm_lin_solve!(sk, JJ, grad_f_c)
```
kellertuer, line 211: could you maybe add a small doc string here as well so that the doc string above could `@ref` to this?

Alternatively when mentioning the default describe short what it does (cholesky or \).

kellertuer:
Yes I once had the idea that any “unknown symbol” (where e.g. `:Cost` is a “known” one) would be used to print the corresponding property from the state. And every now and then I need it and remember, that I think it is nice to have.

I hope it is documented in the debug factory, but good that you already know it and use it :)

src/solvers/LevenbergMarquardt.jl
mateuszbaran: Sure, I've added a docstring.

kellertuer: Ah and a

```@docs
Manopt.default_lm_lin_solve!
```
block or such is added in an internals section for LM in the docs.

mateuszbaran: OK, done.

kellertuer (approved):
LGTM, thanks for stabilising it :)

Issue #273, opened 2023-06-21
Riemannian black box function with JuMP/MOI


blegat:
```julia
"""
    struct RiemannianFirstOrderFunction{F,G} <: MOI.AbstractScalarFunction
        value::F
        gradient!::G
    end

Function defined on a manifold. Given a point `x` of the manifold, `value(x)` provides the value at point `x`
and `gradient!(g, x)` set `g` to be the Riemannian gradient at `x`.

!!! note
    `x` is not in a vectorized format but in the format of points in the manifold.
"""
struct RiemannianFirstOrderFunction{F,G} <: MOI.AbstractScalarFunction
    value::F
    gradient::G
end
```
From JuMP, you can then do something like
```julia
@variable(model, X in manifold)
@objective(model, Min, RiemannianFirstOrderFunction(X -> ..., (g, X) -> ...))
```
If the user sets an objective without `RiemannianFirstOrderFunction` then it is transformed into these as in https://github.com/JuliaManifolds/Manopt.jl/pull/264
The transformation from MOI's functions into `RiemannianFirstOrderFunction`'s function may be doable transparently in a `MOI.Bridges.Objective` so that Manopt does not even have to care about anything other than `RiemannianFirstOrderFunction` an may assume the user always give that.

kellertuer:
This transformation is indeed on my list for a while
https://github.com/JuliaManifolds/Manopt.jl/issues/217
though basically the inverse, since Manopt always assumes the Riemannian case and I would like to offer the Euclidean as well, but I was a bit busy this semester with other duties.

Besides that, I like the notion of `@variable(model, X in manifold)` – it seems to be what I asked for with the domain discussion, just that you embed the domain as part of the variable – which for me is basically the same just a bit of a different viewpoint – and maybe even one that fits better into MOI.

Is this in deed an issue for Manopt or more one for MOI?

Also note that in Manopt, the signatures of cost/grad/hessian/prox/... have the manifold as their first variable, since there might also be different metrics “at work” and such.

blegat:
A better one might indeed be
```julia
struct RiemannianFunction{MO<:AbstractObjective} <: MOI.AbstractScalarFunction
    func::MO
end
```

> Is this in deed an issue for Manopt or more one for MOI?

We always start defining new sets and functions in the packages supporting them and them move to MOI if other solvers support it as well and they don't want more dependencies than MOI. I would just define it in Manopt for now.

kellertuer:
[...]

PR #448, opened 2025-04-10
Add support for Riemanian objective through JuMP

blegat:
Closes https://github.com/JuliaManifolds/Manopt.jl/issues/273

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   especially turn maximisation problems into minimisation ones using a scaling of `-1`.
 * Introduce a `ManifoldConstrainedSetObjective`
 * Introduce a `projected_gradient_method`
+* Allow setting `AbstractManifoldObjective` through JuMP
 
 
 ## [0.5.11] April 8, 2025
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,12 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.13] unreleased
+
+### Added
+
+* Allow setting `AbstractManifoldObjective` through JuMP
+
 ## [0.5.12] April 13, 2025
 
 ### Added
@@ -14,7 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   especially turn maximisation problems into minimisation ones using a scaling of `-1`.
 * Introduce a `ManifoldConstrainedSetObjective`
 * Introduce a `projected_gradient_method`
-* Allow setting `AbstractManifoldObjective` through JuMP
 
 
 ## [0.5.11] April 8, 2025
--- a/ext/ManoptJuMPExt.jl
+++ b/ext/ManoptJuMPExt.jl
@@ -328,7 +328,7 @@ function MOI.optimize!(model::Optimizer)
         end for i in eachindex(model.variable_primal_start)
     ]
     objective = model.objective
-    if model.sense == MOI.FEASIBILITY_SENSE
+    if isnothing(objective) || model.sense == MOI.FEASIBILITY_SENSE
         eval_f_cb(_, _) = 0.0
         function eval_grad_f_cb(M, X)
             x = JuMP.vectorize(X, _shape(model.manifold))
--- a/ext/ManoptJuMPExt.jl
+++ b/ext/ManoptJuMPExt.jl
@@ -29,7 +29,7 @@ function MOI.dimension(set::VectorizedManifold)
     return prod(ManifoldsBase.representation_size(set.manifold))
 end
 
-struct RiemannianFunction{MO<:ManifoldsBase.AbstractManifoldObjective} <: MOI.AbstractScalarFunction
+struct RiemannianFunction{MO<:Manopt.AbstractManifoldObjective} <: MOI.AbstractScalarFunction
     func::MO
 end
 
@@ -45,7 +45,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     # Sense of the optimization, that is whether it is for example min, max or no objective
     sense::MOI.OptimizationSense
     # Objective function of the optimization
-    objective::ManifoldsBase.AbstractManifoldObjective
+    objective::Manopt.AbstractManifoldObjective
     # Solver parameters set with `MOI.RawOptimizerAttribute`
     options::Dict{String,Any}
     function Optimizer()
--- a/ext/ManoptJuMPExt.jl
+++ b/ext/ManoptJuMPExt.jl
@@ -332,7 +332,7 @@ function MOI.optimize!(model::Optimizer)
     ]
     objective = model.objective
     if model.sense == MOI.FEASIBILITY_SENSE
-        objective = Manopt.ManifoldGradientObjective((_, _) -> 0.0, (_, X) -> zero(X))
+        objective = Manopt.ManifoldGradientObjective((_, _) -> 0.0, ManifoldsBase.zero_vector)
     elseif model.sense == MOI.MAX_SENSE
         objective = -objective
     end
--- a/ext/ManoptJuMPExt.jl
+++ b/ext/ManoptJuMPExt.jl
@@ -286,14 +286,7 @@ MOI.get(model::Optimizer, ::MOI.ObjectiveSense) = model.sense
 
 Set the objective function as `func` for `model`.
 """
-function MOI.set(model::Optimizer, ::MOI.ObjectiveFunction, func::RiemannianFunction)
-    model.objective = func
-    model.problem = nothing
-    model.state = nothing
-    return nothing
-end
-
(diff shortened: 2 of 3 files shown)


ext/ManoptJuMPExt.jl
```diff
@@ -304,36 +327,11 @@ function MOI.optimize!(model::Optimizer)
...
-            MOI.eval_objective_gradient(evaluator, grad_f, x)
-        end
-        if model.sense == MOI.MAX_SENSE
-            LinearAlgebra.rmul!(grad_f, -1)
-        end
-        reshaped_grad_f = JuMP.reshape_vector(grad_f, _shape(model.manifold))
-        return ManifoldDiff.riemannian_gradient(model.manifold, X, reshaped_grad_f)
+    objective = model.objective
+    if model.sense == MOI.MAX_SENSE
+        objective = -objective
```
blegat, line 332: This is the main difficulty I see, how do we multiply an `AbstractManifoldObjective` by `-1` ?

blegat: Yes, also we would need to support all types of different `AbstractManifolObjective`. Maybe we can start with just throwing an error and forcing the user to use a minimization ?

kellertuer: One idea could be to write a new wrapper like https://manoptjl.org/stable/plans/objective/#Manopt.EmbeddedManifoldObjective
That one internally stores the objective defined on the embedding space and converts every call to the cost/grad/... to a Riemannian call.

We could do the same with a generic “ScaledManifoldObjective” – and even define the unary minus operator for abstracts then. I would maybe not do that for the constraint objectives but at least for all up to hessian?

blegat: That would be a solution, maybe you can do a separate PR with this feature and then we continue this one once this is merged ?

kellertuer: Yes, sure.

kellertuer:
I just finished #449 which does what you need (hopefully) at least for all objectives that (only) need cost/grad/hess – but it could in principle also be done for the non smooth prox ones.

So if you merge master into yours, you have even an `-obj` that creates the `ScaledManifoldObjective` wrapper :)

blegat:
I prefer smaller PRs whenever possible

ext/ManoptJuMPExt.jl
```diff
@@ -29,6 +29,10 @@ function MOI.dimension(set::VectorizedManifold)
     return prod(ManifoldsBase.representation_size(set.manifold))
 end
 
+struct RiemannianFunction{MO<:ManifoldsBase.AbstractManifoldObjective} <: MOI.AbstractScalarFunction
```
kellertuer, line 32:
```suggestion
struct RiemannianFunction{MO<:Manopt.AbstractManifoldObjective} <: MOI.AbstractScalarFunction
```

ext/ManoptJuMPExt.jl
```diff
@@ -40,8 +44,8 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     variable_primal_start::Vector{Union{Nothing,Float64}}
     # Sense of the optimization, that is whether it is for example min, max or no objective
     sense::MOI.OptimizationSense
-    # Model used to compute gradient of the objective function with AD
-    nlp_model::MOI.Nonlinear.Model
+    # Objective function of the optimization
+    objective::ManifoldsBase.AbstractManifoldObjective
```
kellertuer, line 48:
```suggestion
    objective::Manopt.AbstractManifoldObjective
```

kellertuer (commented):
The objective is an optimization part so it is part of Manopt not ManifoldsBase.
At the end we also need a Changelog entry for the new patch version (that is one of the tests failing).

I had hoped I can edit this locally and contribute a tutorial to this PR as well, that seems to not be the case I can only propose changes here (but maybe accept them to see how far CI gets?)

blegat:
Weird, "Allow edits and access to secrets by maintainers" is checked

kellertuer:
Sure I can pull and I can accept these edit comments here. But doing those feels a bit like hand-craftedly sending postcards to GitHub ;) I can not push from my locally checked out branch.

Whatever, then we do this PR shorter – you like that anyways :) - and just do the feature improvement. I can try locally already to get a tutorial working and that will be a next PR.

For not what is missing is
* Documenter is missing `MathOptInterface.set` which should go to the extension page `docs/src/extensions.md` in the right section
* I could not figure out the CI failure and conversion errors the new code still does.

kellertuer:
Ah and the `Changelog.md` in the main folder needs a short comment for the new version that the interface was generalised.

blegat:
What can we do if `model.objective` is `nothing`. Is there an easy way to get a zero objective to be used as second argument to `decorate_objective!` ?

kellertuer:
What would that be used for? I currently do not have in mind what a “zero objective” mightt be.

The easiest “empty” objective is probably https://manoptjl.org/stable/plans/objective/#Manopt.ManifoldCostObjective so something like

```
obj = ManifoldCostObjective( (M,p) -> -Inf )
```

(or any other constant that `-Inf`)

ext/ManoptJuMPExt.jl
```diff
@@ -304,36 +327,20 @@ function MOI.optimize!(model::Optimizer)
...
-            grad_f .= zero(eltype(grad_f))
-        else
-            MOI.eval_objective_gradient(evaluator, grad_f, x)
-        end
-        if model.sense == MOI.MAX_SENSE
-            LinearAlgebra.rmul!(grad_f, -1)
-        end
-        reshaped_grad_f = JuMP.reshape_vector(grad_f, _shape(model.manifold))
-        return ManifoldDiff.riemannian_gradient(model.manifold, X, reshaped_grad_f)
+        objective = Manopt.ManifoldGradientObjective(eval_f_cb, eval_grad_f_cb)
```
blegat, line 339: I did this at the moment as it's equivalent to what we had before this PR but that's probably a bit convoluted ^^

kellertuer: well if the objective is already nothing then doing a ManifoldDiffGrad of that is not so reasonable ;)

blegat: What is a ManifoldDiffGrad ?

kellertuer: Sorry just me being lazy to not write the full `ManifoldDiff.riemannian_gradient(...)`.

blegat: `riemannian_gradient` is in the removed lines of the diff

blegat: Yes, probably, `zero(X)` would work

Changelog.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   especially turn maximisation problems into minimisation ones using a scaling of `-1`.
 * Introduce a `ManifoldConstrainedSetObjective`
 * Introduce a `projected_gradient_method`
+* Allow setting `AbstractManifoldObjective` through JuMP
```
kellertuer, line 17: I would have loved to edit this but one this that is really horrible on web interface editing is these multiline suggestions, I never can get them to work as I hoped for.

We already released 0.5.12 a few days ago since I needed that for another algorithm so this has to be moved to a new 0.5.13 section (“unreleased” we addd the date later in the tutorial)

ext/ManoptJuMPExt.jl
kellertuer: That is `X` here? The tangent vector (gradient)? Then we have `zero_vector(M,p)` for that. But no need to do a gradient, when you use the `Manifold)CostObjective` (which does not have a gradient)

blegat: But then we'll have to use a zero-order method instead of a first-order one right ?

kellertuer: Sure, but most of that code is magic to me that comes from MOI here, so I have not yet understood why the objective might suddenly be nothing.

And with a constant cost neither zeroth nor first order methods make sense either.

So I am not yet sure what all this case is about.

blegat:
It seems there is an issue with scaled objective because [e98ea6f](https://github.com/JuliaManifolds/Manopt.jl/pull/448/commits/e98ea6fb1114e4442151b4562291a0c16c75c9f1) fixed the tests

blegat:
With the scaled objective that you have added, the point does not move, very weird

blegat:
The tests are passing locally, I won't have time to do more today but feel free to merge if ci is green in order to iterate on a follow up PR

ext/ManoptJuMPExt.jl
blegat: This happens if the user don't use `@objective`. The way you can interpret this is that the solver is free to return any point of the manifold

blegat:
The issue with the current approach is that the flipping of the objective sense don't happen if a riemannian objective is set

kellertuer:
> The tests are passing locally, I won't have time to do more today but feel free to merge if ci is green in order to iterate on a follow up PR

Nice! For me passing tests is fine enough, because we also have a test for the extension so it still works as expected. And I can try to write a short quarto-based tutorial and link to it from the extensions page as well in a next PR

> The issue with the current approach is that the flipping of the objective sense don't happen if a riemannian objective is set

That I do not fully understand but I will check the code.

Thanks for all your work, would be super cool to maybe also get the other open point with non-array things done somewhen then – I can also first check how much I understand of the things that have to be changed then.

PR #452, opened 2025-04-17
Remove leftover from merge

blegat:


kellertuer:
Oh! How did I miss that!? Thanks for the fix.

PR #453, opened 2025-04-18
Use scaled objective

blegat:
All the commits but last are the same compared to https://github.com/JuliaManifolds/Manopt.jl/pull/448.
In https://github.com/JuliaManifolds/Manopt.jl/pull/448, we scale the objective in case of `MAX_SENSE` in the conversion from MOI.ScalarNonlinearFunction to `Manopt.AbstractManifoldObjective`. The issue is that if the user directly sets a `Manopt.AbstractManifoldObjective` then the `MAX_SENSE` is ignored then. This is why we rather need to do `-objective` as in the last commit of that branch but as you can see it is failing.
Any idea what's the issue ?

### DIFF
--- a/ext/ManoptJuMPExt.jl
+++ b/ext/ManoptJuMPExt.jl
@@ -295,19 +295,12 @@ function MOI.set(model::Optimizer, attr::MOI.ObjectiveFunction, func::MOI.Abstra
     evaluator = MOI.Nonlinear.Evaluator(nlp_model, backend, vars)
     MOI.initialize(evaluator, [:Grad])
     function eval_f_cb(M, x)
-        val = MOI.eval_objective(evaluator, JuMP.vectorize(x, _shape(model.manifold)))
-        if model.sense == MOI.MAX_SENSE
-            val = -val
-        end
-        return val
+        return MOI.eval_objective(evaluator, JuMP.vectorize(x, _shape(model.manifold)))
     end
     function eval_grad_f_cb(M, X)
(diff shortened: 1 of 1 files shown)


kellertuer:
Oh I was just merging yours (after reading it carefully and missed the new one here.

Hm, I have no clue why then the optimization should to work and why the test are failing, because I still struggle a lot with all the undocumented internals from MOI. Every now and then I try to read up on them but always end up in code-archeology instead of reading docs and then I get lost.

So for now I can not say why that is happening because the scaled objective literally just rescales cost and grad

kellertuer:
I debugged this a bit and I have no clue why that is not working. As I wrote, the scaled objective does nothing else than scaling the objective, so your first approach to do that manually should be 100% equivalent on the Manopt side.

Then I started with #454 but have to read much much more about JuMP first I fear. Worked for an hour to just write something about model and variable, got a bit lost on the objective, since I thought the current code I use was introduced by now but errors. Probably my main concceptual problem is still that I _only_ grew up with nonlinear objectives – so I probably miss the whole reasoning behind objective being all this magic anyways.

kellertuer:
Thanks for tracking it down, your example helped :) And nice that the scaled objective is of use now.

blegat:
Not sure I need to add a changelog entry since it fits in the one of https://github.com/JuliaManifolds/Manopt.jl/pull/448

kellertuer:
No this is fine :) it is a reminder and you followed that already.

PR #454, opened 2025-04-18
A JuMP tutorial

### LINKED ISSUE #313: Short tutorial on how to use Manopt in/with JuMP
See _https://github.com/JuliaManifolds/Manopt.jl/pull/264#pullrequestreview-1713936091

kellertuer:
closes #313

# 📋
* [ ] understand how to use `@objective`
* [ ] understand why we set the attribute to a state and how that magically selects a solver
* [ ] write a bit about `optimise!` and `summary`.


tutorials/UsingJuMP.qmd
```diff
@@ -0,0 +1,60 @@
...
+
+```{julia}
+model = Model(Manopt.JuMP_Optimizer)
+```
+
+Next we add a [JuMP variable](https://jump.dev/JuMP.jl/stable/manual/variables/),
+where we specify the manifold using the `in` keyword, `start=` cares for the initialisation
+
+```{julia}
+@variable(model, p[1:3] in M, start=p[i])
```
blegat, line 41:
```suggestion
@variable(model, p[i=1:3] in M, start=p[i])
```

kellertuer: Ah, that I just copied from the old snipped in extension.md; so maybe that was improved/changed since then.

tutorials/UsingJuMP.qmd
```diff
@@ -0,0 +1,101 @@
...
+## Variant I: Specify the
+
+One can specify a cost in the embedding and get an approximate gradient in the embedding that is “converted” into a Riemannian one, see [AD](AutomaticDifferentiation.md).
+
+```{julia}
+f(p) = sum(1 / (2 * n) * norm.(Ref(p) - data) .^ 2)
+grad_f(p) = sum(1 / n * grad_distance.(Ref(M), data, Ref(p)));
+# TODO: how to set the objective to an implemented function ?
+# TODO 100% unsure how to set a gradient since that seems to be nowhere documented ?!
+@objective(model, f)
```
kellertuer, line 55: Yo me, two things are completely unclear even after reading a lot about JuMP:

1. If I have a cost function like f in line 51 – how do I set that as the objective?
2. If I have a gradient in the embedding how do I set a gradient in JuMP?

JuMP Seems to have super many magical commands to set stuff, but not explicit cost/grad things? @blegat

tutorials/UsingJuMP.qmd
```diff
@@ -0,0 +1,101 @@
...
+
+The [objective](https://jump.dev/JuMP.jl/stable/manual/objective/) is set for example using our [Objectives](objective.md).
+
+```{julia}
+model2 = Model(Manopt.JuMP_Optimizer)
+@variable(model2, p[i=1:3] in M, start = p[i])
+f(M, p) = sum(1 / (2 * n) * distance.(Ref(M), Ref(p), data) .^ 2)
+grad_f(M, p) = sum(1 / n * grad_distance.(Ref(M), data, Ref(p)));
+# TODO: Does not yet work, but I thought it should by now?
+@objective(model2, Min, Manopt.ManifoldGradientObjective(f, grad_f))
```
kellertuer, line 87: This might work once #455 is covered by tests and merged, still not sure.

tutorials/UsingJuMP.qmd
blegat: You should do `@objective(model, Min, Manopt.ManifoldGradientObjective(f, grad_f))` but it needs https://github.com/JuliaManifolds/Manopt.jl/pull/455/

kellertuer: Ah ok, I thought there was an alternative (the old?) way as well, but in the old code I never found how to provide a gradient.

Sure for the new way we need that PR as well.

blegat: Ah sorry, I missed the context, this is the way:
https://jump.dev/JuMP.jl/stable/moi/submodules/Nonlinear/overview/#User-defined-operators

kellertuer: Ah nice, then I'll try that one as well – would be nice to have both variants working here in the tutorial :)

blegat: I agree

blegat: If the Euclidean objective as a simple algebraic form, like `sum(x)`, you can also just use `@objective(model, Min, sum(x))` and not have to defined the gradient yourself

kellertuer:
At least to finish the day I got a first version running, where a first draft of using a full objective seems to work

https://manoptjl.org/previews/PR454/tutorials/UsingJuMP/

blegat:
What is the expression of the embedding objective ?

kellertuer:
The `f` from the example above (or the one with distances for the first working example, which would be more precise I think).

blegat:
Then don't use a user-defined function, let JuMP's automatic differentiation do the work:
```julia
@objective(model, Min, sum(sum((p - d).^2) for d in data)) / 2 * n)
```

kellertuer:
Hm, that is again all a bit tricky, Why do I have to use yours and not

```
f2(p) = sum(1 / (2 * n) * norm.(Ref(p) .- data) .^ 2)
@objective(model2, Min, f2(p))
```

? That one errors with not-supported.

blegat:
JuMP's AD currently does not support vectors, I'm working on it. This is the first item of our roadmap: https://jump.dev/JuMP.jl/stable/developers/roadmap/#Development-roadmap
That's why `norm` is not supported.

> so when that is the cost used, it is fully wrong since data is about 100 points on the sphere.

The objective is a sum of 100 different quadratic expressions. Because these are quadratic expressions, JuMP will just simplify it by actually doing this sum and simplify. If you don't want JuMP to simplify affine and quadratic expressions and want the expression tree before any simplifications, we have https://jump.dev/JuMP.jl/stable/api/JuMP/#@force_nonlinear
But I don't think that's causing any issues here.

Here is a reduction of the issue that is completely JuMP-independent.
```julia
using Manopt, Manifolds, LinearAlgebra
manifold = Sphere(2)
function eval_f_cb(M, p)
    return sum(sum((p - d).^2) for d in data) / (2 * n)
end
function eval_grad_f_cb(M, p)
    grad_f = 2sum((p - d) for d in data) / (2 * n)
    return Manopt.riemannian_gradient(manifold, p, grad_f)
end
objective = Manopt.ManifoldGradientObjective(eval_f_cb, eval_grad_f_cb)
dmgo = decorate_objective!(manifold, objective)
problem = DefaultManoptProblem(manifold, dmgo)
s = Manopt.GradientDescentState(manifold; p=p0)
state = decorate_state!(s)
solve!(problem, state)
solution = Manopt.get_solver_return(get_objective(problem), state)
norm(solution)
```
Running this, I get as value of `solution`:
```
3-element Vector{Float64}:
 0.5883892594536084
 0.00448648746991528
 0.6216157176252621
```
which has norm
```
0.855937000957987
```
which is indeed not on the sphere. However, as this is pure-Manopt and no JuMP, this is pure magical for me so the floor is yours ;)

kellertuer:
Ok, thanks for narrowing it down, since it is now a Manopt problem , I will check what is going wrong there then. I was thinking first, that JuMP maybe somehow did not do the gradient conversion, but I had no clue how to dive into that code.

For the pure manopt version, I know how to get more insight and will try to check what is happening there. At first glance, the code looks fine, so maybe the initial data is wrong or the approximation of the original cost (the one with distances on the Manifold) has really to be used and the norm approximation idea can not be used here.

mateuszbaran:
One of the top causes of losing normalization in optimization on a sphere in Manopt is accumulation of numerical errors. The default exponential map isn't particularly stable and this is a common result. Projecting the iterate back into the manifold is usually a good solution but currently figuring out when it needs to be turned on is left to the user.

@kellertuer maybe it would be a good idea to add a check at the end of optimization if the result lies on the manifold and if it doesn't suggest to the user using projection?

blegat:
> For the warning, I am not yet sure where I should put that then in a generic JuMP interface

This should be changed to return `MOI.INFEASIBLE_POINT` if the point is not in the manifold:
https://github.com/JuliaManifolds/Manopt.jl/blob/013feadf95736639cdedb1bff2c4f1723f7dfe64/ext/ManoptJuMPExt.jl#L462-L468

If you want to give custom message to the user like advice to set a projection, you can just print it like with `@warn` or you could also include it in the `RawStatusString`:
https://github.com/JuliaManifolds/Manopt.jl/blob/013feadf95736639cdedb1bff2c4f1723f7dfe64/ext/ManoptJuMPExt.jl#L484-L488

kellertuer:
Hm I have to check and first find out why we are leaving the manifold, I did not expect that at all.

kellertuer:
Hm, jeah, the data is completely random and wrong -.-

The current code just stays at the initial point, but that is already not a point on S.

blegat:
The initial point is `p0 = 1 / sqrt(2) * [1.0, 0.0, 1.0]` which **is** in the manifold

kellertuer:
So. Here is a full example with data on the manifold I confused myself with the statement above: one has to copy data[1] to not modify the data for p0,
then this code

```
using JuMP, Manopt, Manifolds, Random
using ManifoldDiff: grad_distance, prox_distance
Random.seed!(42);

n = 100
σ = π / 8
manifold = Sphere(2)
p = 1 / sqrt(2) * [1.0, 0.0, 1.0]
data = [exp(manifold, p, σ * rand(manifold; vector_at=p)) for i in 1:n];

p0 = copy(manifold, data[1]);

function eval_f_cb(M, p)
    return sum(sum((p - d) .^ 2) for d in data) / (2 * n)
end
function eval_grad_f_cb(M, p)
    grad_f = sum((p - d) for d in data) / n
    X = Manopt.riemannian_gradient(M, p, grad_f)
    X2 = project(M, p, grad_f)
    println("EGRad:", grad_f, "\nGrad: ", X , "\n point? ", is_point(M, p; error=:info), " | vector ? ", is_vector(M, p, X, false; error=:warn), "vector2 ? ", is_vector(M, p, X2, false; error=:warn))
    return X2
end
objective = Manopt.ManifoldGradientObjective(eval_f_cb, eval_grad_f_cb)
problem = DefaultManoptProblem(manifold, objective)
s = Manopt.GradientDescentState(manifold; p=p0)
state = decorate_state!(s; debug=[:Iteration, :Cost, :GradientNorm, :Stepsize, "\n"])
solve!(problem, state)
solution = Manopt.get_solver_return(state)
is_point(manifold, solution; error=:info)
```

Fails _even if_ we use projection (X2) instead of `riemannian_gradient` – which actually also should and does do the same.
That means that the gradient in theory since we project it on the tangent space. It is always a valid Riemannian gradient so that exp alsways should work fine.

However at iteration 9 the gradient is 10e-16 off (no clue why since we project) which accumulates and at iteration 44 it is 10e-10 off and that is enough that the points “wander off”.

I have exactly zero clue why this is happening, the code above prints a lot of information why this is happening, but I am getting tired by now after an hour of looking at results.

We should maybe abandon this idea of providing f if this fails even for the most simple example I can come up with? I have no clue how to fix this. Providing Riemannian cost/grad is not so useful for non-experienced users, but maybe the only thing Manopt currently can offer.
If that is too complicated for JuMP, we can also slowly abandon this idea of the extension and not provide it in JuMP. Maybe Manopt is – overall sadly – still not yet of the quality that I had maybe wished it would have reached until now.
If we abandon this, I Can also close this PR and all others relayed to JuMP

Sorry for the inconvenience caused.

mateuszbaran:
This is not an issue with the JuMP interface and Manopt works nearly completely fine. You just didn't anticipate the influence of compounding numerical errors. This doesn't expose any fundamental errors. I don't think we should abandon anything. This is even a good pedagogical example of numerical issues, though it might be good to improve default Manopt's robustness.

kellertuer:
No. That retraction only exists on a few manifolds, as does `exp` (or in that long form `ExponentialRetraction()` for the keyword.

The problem is

**in theory** the exponential map, which is by default used on the sphere is the exact solution one should do as a gradient step (“the thing that replaces + on manifolds”). **In practice** exp on the sphere is a bit numerically instable. But this is very specific to the sphere.
The `ProjectionRetraction()` in long is “do + and project onto the manifold” instead, which approximates the exponential map (up to first order). In theory it is a bit inexact in all steps. In practice it is numerically more stable.

What one should do for any other manifold is not so easy to say, because similar to here, it also heavily depends on which cost and gradient you take.

mateuszbaran:
We could change the default retraction on sphere so that it does `exp` and then projects the point for numerical stability. It would still properly reflect the geometry of the sphere while it would be more numerically stable.

kellertuer:
We would need a new retraction type for that probably.

kellertuer:
Hm, Now that we in principle have the stabilised exponential map after Mateusz update (thanks!) I again got lost in the docs of JuMP with one simple task

Instead of a `s = Manopt.GradientDescentState(manifold; p=p0)
I want a `s = Manopt.GradientDescentState(manifold; p=p0, retraction_method=ManifoldsBase.StabilizedRetraction(ExponentialRetraction()))`

to use within JuMP. I could not find anything how to change things like that? It seems there is only always the plain `GradientDescentState` with defaults possible? The only thing I could find is how to maybe exchange the StateType...

blegat:
Given
https://github.com/JuliaManifolds/Manopt.jl/blob/8e754a367bdd1dc6bedbd4a91588e4e7138d953a/ext/ManoptJuMPExt.jl#L374-L377
Maybe try
```julia
set_attribute(model, "retraction_method", ManifoldsBase.StabilizedRetraction(ExponentialRetraction())
```
Let me know if that works

kellertuer:
Will try. One should maybe document that somewhere if that works, because not everyone is a code-acheologist.

blegat:
I agree, we could also add a test

kellertuer:
Hm added that line and even committed it. The summary is, that that line (yes adapted to `model2` since it is the second example) does exactly nothing, the second model does not work.

But I tried to find where the state is now hidden to look whether the retraction is set, but I have no clue where that is, it seems to be somewhere very secretly hidden in the `model2`. I think the `set_attribute` does not work.

blegat:
I tried it locally and it works. Could you fix it ?

kellertuer:
I am not sure it works, I tried a few things I do not understand and then decided to only continue on Friday and left to sit a bit at the Fjord.

The current version https://manoptjl.org/previews/PR454/tutorials/UsingJuMP/ does something different in the embedding or converge differently, but the cost also is a bit different from Variant 1 (geodesic distance) to 2 (norm), so I am not sure.
One thing I did change along the way was that the `/(2n)` division was not inside the objective, but outside /so it divided the returned nothing by 2n). That might have been a larger problem, since that would lead to super large gradients, that would wrap around several times on the sphere and hence merely resemble a random walk.

blegat:
Yes, some clear docs detailing the behavior with `set_attribute` would be very useful

kellertuer:
Sure, I just have to understand that a bit better before I write about it.

blegat:
The behavior is as follows: we keep track of the key, value pairs set as
```julia
set_attribute(model, "key", value)
```
If there is a value assigned to the key `"descent_state_type"` then we use it as descent state type, otherwise we fall back to using `Manopt.GradientDescentState`.
The rest of the key-value pairs are given as keyword arguments to the constructor of that descent state.

kellertuer:
I think this tutorial is finished. I documented the few things I do understand on the extensions page. In the long run, it would be nice to have more doc strings to the functions defined in the extension file – all of them seem to be “magic functions” somewhere in the structures/design of JuMP whose implications are widely unknown to me. 
So these should either be documented in detail in an extension that lives here or move to an extension in JuMP (when Manopt is loaded).

tutorials/UseManoptWithinJuMP.qmd
```diff
@@ -104,14 +104,15 @@ TODO: Text
 
 One disadvantage of this formulation and leaving the AD to JuMP is, that the
 exponential map computations accumulate an error over time. We use a stabilized version
-to compensate for that by setting
+to compensate for that by setting the retraction to the stabolozed version of the exponential map, that for numerical stability performs a projection onto the manifold as well.
```
blegat, line 107:
```suggestion
to compensate for that by setting the retraction to the stabilized version of the exponential map, that for numerical stability performs a projection onto the manifold as well.
```

blegat:
No worry. The last commit on master as well as https://github.com/JuliaManifolds/Manopt.jl/pull/469 are unrelated to non-array. You can merge https://github.com/JuliaManifolds/Manopt.jl/pull/469

blegat:
https://github.com/JuliaManifolds/Manopt.jl/pull/470, is also unrelated to non-array

kellertuer:
Since for now I have to release a bug fix (for a student of mine) I reverted the current master-commit/PR, will come back to that later.

kellertuer:
Maybe it would be best to merge all 3 here to not have the situation as now, that we had to refer something on master.

blegat:
Sure, no objection

kellertuer:
Hm, I wlways accepted the incoming changes, but something still went wrong. Will have to check some time.

kellertuer:
Hm, though I accepted all incoming changes, the new version still (yet / again?) allocates :(

docs/src/extensions/JuMP.md
```diff
@@ -0,0 +1,55 @@
...
+## Setting s solver and its options
+
+A main thing to choose is the solver to use. By default this is set to the [`GradientDescentState`](@ref). To change the solver you can set it with `set_attribute`. For exmple to use the [`quasi_Newton`](@ref) instead, use
+
+```{julia}
+model =  Model(Manopt.JuMP_Optimizer)
+set_attribute(model, "descent_state_type", Manopt.QuasiNewtonState)
+```
+
+Any of the keywords of the solver you can set with `set_attribute)model, keyword, value)` for example to change the retraction to use, call
```
blegat, line 14:
```suggestion
Any of the keywords of the solver you can set with `set_attribute(model, keyword, value)` for example to change the retraction to use, call
```

docs/src/extensions/JuMP.md
```diff
@@ -0,0 +1,55 @@
...
+
+```{julia}
+set_attribute(model, "retraction_method", ManifoldsBase.ProjectionRetraction())
+```
+
+## Interface functions
+
+Several functions from the [Mathematical Optimization Interface](https://github.com/jump-dev/MOI.jl) (MOI) are
```
blegat, line 22:
```suggestion
Several functions from the [Mathematical Optimization Interface](https://github.com/jump-dev/MathOptInterface.jl) (MOI) are
```

ext/ManoptJuMPExt.jl
```diff
@@ -7,97 +7,173 @@ using ManifoldsBase
...
+"""
 struct RiemannianFunction{MO<:Manopt.AbstractManifoldObjective} <:
        MOI.AbstractScalarFunction
     func::MO
 end
 
+@doc """
+    JuMP.jump_function_type(::JuMP.AbstractModel, F::Type{<:RiemannianFunction})
+
+The JuMP function type of a function of type [`RiemannianFunction`](@ref) for any [`AbstractModel`](@extref JuMP.AbstractModel)
```
blegat, line 62: First time I see `extref` in action, looks very good !

ext/ManoptJuMPExt.jl
```diff
@@ -408,29 +615,70 @@ function JuMP.in_set_string(mime, set::ManifoldsBase.AbstractManifold)
...
+    return ManifoldVariable(p)
+end
+#
+#
+# TODO: Understand parameters here and document them
+function JuMP.add_variable(model::JuMP.AbstractModel, v::ManifoldVariable, name::String="")
+    # What does this do? This looks like pure magic to me.
+    #=
+    function _newvar(i)
+        vref = VariableRef(model)
```
blegat, line 659: You can remove that, this is specific to PolyJuMP

ext/ManoptJuMPExt.jl
```diff
@@ -408,29 +615,70 @@ function JuMP.in_set_string(mime, set::ManifoldsBase.AbstractManifold)
...
+            JuMP.set_binary(vref)
+        end
+        if v.integer
+            JuMP.set_integer(vref)
+        end
+        return vref
+    end
+    return MB.algebra_element(_newvar, v.p.polynomial_basis)
+    =#
+    return nothing
```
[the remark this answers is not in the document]
kellertuer: Thanks for the help, I looked around quite a while and could only find docs that said “you should not implement this, ask a developer” ;)

ext/ManoptJuMPExt.jl
kellertuer: I removed this. But I am still not sure, for this and the other two functions I just merely guessed what might be right without understanding anything.

Do these now work? How would one define a variable to be a `PoincareHalfPlanePoint` on some `Hyperbolic(2)`?

ext/ManoptJuMPExt.jl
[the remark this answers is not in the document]
kellertuer: Hm I am not sure what this was meant to be, but I was too fast in accepting it, since the `...` things of course do not make sense here.

Since I can not find any information about how one would define these – `build_variabl` and `add_variable` are both documented as “ask a developer” (not helpful). I do need help here, because I can not guess the right thing.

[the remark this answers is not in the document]
blegat: `_shape(v.manifold)`

blegat:
> But, if you always only work in small PRs and I got lost in the parallel PRs that contradicted each other when we tried that, then maybe just because our workflows do not fit, we could also agree to not do this or even remove the very preliminary existing JuMP support we have here, since that is really very limited.

Sorry, I didn't want to complicate things. I'm fine with leaving this open.

> Hm, with https://github.com/JuliaManifolds/Manopt.jl/pull/491 it seems we are again going for a parallel path. I would really by now propose to maybe let it be.

What do you mean by parallel path ? I just created another branch because I don't have write access to this one. That PR actually builds on the work you have done on this PR. By the way good job on figuring out how to make it work with the two extensions ^^

kellertuer:
In general I doo agree. that smaller PRs are better, but the three we had were so intertwined, that I neither knew which one to merge first nor what was there and several things were doubled already. That is maybe a sign that it can not be split that easily (to at least not by mere mortals like me).

> What do you mean by parallel path ? I just created another branch because I don't have write access to this one. That PR actually builds on the work you have done on this PR. By the way good job on figuring out how to make it work with the two extensions ^^

Sorry, I misunderstood the last PR, for me PRs are “tings that are worked towards being ready to get merged” – your newest PR will never be in that state because it has to be included here.

Figuring that out took me very very very very very long time. That also why I am a bit tired of this PR and idea in general. I slowly feel it is not worth the effort. I currently feel, this PR will not be finished this year. Then it has been a year of work for...what? I do not think anyone will ever use it. And if, there will be bugs, that I never can fix, because by now the code is already too clever for me.

kellertuer:
Array manifolds already work on master

This currently combines a) the tutorial and b) non-array manifolds. The tutorial is good to also both document/explain _and_ try the new stuff.

So for that reason we could try that you propose something for b) and if I understand it I can write a) but I would prefer to keep it in one.
I could make you developer here so you can push to this PR for the time of the PR if that helps in your workflow?

blegat:
Yes, if you make me dev, I can push directly to this branch :)

kellertuer:
Nice! I added you with write access and hope that is enough to upshot the `fix-313-a-jump-tutorial` branch (yeah by now not the best of names ;)).

kellertuer:
@blegat I added 3 questions were I am still stuck. I am also not yet able to “connect the dots” – or to make it more precise

1. we implement a bunch of functions but to me it is unclear how they interact.
2. to me it is unclear whether the distinction between point and tangent vector variables is taken into account, how, where and when – and whether it even _can_ be taken into account correctly, since that is crucial in the typed point case
3. to me it is unclear (see also 1) which “magic functions” are still required to complete the extension
4. to me it is most unclear, how the new/typed points would be used in practice, that is in the second part of the tutorial.

kellertuer:
I have another question. While carefully reading a lot of docs, I noticed that I do not understand where when calling

```
@variable(model, p[i=1:3] in M, start=p0[I])
```

the `in M` is processed to a set? Also would `p` just be a `HyperboloidPoint` for the non-array case? Or would that require a keyword argument? I saw both in my todays reading session, so I am also unsure here which one is “right” and how it would reach either `build_variable` (which is called first?) or `add_variable` (which is called second, but does it add anything anywhere?).

kellertuer:
I downgraded this to a draft and closed the corresponding issue #313. There is actually the example https://manoptjl.org/stable/extensions/#JuMP.jl (but also see below)

The main reason are discussions on slack, that by now discouraged me from continuing this because the idea would not fit JuMP well, for example we would need `p in M` without parameters for all typed ones. For the array ones `p[1:3] in Sphere(2)`, which is the current way it has to be specified is misleading, since the parameter space of the sphere is just 2-dimensions.

From all the questions and things that do not work (mainly because I really can not get to understand JuMP). I am leaning towards even removing the existing extension after the discussion today with @odow, since I was convinced therein that Manopt simply does not fit into JuMP.

blegat:
> Also would `p` just be a `HyperboloidPoint` for the non-array case? Or would that require a keyword argument?

The type of `p` is picked as the same type of `p0` so no keyword argument needed

> some `error_fn`? What is that? How is that used? I can not find documentation about that.

It's like `Base.error` but adds more info for the user to be able to see from which macro call this came from

> Maybe I'll pause here again, I feel like running in complete darkness, when the things I would like to use and understand are not documented and feel again very lost in code archeology or when doc strings say, I should probably not implement something.

I think it's safer if I do this part then, what would be helpful for me is a complete simple example using non-array point in pure Manopt syntax :)

kellertuer:
The simplest example is to stick with the Riemannian center of mass as in the intro tutorial, maybe just on Hyperbolic. space. There arrays would be the hyperboloid representation and we would do that on the Poincaré half plane. I can try to provide an MWE for that, sure.

blegat:
Yes, that would be useful or just a pointer to an test or something like that

PR #462, opened 2025-04-26
Fixes printing of JuMP models containg Manopt solver

blegat:
Fixes the error of https://github.com/JuliaManifolds/Manopt.jl/pull/454#issuecomment-2831358678


kellertuer (approved):
Nice! Thanks for the quick fix.

PR #465, opened 2025-05-02
Switch to using QuartoNotebookRunner

kellertuer:
Which removes the necessity for one of the caches and makes rendering the docs independent of Python \o/


kellertuer:
This works great! We should update all our other docs the same way as here! Will try to find time for that.

PR #466, opened 2025-05-03
Jump vectorize (to work on non-array points)

### LINKED ISSUE #274: Non-array manifold with JuMP/MOI
In JuMP, we only define array-shaped points in manifold but as mentioned, e.g., in https://github.com/JuliaManifolds/Manopt.jl/pull/264#issuecomment-1600473005, for `FixedRankMatrices` they could also be represented by their SVD.
We should make this work:
```julia
@variable(model, X in FixedRankMatrices(m, n, r))
@objective(model, Min, RiemannianFirstOrderFunction(X -> ..., (g, X) -> ...)))
```
Any good example of `m`, `n`, `r` and riemannian objective and gradient we could use as test for this ?

Requires

- [x] https://github.com/JuliaManifolds/Manopt.jl/issues/273

kellertuer:
This is an approach to fix #274, but a very early stage, since I can not find any information about the functions to implement and how they are supposed to work. Any help appreciated.

For now this just documents may state of knowledge and some guesswork.


ext/ManoptJuMPExt.jl
```diff
@@ -456,17 +485,17 @@ function JuMP.in_set_string(mime, set::ManifoldsBase.AbstractManifold)
...
-Build a `JuMP.VariablesConstrainedOnCreation` object containing variables
+Build a [`JuMP.VariablesConstrainedOnCreation`](@extref) object containing variables
 and the [`Manopt.JuMP_VectorizedManifold`](@ref) in which they should belong as well as the
 `shape` that can be used to go from the vectorized MOI representation to the
-shape of the manifold, that is, [`Manopt.JuMP_ArrayShape`](@ref).
+shape of the manifold, that is, [`Manopt.ManifoldDataShape`](@ref).
 """
-function JuMP.build_variable(::Function, func, m::ManifoldsBase.AbstractManifold)
-    shape = _point_shape(m)
+function JuMP.build_variable(::Function, func, manifold::ManifoldsBase.AbstractManifold)
```
kellertuer, line 495: @blegat I have a question on this line, since I am currently fighting a bit with technical details in extensions: What is `func` here? A point on `manifold` in its original form? The `vectorize`(l. 498) seems to indicate so, but this is nowhere documented nor would the name make sense then.

kellertuer:
Hm I nearly lost another full day now. My problem now is:

I think we have to do an extension that requires `JuMP` _and_ `Manifolds` to be loaded, then I think I know what to do,

For now I tried two extensions
* one providing the base functionality and arrays based on `ManifoldsBase` (anyways available since it is a dependency of `Manopt`
* one that tries to extend that extension once `Manifolds` is loaded to also cover all the point/vector types.

But the problem is that the wrappers we have in the first extension, are not available in the second, so I would have to duplicate all of them (`VectorizedManifold`, maybe `Optimiser`, `ArrayShape`, ...).

And I slowly think for the effect this work will have (probably strictly upper bounded by zero) – I am no longer sure it is worth that I spend weeks and weeks fighting both JuMP and its undocumented functions and ask a million of questions and now also the internals of extensions.
No one will use this extension I fear, so all those hours I am trying to figure out something by now are mainly wasted time :(

ext/ManoptJuMPExt.jl
blegat: Good point, that's a bit tricky. Here, we let the user create the right shape. For instance here:
https://github.com/JuliaManifolds/Manopt.jl/blob/56ba724b31f0e13b5474906ab59b533ed4f4d821/test/MOI_wrapper.jl#L90
The user specifies that it's a 2x2 matrix so `func` is a 2x2 matrix.
If you need the variables to be of a shape that is not an array that won't work.
Then we'll have to go the route of PolyJuMP.jl were we have custom non-array shapes but that will use some JuMP dark magics as you like to call it so I'll need to make a PR :)

blegat:
That's also why I started with just array shapes and I suggested to way for users to request it to work for additional manifolds.
If there is a nice error message for the manifolds for which it's not implemented there is no rush.
Once we have a concrete use case, we can interact to make sure I understand the Manifold side and you understand to JuMP side and find a simple solution before we spend hours going through each other's code.
Directly targeting manifolds that have 1) non-array shapes and 2) different possible shapes is maybe too ambitious

ext/ManoptJuMPExt.jl
```diff
@@ -235,20 +236,20 @@ Return whether `vi` is a valid variable index.
...
-           1 <= vi.value <= MOI.dimension(VectorizedManifold(model.manifold))
+           1 <= vi.value <= MOI.dimension(ManifoldsBaseSet(model.manifold))
 end
 
 """
     MOI.get(model::Optimizer, ::MOI.NumberOfVariables)
 
 Return the number of variables added in the model, this corresponds
-to the [`MOI.dimension`](@ref) of the [`Manopt.JuMP_VectorizedManifold`](@ref).
+to the [`MOI.dimension`](@ref) of the ManifoldsBaseSet`.
```
blegat, line 246:
```suggestion
to the [`MOI.dimension`](@ref) of the `ManifoldsBaseSet`.
```

kellertuer:
> That's also why I started with just array shapes and I suggested to way for users to request it to work for additional manifolds.
If there is a nice error message for the manifolds for which it's not implemented there is no rush.

For now my honest opinion is, the non-array manifolds are basically only doable with a lot a lot a lot of code, if at all.
We would have to copy _all_ of the current extensions code in the second new extension and have to find a way both work in parallel (no clue how to get that to work).

So for now, I am not sure this is doable, and no that also means there is zero error messages for those cases until now.

kellertuer:
I think it is time to give up. I do not see how non-array cases can be done in reasonable time. I thought it might have been a nice feature, but since I also do see that no one will ever use it, maybe, if it is so complicated as it turns out do be, it is maybe time to really just leave it be.

blegat:
So the issue is that the JuMP extension would have to define methods for types defined in Manifolds but it's not a dependency of Manopt right ? Yes I think that's currently a limitation of the extension mechanisms. I don't think you can create an extension that needs two packages.
The only fix I would see is to define a vectorization API in ManifoldsBase that is independent of JuMP and then use it in this JuMP extension.

blegat:
When you create a JuMP variable that is for example an `SVDMPoint`, you would need to hold store the JuMP variable as values of the points. The good thing is that you do not require the entry to be subtypes of `Number`  so it just works:
```julia
julia> model = Model();

julia> @variable(model, x[1:3])
3-element Vector{VariableRef}:
 x[1]
 x[2]
 x[3]

julia> y = SVDMPoint([x[1];;], [x[2]], [x[3];;])
SVDMPoint{Matrix{VariableRef}, Vector{VariableRef}, Matrix{VariableRef}}
U factor:
 1×1 Matrix{VariableRef}:
  x[1]
singular values:
 1-element Vector{VariableRef}:
  x[2]
Vt factor:
 1×1 Matrix{VariableRef}:
  x[3]
```

kellertuer:
The “rush” was only that I wanted to close some of the open points that bring us towards a Manopt 1.0 – but maybe Manopt is also not yet ready for that.

blegat:
Sure I'd also be happy to close it, I just meant I'm willing to help if you had a vectorization API to Manifolds but I'm not adding pressure ;)

kellertuer:
I am quite unsure how such an API should look like and I do not see how that would come along. The main danger I see is, that the ”vector form“ that API would be able to provide would nearly never (Euclidean space is an exception) agree with the DoF (manifold_dimension), so it would probably cause more confusion than it is useful.

I'll put this aside for now. Will add the in-place things in the tutorial PR from master and fix them a bit, but leave the non-array parts until either the thread linked above has a solution or some other magic happens.

kellertuer:
I noticed another problem after the discussions from yesterday: There is no way to declare a variable in JuMP to be anything else then vector, so while `Hyperbolic` has 4 different point types it can work with, in `@variable(model, p in Hyperbolic(2))` I can never specify `p` to be anything else then array/vector.

So even _if_ there is some day any chance to have a better backend and reuse the type and such (all the topics discussed yesterday), the frontend of JuMP is basically incompatible with Manopt/Manifolds.

So let's maybe really end all this, close this, and the issue – since I do no longer see how any non-array manifold should ever be able to be covered, neither in the extension code nor in the user-facing part of JuMP.

blegat:
No you can, that's what I meant in https://github.com/JuliaManifolds/Manopt.jl/pull/466#discussion_r2077814559 and https://github.com/JuliaManifolds/Manopt.jl/pull/466#issuecomment-2859187131
When the user write `@variable(model, p in Hyperbolic(2))`, you can have `p` be of the point type you want.
Look for example at this example : https://jump.dev/SumOfSquares.jl/stable/generated/Systems%20and%20Control/barrier_certificate/
When I write `@variable(model, B, Poly(monos))`, `B` has the type of a polynomial which is not an array.
But then, overloading `build_variable` alone is not sufficient, you also need to overload `add_variable`. Here is how it's done:
https://github.com/jump-dev/PolyJuMP.jl/blob/b28143bb2f68ff31b3c34d129ff2f403eb1752b4/src/variable.jl#L89-L122
But again, I can do this part if you have a concept of vectorization for manifolds ;)

> The main danger I see is, that the ”vector form“ that API would be able to provide would nearly never (Euclidean space is an exception) agree with the DoF (manifold_dimension), so it would probably cause more confusion than it is useful.

Yes, it has to be clear that the length of the parametrization is not the dimension of the manifold. Using the word length instead of dimension is important to avoid confusion

kellertuer:
Ah, ok, I think I understand, so in the interface level it can be done, but in a few areas in the deep details of JuMP (MOI? I never can keep those two apart) it has to be verctorized. 

Well, For now still the technical details from the discourse thread I opened yesterday also stand in the way, so I think it is still at the point where I have to put it aside, because I am out of ideas what to do.

blegat:
Yes, at the JuMP level, you can have any shape you want, it's only when going through MOI than you need to be vectorized

blegat:
Sounds good let's close this for now then

ext/ManoptJuMPExt.jl
kellertuer: I had a super small idea but looking at your explanation here again, I got lost in how this is supposed to ever work without `func` being an array, mainly because the magic you refer to in `PolyJuMP` also is 100% undocumented, so I have still no chance to understand how `func` is supposed to work nor which type it is. 

AllI could find is https://jump.dev/JuMP.jl/stable/api/JuMP/#JuMP.VariableInfo and that helps some new person like me with the feeling that my knowledge on JuMP is strictly upper bounded by zero – exactly no info how the “may be used” maybe can actually be used

Even though for now I just mainly look towards fixing the tutorial – I am by now also lost on that PR – so basically on all open PRs.

PR #468, opened 2025-05-06
Inplace version of vectorize and reshape

blegat:
Here is a stab at https://github.com/JuliaManifolds/Manopt.jl/issues/274#issuecomment-2855414411

### DIFF
--- a/ext/ManoptJuMPExt.jl
+++ b/ext/ManoptJuMPExt.jl
@@ -391,12 +391,29 @@ function MOI.optimize!(model::Optimizer)
     return nothing
 end
 
+"""
+    struct ArrayShape{N} <: JuMP.AbstractShape
+
+Return a [`JuMP.AbstractShape`](@ref) that can be used to vectorize points
+and tangent vectors of the manifold and reshape the vectorized representation
+to the original objects of the manifold.
+"""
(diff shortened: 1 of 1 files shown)


ext/ManoptJuMPExt.jl
```diff
@@ -26,7 +26,7 @@ As the MOI variables are real, this means if the [`representation_size`](@extref
 yields (in product) `n`, this refers to the vectorized point / tangent vector  from (a subset of ``ℝ^n``).
 """
 function MOI.dimension(set::VectorizedManifold)
-    return prod(ManifoldsBase.representation_size(set.manifold))
+    return length(_point_shape(set.manifold))
```
kellertuer, line 29: Is it (by the way) ok that the actual degrees of freedom are less than this?

For example for `M = Sphere(2)` this dimension is 3 while the manifold-dimension (actual degrees of freedom) are 2.

blegat: Yes, this is not the dimension in the manifold sense, it's the length of the vectorization

kellertuer: Ok, because that sometimes caught us in DiffEq, that they expected DoF and vector dimension to be the same ;)

PR #469, opened 2025-05-09
Allocation-free cost and gradient callback for JuMP interface

blegat:
I wanted to have `_EmbeddingObjective <: AbstractManifoldGradientObjective` so that I don't even have to create a `ManifoldGradientObjective` as the `_EmbeddingObjective` would serve as the objective but I saw that I needed to specify `TC` and `TG` which are the types of the cost and gradient function which didn't really apply if I just wanted to implement `get_cost` and `get_gradient!` directly without creating closures.
So instead, I created (allocation-free !) closures that I gave to `ManifoldGradientObjective` which I think is fine but I left a comment to explain the design choice.
This also moves some code out of the `MOI.set` method which makes it easier to parse both for us and for the compilers :)
I also added tests to make sure it stays allocation-free.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,12 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.16] TBD
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


kellertuer:
That looks nice. Interesting, that you write the wrapper “the other way around” compared to 
https://manoptjl.org/stable/plans/objective/#Manopt.EmbeddedManifoldObjective

kellertuer:
manually merged into #454.

PR #470, opened 2025-05-09
Alloc stiefel

blegat:
I wanted to test allocations in the Stiefel example as well but it is allocating.
The allocation does not come from the JuMP extension but form the Manopt part. here is a pure-Manopt reproducer:
```julia
using Manifolds, ManifoldDiff
function alloc()
    M = Stiefel(2, 2)
    embedding_tangent = [-2.0 2.0; 2.0 -2.0]
    gradient = [-2.0 2.0; 2.0 -2.0]
    p = [0.0 0.0; 0.0 0.0]
    ManifoldDiff.riemannian_gradient!(M, gradient, p, embedding_tangent)
    @time ManifoldDiff.riemannian_gradient!(M, gradient, p, embedding_tangent)
end
alloc()
```
This gives
```
  0.000001 seconds (4 allocations: 224 bytes)
2×2 Matrix{Float64}:
 -2.0   2.0
  2.0  -2.0
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,12 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.16] TBD
+
+### Fixed
+
+* Fixed allocations in the callbacks of the JuMP interface so that the solver can query the cost and gradient without allocating.
+
 ## [0.5.15] 2025-05-06
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


blegat:
> Hm, was this meant to be an issue not a PR? the changes look nearly the same as in the previous one?

The last commit is the one that matters, I will rebase once https://github.com/JuliaManifolds/Manopt.jl/pull/469 is merged

mateuszbaran:
`riemannian_gradient!` allocates because it calls `project!` which allocates. If the gradient doesn't need projecting, you can safely just use `change_representer!`. Making `project!` on `Stiefel` non-allocating would be quite challenging.

kellertuer:
manually merged into #454.

PR #471, opened 2025-05-11
Fix retracts in LineSearches extension

kellertuer:
Thanks @elorannug for reporting.


kellertuer:
We either have to downgrade Documenter or wait a bit, see https://github.com/JuliaDocs/Documenter.jl/pull/2676#discussion_r2082477911

PR #479, opened 2025-06-05
Introduce proximal gradient method

hajg-ijk:


hajg-ijk:
> This seems to be a bit mixed up.
> 
> It seems to not have run the JuliaFormatter yet? It also seems to include already all changes from the other PR on the bundle method? That should only be merged here from master afterwards for best of cases, to keep those two PRs separate.

Yeah I'm not sure what happened. Maybe I just close this and try to fix things locally first. I also ran the formatter, so this is weird.

Issue #477, opened 2025-06-03
Ideas for meta stopping criteria


kellertuer:
Since stopping ceriteria are types we could introduce 2 meta criteria that are even nice to type

* [x] `StopWhenCriterionRepeated(sc,n)` or in short `sc × 3` to indicate to stop only after a criterion indicated to stop n times consecutively (name of the structure to be improved maybe)
* [ ] `StopWhenCriterionHitAfterIteration(sc,K) or in short maybe `sc > K` to only check/stop after iteration `K`.

The second one is easy to do, for the first one could collect the iterations when sc indicates to stop and check the length of it (and reset if it does not hit).

PR #480, opened 2025-06-09
Add callback (debug) and a repeated stopping criterion

kellertuer:
This PR adds two things

* A `StopWhenRepeated` stopping criterion to only stop when a stopping criterion indicated multiple times to stop (shortcut: `sc × n`)
* A `DebugCallback` to simplify to add a callback to arbitrary placed where debug can be issued already (at beginning or end of iteration, at beginning or end of algorithm)
  * a default variant for callbacks of the form `(problem, state, iteration) -> nothing` that have access to both problem and state
  * a simple variant `() -> nothing` that has to take care of tracking things itself. 
* handle `Function`s in the `debug=` keyword as the default variant (more complicated constructions like `Functors` have to use the `DebugCallback(f)` constructor)
* add a `callback=` keyword to easily  add the simple for to debug as well

# 📋
- [x] check documentation
- [x] check test coverage
- [x] add a (small) section to the debug tutorial
- [x] add a small explanation to the stopping criteria examples

### DIFF
--- a/src/plans/stopping_criterion.jl
+++ b/src/plans/stopping_criterion.jl
@@ -1436,15 +1436,16 @@ end
 function status_summary(sc::StopWhenCriterionWithIterationCondition)
     has_stopped = (sc.at_iteration >= 0)
     s = has_stopped ? "reached" : "not reached"
-    return "$(sc.comp) && $(sc.stopping_criterion): $(s)"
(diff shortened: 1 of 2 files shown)


kellertuer:
This now also resolves #477 since both ones mentioned there are now introduced here.

PR #481, opened 2025-06-12
Introduce the proximal gradient method

kellertuer:
Cf. https://arxiv.org/abs/2506.09775

This just needs a few final touches

* [x] discuss a new name for `get_cost_g`
* [x] discuss a new name for `get_cost_h`
* [x] check / write test coverage (without using `ManoptExamples`, since that would introduce a circular dependency)
* [x] carefully read the docs

### DIFF
--- a/src/plans/proximal_gradient_plan.jl
+++ b/src/plans/proximal_gradient_plan.jl
@@ -131,19 +131,27 @@ end
 @doc raw"""
     ProximalGradientNonsmoothCost{F, R, P}
 
-Stores the nonsmooth part of the proximal gradient objective and the stepsize parameter ``λ ∈ ℝ``.
+Stores the nonsmooth part `h` of the proximal gradient objective `f = g + h`, as well as the stepsize parameter ``λ ∈ ℝ``.
 
-This struct is also a functor `(M, p) -> v` that can be used as a cost function within a solver, primarily for solving the proximal map subproblem formulation in the proximal gradient method.
+This struct is also a functor `(M, q) -> v` that can be used as a cost function within a solver, primarily for solving the proximal map subproblem formulation in the proximal gradient method, which reads
+```math
+    \operatorname{prox}_{λ} h(p) = \operatorname{argmin}_{q \in \mathcal M} \left( h(q) + \frac{1}{2λ} d^2(q, p) \right)
+```
+Hence, the functor reads
+```math
+    (M, q) \mapsto h(q) + \frac{1}{2λ} d^2(q, p)
+```
+and `p` is the proximity point where the proximal map is evaluated, i.e. the argument `p` of the proximal map ` \operatorname{prox}_{λ} h`.
 
 ## Fields
-* `cost::F` - the nonsmooth part of the proximal gradient objective, i.e. the part of the objective whose proximal map is sought
+* `cost::F` - the nonsmooth part `h` of the proximal gradient objective, i.e. the part of the objective whose proximal map is sought
 * `λ::R` - the stepsize parameter for the proximal map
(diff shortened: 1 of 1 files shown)


src/plans/proximal_gradient_plan.jl
```diff
@@ -14,7 +14,7 @@ as well as ``\operatorname{grad} g`` and ``\operatorname{prox}_{λ} h``.
 # Fields
 
 * `cost`: the overall cost ``f = g + h``
-* `cost_g`: the smooth cost component ``g``
+* `cost_smooth`: the smooth cost component ``g``
 * `gradient_g!!`: the gradient ``\operatorname{grad} g``
 * `proximal_map_h!!`: the proximal map ``\operatorname{prox}_{λ} h``
```
kellertuer, line 19:
```suggestion
* `cost_smooth`: the smooth cost summand ``g``
* `gradient_smooth!!`: the gradient of the smooth cost summand``\operatorname{grad} g``
* `proximal_map_nonsmooth!!`: the proximal map ``\operatorname{prox}_{λ} h`` of the non smooth cost summand ``h``.
```

If we rename one to smooth, maybe we should consistently speak about smooth and non smooth summands (or components) of the cost?

src/plans/proximal_gradient_plan.jl
```diff
@@ -33,7 +33,7 @@ Generate the proximal gradient objective given the total cost `f = g + h`, smoot
 struct ManifoldProximalGradientObjective{E<:AbstractEvaluationType,TC,TG,TGG,TP} <:
        AbstractManifoldCostObjective{E,TC}
     cost::TC # f = g + h
-    cost_g::TG # smooth part
+    cost_smooth::TG # smooth part
     gradient_g!!::TGG
     proximal_map_h!!::TP
```
kellertuer, line 38:
```suggestion
    gradient_smooth!!::TGG
    proximal_map_nonsmooth!!::TP
```

src/plans/proximal_gradient_plan.jl
```diff
@@ -126,6 +126,87 @@ function get_proximal_map!(
...
+    ProximalGradientNonsmoothCost{F, R, P}
+
+Stores the nonsmooth part of the proximal gradient objective and the stepsize parameter ``λ ∈ ℝ``.
+
+This struct is also a functor `(M, p) -> v` that can be used as a cost function within a solver, primarily for solving the proximal map subproblem formulation in the proximal gradient method.
+
+## Fields
+* `cost::F` - the nonsmooth part of the proximal gradient objective, i.e. the part of the objective whose proximal map is sought
+* `λ::R` - the stepsize parameter for the proximal map
+* `proximity_point::P` - point where the proximal map is evaluated
```
kellertuer, line 141: Hm, I am not so sure what this refers to, the argument of the prox? One could maybe at least mention that in the docs, or even the full formulae above.

kellertuer (commented):
Code structure wise this looks great!

My comments are merely towards naming and doing that consistent (and sure, naming is hard!)

src/plans/proximal_gradient_plan.jl
kellertuer: I was first thinking about changing the name, but if we thoroughly document the cost as a math formula in the disc, we can explain what this variable refers to (i.e. this fixed point from the argument that we have in the dist to the variable p).

kellertuer:
Nice, I would now just add a bit more details to the docs of the prox-sub-cost-struct so that its fields become a bit clearer; and either refer to that in its subgrad or even write down the subgrad formula as well.

All the rest looks good :)

PR #484, opened 2025-07-04
Correctly pass starting point to MADS

hajg-ijk:
Pretty self explanatory. 
The call to `mesh_adaptive_direct_search(M::AbstractManifold, f, p=rand(M); kwargs...)` effectively ignored the user's choice of `p`, which is now fixed.


kellertuer:
Nice, I can add a Changelog Entry real quick, then this should be ready :)

PR #486, opened 2025-07-08
Debug warning for RPG backtracking

hajg-ijk:
Introduce a `DebugAction`,`DebugWarnIfBacktrackingNotConverged`, and a related `:WarnBacktracking` symbol for the debug dictionary. This is to be used in conjunction with the `ProximalGradientMethodBacktracking` stepsize to warn if the backtracking procedure of the `proximal_gradient_method` hit the stepsize length threshold without converging.
Also fixed a couple of typos in the docs.

### DIFF
--- a/src/plans/debug.jl
+++ b/src/plans/debug.jl
@@ -229,7 +229,7 @@ end
 # Special single ones
 #
 @doc raw"""
-    DebugCallback <: DDebugAction
+    DebugCallback <: DebugAction
 
 Debug for a simple callback function, mainly for compatibility to other solvers and if
 a user already has a callback function or functor available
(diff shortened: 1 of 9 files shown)


src/plans/debug.jl
```diff
@@ -1364,7 +1364,7 @@ function DebugActionFactory(d::Symbol)
     (d == :Feasibility) && return DebugFeasibility()
     (d == :Stepsize) && return DebugStepsize()
     (d == :Stop) && return DebugStoppingCriterion()
-    (d == :WarnBacktracking) && return DebugWarnIfBacktrackingNotConverged()
+    (d == :WarnBacktracking) && return DebugWarnIfStepsizeCollapsed()
```
kellertuer, line 1367:
```suggestion
    (d == :WarnStepsize) && return DebugWarnIfStepsizeCollapsed()
```

since we could use that for general step size rules, when they collapse, maybe also the symbol could reflect that? (Fix L 1344 then as well)

src/plans/proximal_gradient_plan.jl
```diff
@@ -517,7 +517,6 @@ function (s::ProximalGradientMethodBacktrackingStepsize)(
         # Reduce step size
         λ *= s.contraction_factor
     end
-    @warn "Backtracking stopped because the stepsize fell below the threshold $(s.stop_when_stepsize_less)."
     return λ
```
kellertuer, line 520: This line now seems to never be reached says code cov, is this true or is code cov incorrect?

kellertuer (approved):
lgtm, besides the one naming thingy and the one line CC I already commented on. Should this be released as its own version (probably tomorrow) or can it wait a bit?

hajg-ijk:
> lgtm, besides the one naming thingy and the one line CC I already commented on. Should this be released as its own version (probably tomorrow) or can it wait a bit?

It can definitely wait :)

kellertuer:
Nice. Then I‘ll release it with the next reasonable feature we add.

PR #489, opened 2025-07-10
Add warm start to CRPG backtracking

hajg-ijk:
Just like the title says. The user can now use a `warm_start_factor` keyword argument (which defaults to `1`) when initializing a `ProximalGradientMethodBacktracking` stepsize.
This can also wait to be released later of course :)


kellertuer (approved):
lgtm. Will merge that tomorrow.

PR #491, opened 2025-07-19
Proof of concept for non-array point

blegat:
This gives
```julia
julia> @variable(model, x in Hyperbolic(2), start = HyperboloidPoint([1, 2]))
HyperboloidPoint(VariableRef[_[5], _[6]])
```
What are the operations we could do on `x` that would result in a scalar objective function ?

### DIFF
--- a/ext/ManoptJuMPExt.jl
+++ b/ext/ManoptJuMPExt.jl
@@ -623,37 +623,31 @@ and the [`ManifoldSet`](@ref) in which they should belong as well as the
 shape of the manifold, that is, a [`ManifoldArrayShape`](@ref).
 """
 function JuMP.build_variable(::Function, array, M::ManifoldsBase.AbstractManifold)
-    shape = _shape(M)
-    return JuMP.VariablesConstrainedOnCreation(
-        JuMP.vectorize(array, shape), ManifoldSet(M), shape
-    )
+    if array isa JuMP.ScalarVariable
+        return ManifoldVariable(array.info.start, M)
+    else
+        shape = _shape(M)
+        return JuMP.VariablesConstrainedOnCreation(
+            JuMP.vectorize(array, shape), ManifoldSet(M), shape
+        )
+    end
 end
 
 #
 #
 # NonArrayPoints define own variable
 # TODO: Document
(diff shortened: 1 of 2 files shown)


kellertuer:
I do not understand the question, sorry. Maybe call the cost function?

ext/ManoptJuMPExt.jl
```diff
@@ -623,37 +623,31 @@ and the [`ManifoldSet`](@ref) in which they should belong as well as the
...
+    start::P
+    manifold::M
 end
-#
-#
+
 # TODO: Understand parameters here and document them
-function JuMP.add_variable(model::JuMP.AbstractModel, v::ManifoldVariable, name::String="")
-    return nothing
+function JuMP.add_variable(model::JuMP.AbstractModel, v::ManifoldVariable, ::String="")
```
kellertuer, line 646: Can we maybe have a it of documentation here? The official docs mainly state “please ask a developer”. If that code is here in Manopt _I_ am a developer and by now I tried for about half a year maybe to understand something – and I simply do not.

kellertuer:
This concept also can not work, since the point type is not “available” in `Manopt`. We can not just “rewrite” the existing parts.

The existing one is an extension

* Manopt & JuMP

But for specific manifold we need an extension to 

* Manopt, Manifold & JuMP

The Other PR defines the Manifolds-type-wrapper in the first extension and tries a proof of concept in the second.
I am not yet sure how this idea could be fit into this general setting, nor have I understood how it is meant to work.

blegat:
Yes I mean what would be a typical objective I can use in the tests.

> But the prior PR therefore “unwrapped” the variables before going to the state/solver.

The solver still receives them unwrapped, we'll have to pass the point type as well. The only difference here is that they are wrapped at the JuMP level so the user will be able to use the operators you have define for `HyperboloicPoint` to build the objective function

ext/ManoptJuMPExt.jl
kellertuer: ...your docs do not help https://jump.dev/JuMP.jl/stable/api/JuMP/#JuMP.add_variable – they state I should ask you, and neither are my code archeology skill that good nor have all my tries of that not lead to me understanding how this is supposed to work, nor what for example the string is.

Issue #490, opened 2025-07-18
Three problems with RCG


jonas-pueschel:
Hi all,
while playing around with RCG, I've encountered three pretty severe issues, wich in part are caused by `Manopt` in general but also specific to the `conjugate_gradient_descent` method. These issues are

1. using exponential retraction, RCG leaves the sphere because of accumulating roundoff errors in the formula,
2. RCG with Armijo backtracking stagnates, if the direction produced by RCG is not a descent direction,
3. the method `get_gradient` gets called twice per iteration in RCG, although it should only be called once. Possibly, the same occurs with `get_cost`.

## Code

Here is the complete code to reproduce the issues. Below the code I will go into more detail regarding the three points. The issues explained can be observed in the console output. Please excuse the verbose nature of the code.
```julia
using Manifolds
using Manopt
using LinearAlgebra
using Printf

n = 4
A = diagm(1:n)

M = Sphere(n-1)

mutable struct ObjectiveInfo 
    grad_calls
    cost_calls
    function ObjectiveInfo()
        return new(0,0)
    end
end

function clear_oi(oi::ObjectiveInfo)
    oi.grad_calls = 0
    oi.cost_calls = 0
end

function get_cost_oi(p, oi::ObjectiveInfo)
    oi.cost_calls += 1
    return dot(p, A*p) 
end

function get_gradient_oi(p, oi::ObjectiveInfo)
    oi.grad_calls += 1
    return A*p - dot(p, A * p) * p
end

oi = ObjectiveInfo()
obj =ManifoldFirstOrderObjective(;
    cost = (M,p) -> get_cost_oi(p, oi),
    gradient = (M,p) -> get_gradient_oi(p, oi)
)


function debug_oi(problem, state, n_iter; oi::ObjectiveInfo, display_errors = false, display_calls = false)
    if (n_iter == 0)
        err_head1 = display_errors ? "| log10(e_norm) | log10(e_ts) " : ""
        err_head2 = display_errors ? "+---------------+-------------" : ""
        calls_head1 = display_calls ? "| calls cost | calls grad " : ""
        calls_head2 = display_calls ? "+------------+------------" : ""
        println(" Iter   $err_head1| d | cost       | log10(|g|) | stepsize          $calls_head1")
        println("--------$err_head2+---+------------+------------+-------------------$calls_head2")
    end
    p = state.p
    Ap = A*p
    cost = dot(p, Ap) #This should not be necessary, why is cost not in state?
[...]

PR #492, opened 2025-07-22
Add restart condition to RCG [fixes problem 2. from #490]

jonas-pueschel:
Hi all,

I've implemented restart conditions for `conjugate_gradient_descent`. As a proof of concept, I implemented both checking for descent and checking for the sufficient descent condition discussed by Sakai, Iiduka. It is currently implemented for CG, but could easily be extended to other methods too.

Restart Conditions extend the abstract type `AbstractRestartConditon` and are functors called after calculating `cgs.δ`. 
If the condition for restarting is met, the direction gets set to the negative gradient and the storage of `cgs.coefficient` is updated.

The default value of the newly added function variable `restart_condition` is `RestartNever()`, not changing the behaviour of previously implemented calls to `conjugate_gradient_descent`.

I also have a script showing the proficiency of the changes, showing that convergence improves significantly.
```julia
using Manifolds
using Manopt
using LinearAlgebra
using Random

n = 4
A = diagm(1:n)
M = Sphere(n-1)

obj = ManifoldFirstOrderObjective(;
    cost = (M,p) -> dot(p, A*p),
    gradient = (M,p) -> A*p - dot(p, A * p) * p
)

function debug(problem, state, n_iter)
    if (state.stop(problem, state, n_iter))
        if state.stop.criteria[2].at_iteration == -1
            println("no convergence, ||g|| = $(norm(state.X)).")
        else
            println("convergence after $n_iter steps.")
        end
    end
end

retraction_method = ProjectionRetraction()
vector_transport_method = ProjectionTransport()
stopping_criterion= StopAfterIteration(500) | StopWhenGradientNormLess(1e-6)
function get_stepsize_strat() 
    Manopt.ArmijoLinesearchStepsize(
        M; retraction_method=retraction_method, initial_stepsize=1.0, stop_decreasing_at_step = 100
    )
end

# Sufficient Descent > Steepest Descent
Random.seed!(3) 

# Sufficient Descent < Steepest Descent
# Random.seed!(1) 

p0 = rand(M)

# 1. No Correction
print("1: ")
restart_condition = RestartNever()
conjugate_gradient_descent!(M, obj, copy(p0);
    debug, retraction_method, vector_transport_method, 
    restart_condition, stopping_criterion, 
    stepsize = get_stepsize_strat());

# 2. Descent Correction
print("2: ")
restart_condition = RestartDescent()
conjugate_gradient_descent!(M, obj, copy(p0);
    debug, retraction_method, vector_transport_method, 
    restart_condition, stopping_criterion, 
    stepsize = get_stepsize_strat());

# 3. Sufficient descent CG
print("3: ")
restart_condition = RestartSufficientDescent(0.5)
conjugate_gradient_descent!(M, obj, copy(p0);
    debug, retraction_method, vector_transport_method, 
    restart_condition, stopping_criterion, 
    stepsize = get_stepsize_strat());

# 4. Steepest Descent
print("4: ")
conjugate_gradient_descent!(M, obj, copy(p0);
    debug, coefficient = SteepestDescentCoefficient(),
    restart_condition, stopping_criterion, 
    stepsize = get_stepsize_strat());
```
**Edit**
Here is the outputs for `Random.seed!(3)`:
```
1: no convergence, ||g|| = 0.7046071312104397.
2: convergence after 283 steps.
3: convergence after 83 steps.
4: convergence after 119 steps.
```
and for `Random.seed!(1)`:
```
1: no convergence, ||g|| = 0.7330884915851944.
2: convergence after 328 steps.
3: convergence after 84 steps.
4: convergence after 78 steps.
```
**/Edit**
Note that steepest descent still sometimes outperforms RCG, highlighting the need for an improved step size control.
I am not sure if this is the correct/best way to share these kind of scripts, so please let me know what would be the correct way of doing it.

Best,
Jonas


kellertuer:
Very nice and thanks for your first contribution PR!

The restart already like a very well modelled idea; maybe a function is also ok instead of the functor, but that is just a matter of taste maybe,
one could (but does not have to) also do a `function_name_to_be_found(problem, state, iteration, restart_strategy)` instead.

You evenalready used the variable dictionary, great!

And sure, we could even use the same idea for quasi Newton, which currently has a separate type I think (the cautious quasi newton with a name due to Huang et al.)

Not changing the behaviour is a good idea. We could change the default on a next breaking change.

For sharing the script, could you maybe add your example output here as well? Otherwise that is a good way to share it; just of one sees the result here as well one does not have to checkout your branch.

Small remarks:
* we need a changelog entry still
* we have to add a few tests for the new functionality

kellertuer:
Feel free to keep that in the PR if you add it to the changelog. I do not mind a bit of chaotic PRs.

src/Manopt.jl
```diff
@@ -465,6 +465,11 @@ export SteepestDescentCoefficient,
     DaiYuanCoefficient,
     HagerZhangCoefficient,
     ConjugateGradientBealeRestart
+# Direction Corrections
+export AbstractRestartCondition
+export RestartNever,
+    RestartDescent,
+    RestartSufficientDescent
```
kellertuer, line 472: I think naming wise I would prefer the grammatically easier to parse `NeverRestart` or even `NeverRestartCondition` for the first?

src/plans/conjugate_gradient_plan.jl
```diff
@@ -87,13 +89,14 @@ mutable struct ConjugateGradientDescentState{
         sC::StoppingCriterion,
         s::Stepsize,
         dC::DirectionUpdateRule,
+        res_cond::AbstractRestartCondition=RestartNever(),
```
kellertuer, line 92: We could make this a bit nicer by writing `::RC` and adding `RC<:AbstractRestartcondition` in the `where` – then we can avoid the `typeof(...)` and just write `RC` there (also for the other other cases, just excluding `coef` and `βT`.

src/plans/conjugate_gradient_plan.jl
```diff
@@ -77,23 +78,25 @@ mutable struct ConjugateGradientDescentState{
     δ::T
     β::F
     coefficient::TCoeff
+    restart_condition::TRC
     stepsize::TStepsize
     stop::TStop
     retraction_method::TRetr
     vector_transport_method::VTM
-    function ConjugateGradientDescentState{P,T}(
+    function ConjugateGradientDescentState{P,T,TsC,TStep,TRC,TRetr,VTM}(
```
kellertuer, line 86:
```suggestion
    function ConjugateGradientDescentState(
```
I think it does not like (nor need) these types here

src/plans/conjugate_gradient_plan.jl
```diff
@@ -112,21 +116,23 @@ function ConjugateGradientDescentState(
...
-    stopping_criterion::StoppingCriterion=StopAfterIteration(500) |
+    stopping_criterion::TsC=StopAfterIteration(500) |
                                           StopWhenGradientNormLess(1e-8),
-    vector_transport_method=default_vector_transport_method(M, typeof(p)),
+    vector_transport_method::VTM=default_vector_transport_method(M, typeof(p)),
     initial_gradient::T=zero_vector(M, p),
-) where {P,T}
-    return ConjugateGradientDescentState{P,T}(
+) where {P,T,TsC<:StoppingCriterion,TStep<:Stepsize,TRC<:AbstractRestartCondition,TRetr<:AbstractRetractionMethod,VTM<:AbstractVectorTransportMethod}
+    return ConjugateGradientDescentState{P,T,TsC,TStep,TRC,TRetr,VTM}(
```
kellertuer, line 129:
```suggestion
    return ConjugateGradientDescentState(
```
missed to remove it here (and have to add those bz comment anyways).

kellertuer:
From the error message (some place it gets a stepsiye where it expects a restart),, somewhere the order of parameters is still wrong or the new restart is missing. I might check this out locally maybe to find it, just looking at the changes on this PR it is a bit hard to spot. 

A usual good trick is to put new parameters _last_, unless there is a good reason not to, but here I think it might help to put it last.

test/solvers/test_conjugate_gradient.jl
```diff
@@ -278,4 +278,53 @@ using LinearAlgebra: Diagonal, dot, eigvals, eigvecs
...
+            M,
+            obj,
+            p0;
+            restart_condition = RestartOnNonSufficientDescent(0.5),
+            stopping_criterion,
+            stepsize = get_stepsize()
+        )
+        # sufficient descent should perform best, descent better than no restart
+        @test e_func(p3) < e_func(p2) && e_func(p2) < e_func(p3)
+        @test e_func(p3) ≈ 0 atol = 1e-3
```
kellertuer, line 328: the first test will always evaluate to false. 
Why should the second be true? currently that is about 1 not 0.

kellertuer:
You just merged your own master (the one from your fork) instead of the actual one. Usually GitHub helps with a “sync” button by now – but now worries, I did the merge then now :)

src/plans/conjugate_gradient_plan.jl
```diff
@@ -64,36 +64,39 @@ $(_var(:Keyword, :vector_transport_method))
...
+        res_cond::TRC=NeverRestart(),
+        retr::TRetr=default_retraction_method(M, typeof(p)),
+        vtr::VTM=default_vector_transport_method(M),
         initial_gradient::T=zero_vector(M, p),
-    ) where {P,T}
+    ) where {P,T,TsC<:StoppingCriterion,TStep<:Stepsize,TRC<:AbstractRestartCondition,TRetr<:AbstractRetractionMethod,VTM<:AbstractVectorTransportMethod}
         coef = DirectionUpdateRuleStorage(M, dC; p_init=p, X_init=initial_gradient)
         βT = allocate_result_type(M, ConjugateGradientDescentState, (p, initial_gradient))
-        cgs = new{P,T,βT,typeof(coef),typeof(s),typeof(sC),typeof(retr),typeof(vtr)}()
+        cgs = new{P,T,βT,typeof(coef),TRC,TStep,TsC,TRetr,VTM}()
```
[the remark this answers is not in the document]
kellertuer: I added this as a commit but will keep this unresolved to discuss it. Either here or the next days in a zoom :)

src/plans/conjugate_gradient_plan.jl
```diff
@@ -64,36 +64,39 @@ $(_var(:Keyword, :vector_transport_method))
...
     T,
-    F,
-    TCoeff<:DirectionUpdateRuleStorage,
-    TStepsize<:Stepsize,
     TStop<:StoppingCriterion,
+    TStepsize<:Stepsize,
+    TRC<:AbstractRestartCondition,
     TRetr<:AbstractRetractionMethod,
     VTM<:AbstractVectorTransportMethod,
+    F,
```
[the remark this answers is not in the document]
jonas-pueschel: I remember this being important, basically the constructor needs the template order in the very same order as the templated types of the struct itself, if any are given. Since the constructor does not template `F` or `TCoeff`, these need to go behind all the ones that are templated, Re-ordering leads to very unhelpful error messages, that were a real pain to debug. Note that this is only the case if the constructor is defined inside of the struct, when we put it out, this will no longer be necessary.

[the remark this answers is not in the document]
kellertuer: Yes sure, but the order was totally off. Currently F is in last place, but I have no clue why it was moved at all. The order in the last commit is the correct one corresponding to the ordering here (since I can only do small changes on this PR by these edits here).

kellertuer:
Dankeschön!

PR #493, opened 2025-07-24
Switch to using Runic.jl as code formatter.

kellertuer:
Inspired by the short JuliaCon talk, this tests out how Runic would format Manopt and whether the CI would work.


mateuszbaran:
I think the Runic.jl-formatted code looks fine, and Runic might be a more future-proof solution than staying with JuliaFormatter v1 or hoping JuliaFormatter v2 is correct now.

mateuszbaran:
I actually always run JuliaFormatter from REPL because I wasn't sure the VSCode integration does the right thing, so that's not a problem for me. As long as formatting is one REPL command, I'm fine (I just grab it from REPL history).

PR #494, opened 2025-07-25
Avoid gradient computations [#480, P3]

kellertuer:
This is an idea to reduce gradient computations.

It allows to pass the evaluated gradient to any stepwise `gradient=`

The example 3 from #480 now produces (where I am not sure what it produced before)

```
julia> conjugate_gradient_descent!(M, obj, p;
           debug = (args...) -> debug_oi(args...; oi, display_calls = true), 
           coefficient = SteepestDescentCoefficient(),
           stopping_criterion = Manopt.StopAfterIteration(200))
 Iter   | d | cost       | log10(|g|) | stepsize          | calls cost | calls grad 
--------+---+------------+------------+-------------------+------------+------------
      0 | ✓ | +1.1908973 | -0.2969576 | +1.00000000000000 |          0 |          1 
      1 | ✓ | +1.1546733 | -0.2591312 | +1.36037414199981 |         10 |          2 
      2 | ✓ | +1.1225207 | -0.2778392 | +1.00000000000000 |         18 |          3 
      3 | ✓ | +1.0909885 | -0.3035586 | +0.81450625000000 |         24 |          4 
      4 | ✓ | +1.0714433 | -0.3415070 | +0.69833729609375 |         29 |          5 
    [...]
    200 | ✓ | +1.0000000 | -7.4239552 | +0.66342043128906 |        692 |        201 
```
where I am not sure what the output before was. Sure for every step size checked there is a cost evaluation, one could maybe also check to reduce that, but it would maybe only be a reduction by 1 – but we would have to start to store a cost in states to do the same trick.

I am not yet fully happy with this

* [ ] is `X` as memory really needed? If no gradient is given we could also just allocate – the case that someone has memory but not the gradient is maybe a bit niche. Maybe removing it makes the code easier again.
* [ ] Check where else we can pass down the gradient to step sizes besides in CG

### DIFF
--- a/src/plans/stepsize.jl
+++ b/src/plans/stepsize.jl
@@ -9,6 +9,10 @@ the interface `(p,o,i)` where a [`AbstractManoptProblem`](@ref) as well as [`Abs
 and the current number of iterations are the arguments
 and returns a number, namely the stepsize to use.
 
+The functor usually should accept arbitrary keyword arguments. Common ones used are
+* `X`: memory for a tangent vector, initialised to a zero vector at the current iterate
+* `gradient`: the actual gradient at the current iterate, evaluated in-place of `X`.
+
 For most it is advisable to employ a [`ManifoldDefaultsFactory`](@ref). Then
 the function creating the factory should either be called `TypeOf` or if that is confusing or too generic, `TypeOfLength`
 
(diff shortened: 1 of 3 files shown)


mateuszbaran (approved):

kellertuer:
@jonas-pueschel please take a look, this is ready to merge and should resolve your point 3 :)

PR #499, opened 2025-08-08
Maybe at least improve JuMP docs

kellertuer:
This is an approach to keep something minimal from #454 – external links in the docs and not hacking around with global sets on load.

But after an hour of twiddling, the docs still do not render and either do not find certain functions or not knowing them when I type them out – so maybe also this might stall if I do not find a way to fix the docs.

The main feature besides a bit more precise naming internally is that all functions to `JuMP` now work with `@extref` and CSS adds the logo upfront to indicate in the docs that these are external links like in

Screenshot 2025-08-08 at 15 56 43


Changelog.md
```diff
@@ -21,6 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   * `quasi_Newton`
   * `projected_gradient_method`
 
+### Changed
+
+* remodeeled the docs for the extensions a bit, added `JuMP` to the DocumenterInterlinks.
```
blegat: remodeeled -> remodeled

kellertuer: Thanks for spotting.

kellertuer:
Nice.
Yeah the one thing that does change a bit are the names of the inner structs – also something I did back on the other PR to (a) maybe reflect a bit better what they do (and sure, naming is hard) and (b) allow to maybe some day still do the non-array ones.

I love DocumenterInterlinks, since Manopt very often refers to both ManifoldsBase and Manifolds – and I also tweaked the CSS to do the prefix icons, which is especially handy as soon as you have structs that mix quite a few things like our boilerplate Optimiser struct:

Screenshot 2025-08-09 at 08 45 34

I'll merge this later today once the tests have all passed (again).

kellertuer:
Ah you do not have to rebase, we usually just merge – and then squash-merge to master (I should one day rename that to main) – to keep the main branch more organised.

blegat:
I really like the prefix icons, it makes it very clear where this come from. They would be conflict in any case so I prefer rebasing, it keeps the branch cleaner I feel ^^

PR #503, opened 2025-08-21
An idea to provide Hints and Errors on Keywords

kellertuer:
Since a few things are nested and some keywords are “passed on”, e.g. to the debug or record factory,
currently there is no errors or warnings when a keyword is provided, that says “unused”.

There is also no easy way to tell which keywords could be passed e.g. to a certain solver,

This is a small experiment based on the approach from [Makie.jl](https://github.com/MakieOrg/Makie.jl/blob/d8f53444bc7d328d351d3265b1380d279eb2058b/Makie/src/recipes.jl#L923-L944) to see whether we can warn, error and/or give hints on keywords. Similarly, this should be usable to upfront determine all keywords that are accepted by a function … in the long run maybe even with a remark what/where it would be used internally for

# 🗺️ 
* [x] test with one state and solver how to best create their list of keyword list
* [x] find a way to verify existing keywords
* [x] implement a difference to find unused ones
* [x] warn or even error (a strict mode?) on unused keywords
* [x] export the function that returns all keywords
* ~~implement a Levenstein distance to find closest keywords if someone has a typo?~~


kellertuer:
A first small thing that works by now – even using two macros to define keywords for a single function and one collecting all it passes on is that a first function can now report its keywords and where they are passed on to in the end.

```
Julia> Manopt.accepted_keywords(gradient_descent)
Keywords for gradient_descent:

accepted: 
  * cache (passed on to decorate_objective!)
  * callback (passed on to decorate_state!)
  * count (passed on to decorate_objective!)
  * debug
  * differential
  * direction
  * evaluation
  * objective_type (passed on to decorate_objective!)
  * p (passed on to decorate_objective!)
  * record (passed on to decorate_state!)
  * retraction_method
  * return_objective (passed on to decorate_objective!)
  * return_state (passed on to decorate_state!)
  * stepsize
  * stopping_criterion
  * X
  ```

**Comment by kellertuer** [comment]
This now indeed works. I still have to go though solvers and maybe some constructors to see where they pass `kwargs...` to and set the `calls_with_kwargs(f) = Tuple...` for those functions correctly.

in `mode=:error` one can already get (for solvers until CMAES&grad_desc):

```
julia> gradient_descent(M, f, grad_f; fnump=1)
ERROR: gradient_descent! does not accept the keyword(s)

  * fnump

Hint: gradient_descent! does accept the following keywords:

  X, cache, callback, count, debug, differential, direction, evaluation, objective_type, record, retraction_method, return_objective, return_state, stepsize, stopping_criterion
```

And the mode (whether to warn or to error) I maybe want to set as a preference like the tutorial mode.

So this small experiment is on a good way.

**Comment by kellertuer** [comment]
good point, maybe something like `accepted_in`?

**Comment by kellertuer** [comment]
it actually states the “current function” the keywords are meant for, so something along/similar to function (again not directly that ;)) would be fitting as well.

File: src/plans/keywords.jl [hunk 1]
```diff
@@ -0,0 +1,252 @@
...
+struct Keywords{I}
+    accepted::Set{Symbol}
+    deprecated::Set{Symbol}
+    origins::Dict{Symbol, Union{T, Vector{T}} where {T <: Union{Function, Type}}}
+end
+function Keywords(
+        accepted::Set{Symbol} = Set{Symbol}(),
+        deprecated::Set{Symbol} = Set{Symbol}();
+        from = nothing,
+        origins = nothing,
```
mateuszbaran, line 34:
```suggestion
        origins::Union{Dict,Nothing} = nothing,
```

src/plans/keywords.jl
```diff
@@ -0,0 +1,252 @@
...
+    )
+    if !isnothing(from)
+        _origins = isnothing(origins) ? Dict{Symbol, Any}() : origins
```
mateuszbaran, line 37: `_origins` has three different types here: `Dict{Symbol, Union{T, Vector{T}} where {T <: Union{Function, Type}}}`, `Dict{Symbol, Any}` or `_origins = Dict{Symbol, Union{<:Function, <:Type}}`. It'd be preferable to stick to one. I personally prefer `Any` over `Union{Function, Type}` because `Any` would also work for functors.

BTW, is a `Vector` value in `origins` a remnant from older drafts or something you'd like to support? Why not having it always be a vector for consistency?

src/plans/plan.jl
```diff
@@ -65,8 +75,8 @@ function get_parameter(
     return @load_preference("$(e)", default)
 end # Handle empty defaults
 get_parameter(::Symbol, ::Val{:default}) = nothing
-get_parameter(::Val{:Mode}, v::Val{:default}) = nothing
-
+get_parameter(::Val{:Mode}, ::Val{:default}) = nothing
+get_parameter(::Val{:KeywordsErrorMode}, ::Val{:default}) = "warn"
```
mateuszbaran, line 79: Shouldn't it be `:warn` for consistency with `keywords_accepted`?

src/plans/keywords.jl
kellertuer: Ok, sure, this is also not performance critical since it is meant to only be called outside of those.

We could at least do `Dict{Symbol, Vector{V}}` and make the `V` a parameter in the corresponding struct?

src/plans/plan.jl
mateuszbaran: I see, OK, then it is what it is.

src/plans/keywords.jl
kellertuer: Hm that is not so easy, every now and then I need it to be nothing as well

mateuszbaran: Making this faster than what what it would be with `Dict{Symbol, Vector{Any}}` would be challenging. Having such `V` would be a part of it, sure, but you'd also have to set `V` to small unions of concrete types and inspect that union splitting actually can happen, which is IMO not worth the small benefits it could provide.

mateuszbaran:
Everything else looks fine to me, so it's just the `origins` thing that I'd suggest improving.

mateuszbaran (approved):

PR #505, opened 2025-09-05
Add Distance over Gradients Stepsize

Nimrais:
This PR adding a new stepsize schedule — Distance over Gradients (RDoG). RDoG is a learning-rate-free stepsize that adapts automatically without hyperparameter tuning https://arxiv.org/pdf/2406.02296.

You can use it like any other stepsize schedule.

```julia
using Manopt
using Manifolds
using LinearAlgebra
using Random

Random.seed!(42)

# Minimize negative Rayleigh quotient on the sphere S^1
M = Sphere(1)
A = randn(2, 2); A = A' * A  # symmetric positive definite

f(M, p) = -p' * A * p
function grad_f(M, p)
    g = -2 * A * p
    return g - dot(g, p) * p  # project to tangent space
end

p0 = rand(M)

x = gradient_descent(
    M, f, grad_f, p0;
    stepsize = DistanceOverGradients(M; initial_distance = 1e-2, use_curvature = false),
    stopping_criterion = StopAfterIteration(200) | StopWhenGradientNormLess(1e-8),
)

println("final cost = ", f(M, x))
```

### DIFF
--- a/src/plans/stepsize.jl
+++ b/src/plans/stepsize.jl
@@ -1767,12 +1767,7 @@ function geometric_curvature_function(κ::Real, d::Real)
     if κ < 0 && d > 0
         sqrt_abs_κ = sqrt(abs(κ))
         arg = sqrt_abs_κ * d
-        # Avoid numerical issues for small arguments
-        if arg < 1.0e-8
-            return 1.0 + arg^2 / 3.0  # Taylor expansion
-        else
-            return arg / tanh(arg)
-        end
+        return arg / tanh(arg)
     else
         return 1.0
     end


kellertuer:
Nice, thanks!

At first glance it seem the current constructor is “the old one” with `M` necessary – maybe we can have a second ode employing the factory (I wrote that after our discussions at last years JuliaCon)?

I will try to find time to review your code, but it would also be nice to write a test case and have test coverage. Maybe just one further run in the gradient descent tests where you have a cost and gradient already...

docs/src/references.bib
```diff
@@ -5,6 +5,13 @@
 % --- A
 %
 %
+@article{dodd2024learning,
```
kellertuer, line 8:
```suggestion
@article{DoddSharrockNemeth:2024,
```

This would be the unified naming scheme we use. Could we use that here as well?
Also the literature is ordered by first-authors-lastname.

src/Manopt.jl
```diff
@@ -522,7 +522,12 @@ export SmoothingTechnique, LinearQuadraticHuber, LogarithmicSumOfExponentials
 #
 # Stepsize
 export Stepsize
-export AdaptiveWNGradient, ConstantLength, DecreasingLength, Polyak
+export AdaptiveWNGradient,
```
kellertuer, line 525: Since we (ok only since yesterday or so) use Runic, we can also do this in two lines.

src/plans/stepsize.jl
```diff
@@ -1687,3 +1687,193 @@ end
...
+# Fields
+
+* `initial_distance::R`: initial distance estimate ``ϵ > 0``
+* `max_distance::R`: maximum distance from initial point ``\bar{r}_t = \max_{s≤t} d(p_0, p_s)``
+* `gradient_sum::R`: accumulated squared gradient norms ``G_t = \sum_{s=0}^t \|∇f(p_s)\|^2``
+* `initial_point`: the initial point ``p_0``
+* `use_curvature::Bool`: whether to use sectional curvature in the stepsize
+* `sectional_curvature_bound::R`: lower bound on sectional curvature ``κ`` (only used if `use_curvature=true`)
+
+The stepsize at iteration ``t`` is computed as:
```
kellertuer, line 1709: I think we usually documented the formulae with the variant without `Stepsize` (so the one with the factory note). That one is considered the public one. This one should document its keywords and refer to the other one.

src/plans/stepsize.jl
```diff
@@ -1687,3 +1687,193 @@ end
...
+    Current state:
+      max_distance = $(rdog.max_distance)
+      gradient_sum = $(rdog.gradient_sum)
+      last_stepsize = $(rdog.last_stepsize)
+    """
+    return print(io, s)
+end
+
+"""
+    DistanceOverGradients(; kwargs...)
```
kellertuer, line 1864: The `docs/src/stepsize.md` has a block (the third code block) documenting all available step sizes – can you please add this there as well? 

Currently it is automatically added a bit later in the file - but it should be listed in the available step sizes.

src/plans/stepsize.jl
```diff
@@ -1687,3 +1687,193 @@ end
...
+        NaN,              # last_stepsize
+    )
+end
+
+"""
+    geometric_curvature_function(κ::Real, d::Real)
+
+Compute the geometric curvature function ``ζ_κ(d)`` for sectional curvature bound ``κ``.
+
+```math
```
kellertuer, line 1769: One could either do a `raw"""` string to avoid the double `\\`,
or use interpolation and the tricks I started in the `documentation_glossary`.

The first is easier, the second complicated but a bit more flexible in the end.

If you feel this is too much to do, I can also refactor this later.

src/plans/stepsize.jl
```diff
@@ -1687,3 +1687,193 @@ end
...
+The stepsize at iteration ``t`` is computed as:
+
+```math
+η_t = \frac{\bar{r}_t}{\sqrt{ζ_κ(\bar{r}_t)} \sqrt{G_t}}
+```
+
+where ``ζ_κ`` is the geometric curvature function:
+- ``ζ_κ(d) = 1`` if ``κ ≥ 0`` or `use_curvature=false`
+- ``ζ_κ(d) = \frac{\sqrt{|κ|} d}{\tanh(\sqrt{|κ|} d)}`` if ``κ < 0``
+
```
kellertuer, line 1718: Since you have a separate function, you could also just link to that (`@ref`) and avoid documenting the formula twice.

kellertuer (commented):
Thanks for the PR!

I am (maybe I said tha already) not so much a fan of that paper. Illustrating that a gradient descent with a constant stepsize does not converge, is no wonder – one needs an Armijo linesearch for example.
Sure they could have compared to Armijo and argued why theirs is easier, ... that would have been the proper comparison. But well, that is not your fault of couse

Here are a few small comment, mainly on the documentation, which I rendered locally.
Test coverage would be nice.

docs/src/references.bib
Nimrais: ah sure :). I just assumed you already have done it so I just hit the button. No problem.

kellertuer: yeah, sorry for not being precise.
If I could've, I would have – but I can not push to this branch, so all I can change is files that were already changed – and for those even a bit limited with the frontend here.

Nimrais: ok now it is correct.

kellertuer:
Great!

And sure, shorter PRs that focus on one thing are better than super long PRs, so that decision is a very good one I think :)

Let me just check the rendered docs during the day somewhen, the rest looks already good I think.

src/plans/stepsize.jl
```diff
@@ -1687,3 +1687,188 @@ end
...
+    """
+    return print(io, s)
+end
+
+"""
+    DistanceOverGradients(; kwargs...)
+    DistanceOverGradients(M::AbstractManifold; kwargs...)
+
+Creates a factory for the [`DistanceOverGradientsStepsize`](@ref).
+
```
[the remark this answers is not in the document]
Nimrais: I moved it from Stepsize docs, is it what you asked?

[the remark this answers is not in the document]
kellertuer: The step size ones are good I think. Technical and detailed, good for an internal doc (as the Stepsize types are)

Here a general short description and a cite to the paper might be nice.

src/plans/stepsize.jl
```diff
@@ -1687,3 +1687,223 @@ end
...
+For small arguments, a Taylor approximation is used for numerical stability.
+"""
+function geometric_curvature_function(κ::Real, d::Real)
+    if κ < 0 && d > 0
+        sqrt_abs_κ = sqrt(abs(κ))
+        arg = sqrt_abs_κ * d
+        # Avoid numerical issues for small arguments
+        if arg < 1.0e-8
+            return 1.0 + arg^2 / 3.0  # Taylor expansion
```
kellertuer, line 1772: This one is currently not hit by tests (missed it on first check).

Do you have a good check/test for this? Otherwise we can also leave this one line untested, if a test would be too tricky.

src/plans/stepsize.jl
```diff
@@ -1687,3 +1687,223 @@ end
...
+# Constructor
+
+    DistanceOverGradientsStepsize(M::AbstractManifold; kwargs...)
+
+## Keyword arguments
+
+* `initial_distance=1e-3`: initial estimate ``ϵ``
+* `use_curvature=false`: whether to use ``ζ_κ``
+* `sectional_curvature_bound=0.0`: lower curvature bound ``κ`` (if known)
+$(_var(:Keyword, :p; add = "initial point, used to track distances"))
```
kellertuer, line 1713:
```suggestion
* `p`: initial point, used to track distance
```

src/plans/stepsize.jl
Nimrais: Honestly, I think we should remove it. This Taylor expansion is not better — I checked it and it produces exactly the same results.

PR #506, opened 2025-09-08
add methods to the functions `<` and `<=` instead of to `>`, `>=`

nsajko:
As documented, the intended way to implement `>` is to add a method to `<`. Similarly with `>=`. A package should never add a method to either `>` or `>=`.

### DIFF
--- a/src/plans/stopping_criterion.jl
+++ b/src/plans/stopping_criterion.jl
@@ -1391,10 +1391,10 @@ function StopWhenCriterionWithIterationCondition(
     ) where {SC <: StoppingCriterion, F}
     return StopWhenCriterionWithIterationCondition{SC, F}(sc, comp, -1)
 end
-function Base.:>(sc::StoppingCriterion, n::Int)
+function Base.:<(n::Int, sc::StoppingCriterion)
     return StopWhenCriterionWithIterationCondition(sc, n; comp = (>(n)))
 end
-function Base.:>=(sc::StoppingCriterion, n::Int)
+function Base.:<=(n::Int, sc::StoppingCriterion)
     return StopWhenCriterionWithIterationCondition(sc, n; comp = (>=(n)))
 end
 function Base.:(==)(sc::StoppingCriterion, n::Int)


kellertuer:
Thanks for the note, this is mainly meant as syntactic sugar, to say `stopping_criterion > n` that it should hit (indicate to stop) more than `n` times. I am not sure it would make so much sense to write `n < sc`? 

But you are right this is a bit of abuse of notation - if you have a better Idea I am open to that.

kellertuer:
No I fully understand the PR.

It introduces `n < stopping_criterion` as well and the old thing Is possible still since it falls back to that version.
But I am not sure that notation makes much sense or is useful. So the original format was more like on purpose. Or in short: Will this new (yes additional) thing confuse users?
(I simply do not know)

Or since it is not meant to be symmetric, was my choice of notation maybe in the end not that good?
(I also do not know, but I am very open to alternatives)

nsajko:
Oh I get you now, sorry. IMO, given the lack of symmetry, perhaps you want to introduce a new function instead of `>`.

In any case, be aware that adding methods to `>` or `>=` tends to trigger invalidation, see, for example:

* https://github.com/JuliaDiff/ForwardDiff.jl/pull/771#issuecomment-3267180470

* https://github.com/jump-dev/JuMP.jl/pull/4065#issuecomment-3264639753

* GenericMappingTools/GMT.jl#1802

kellertuer:
Sorry missed the comment in-between.

So the idea of `>` was that one can easily combine them. There is a separate function, namely `StopWhenCriterionWithIterationCondition` (sorry there is also a typo in the docs, namely a wrong constructor in https://manoptjl.org/stable/plans/stopping_criteria/#Manopt.StopWhenCriterionWithIterationCondition)

But similar to `StopWhenAll` which has the short `&` operator, `StopWhenAny` which has `|` and `StopWhenRepeated` which has `×` (not abelian, order is important=, so that one can easily write to combine two existing stopping criteria `sc1, sc2` for these
`sc1 & sc2`, `sc1 | sc2`, and `sc1 × 3`.
 I wanted to have a short form for the generic case here as well.

So instead of writing `StopWhenCriterionWithIterationCondition(sc1, , n; comp = (>(n)))` this allows to write `sc1 > n`.

But if no one likes that, we can also delete the shortcut and wait until someone is annoyed and/or comes up with a better idea. It recently seems anyways that most of my ideas are bogus, so maybe we should just remove it.
I do not understand the linked PRs and the reason for opening all those PRs, but that is ok, that sounds so super technical, that it is beyond me anyways.

So yeah. Maybe lets just delete all the nice things I thought would be neat to have.

kellertuer:
The \times one I still feel should be ok since the cross product is not abelian. That was also the main reason I started thinking about the others. But maybe lets really just remove those then.

PR #512, opened 2025-09-11
Add hybrid CG parameters

jonas-pueschel:
Hi all,

I've added hybrid CG parameters, as described in the papers by Sakai and Iiduka. Some examples:

The FR-PRP parameter (my favourite) is given by
```math
 \beta_k^{FR-PRP} = \max\{0, \min\{\beta_k^{FR}, \beta_k^{PRP}\}\}
 ``` 
and can now be used via `HybridCoefficient(FletcherReevesCoefficient(), PolakRibiereCoefficient())`

The "original" HS-DY parameter with $0<\sigma < 1$ is given by
```math
 \beta_k^{HS-DY} = \max\{-\sigma \beta_k^{DY} , \min\{\beta_k^{HS}, \beta_k^{DY}\}\}
 ``` 
and can now be used via 
```julia
HybridCoefficient(HestenesStiefelCoefficient(), DaiYuanCoefficient();
    lower_bound = DaiYuanCoefficient(),
    lower_bound_scale = -σ
)
```
Note that the constructor allows an arbitrary amount of coefficients as arguments.

### DIFF
--- a/src/plans/conjugate_gradient_plan.jl
+++ b/src/plans/conjugate_gradient_plan.jl
@@ -1061,7 +1061,7 @@ Construct the hybrid coefficient update rule.
 
 [`HybridCoefficient`](@ref), [`conjugate_gradient_descent`](@ref)
 """
-struct HybridCoefficientRule{F<:Real} <: DirectionUpdateRule
+struct HybridCoefficientRule{F <: Real} <: DirectionUpdateRule
     coefficients::Vector{<:DirectionUpdateRule}
     lower_bound::DirectionUpdateRule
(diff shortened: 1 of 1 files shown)


kellertuer (approved):
I fixed all issues here, so this is good to go.

One thing one could look at in a future PR is, that currently all CG rules with the hybrid one build their own Storage (e.g. for the last iterate or last gradient). One could also build one joint storage for them all. The challenge there is to find out which things the storage has to provide to make all inner rules “happy”.

Thanks for this nice PR.

PR #518, opened 2025-10-03
RPG fallback

hajg-ijk:
Add a small fallback to the backtracking in the `proximal_gradient_plan` for cases where the `LogarithmicInverseRetraction()` is not available.


src/plans/proximal_gradient_plan.jl
```diff
@@ -508,8 +508,11 @@ function (s::ProximalGradientMethodBacktrackingStepsize)(
 
         # Compute log_p(candidate_point) and its squared norm for the conditions
         log_p_q = inverse_retract(M, p, candidate_point, st.inverse_retraction_method)
-        squared_distance = distance(M, p, candidate_point)^2
-
+        squared_distance = if st.inverse_retraction_method == LogarithmicInverseRetraction()
```
mateuszbaran, line 511: Wouldn't it make more sense to always calculate it as `norm(M, p, log_p_q)^2`? `norm` is generally faster and easier to compute than `distance`, and the result should be the same for `LogarithmicInverseRetraction`.

Changelog.md
```diff
@@ -50,7 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * a `gradient=` keyword in several `Stepsize`s, such that one can avoid to internally avoid computing the gradient again.
 * used the ``gradient=` keyword in
   * `alternating_gradient_descent`
-  * `conjugate_gradient`
+  * conjugate_gradient
```
kellertuer, line 54:
```suggestion
  * `conjugate_gradient`
```

why were those removed?

PR #529, opened 2025-11-02
Introduce a VectorBundle Newton method

kellertuer:
# 📋

* [x] that the tests are running
* [x] that the documentation looks nice
* [x] that the `docs/src/about.md` page is properly updated
* [x] that the `News.md` is properly updated
* [x] check that we have enough test coverage.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.5.25"
+version = "0.5.26"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"


src/solvers/vectorbundle_newton.jl
```diff
@@ -0,0 +1,340 @@
...
+            θ_new = norm(amp.manifold, ams.p, simplified_newton, acs.outer_norm) / norm(amp.manifold, ams.p, ams.X, acs.outer_norm)
+        else
+            θ_new = norm(simplified_newton) / norm(ams.X)
+        end
+
+        α_new = min(1.0, ((α_new * acs.θ_des) / θ_new))
+        if α_new < 1.0e-15
+            println("Newton's method failed")
+            return
+        end
```
kellertuer: This is currently never reached. What happens if it does reach this? Currently it does nothing and would continue in the next iterate - to end up here again? Is that if necessary?

src/solvers/vectorbundle_newton.jl
```diff
@@ -0,0 +1,340 @@
...
+
+        rhs_next = amp.newton_equation(M, get_vectorbundle(amp), ams.p, ams.p_trial)
+        rhs_simplified = rhs_next - (1.0 - α_new) * b
+        amp.newton_equation.b .= rhs_simplified
+
+        simplified_newton = ams.sub_problem(amp, ams)
+
+        if has_components(M) && !ismissing(acs.outer_norm)
+            θ_new = norm(amp.manifold, ams.p, simplified_newton, acs.outer_norm) / norm(amp.manifold, ams.p, ams.X, acs.outer_norm)
+        else
```
kellertuer: This is currently not tested. Can we test that maybe?

PR #531, opened 2025-11-07
Introduce two maxiter safeguards to WolfePowell and rewrite the whole messaging system

### LINKED ISSUE #495: Issue WolfePowellLinesearchStepsize
Hi all,
the `WolfePowellLinesearchStepsize` does not have a `maxiter` for inner iterations. Passing a "bad" search direction, e.g. a non-descent direction makes the inner loop go on indefinitely, freezing the method. Here is a small example how to reproduce it:
```julia
using Manopt
using Manifolds
using LinearAlgebra
using Random

n = 4
A = diagm(1:n)
M = Sphere(n-1)

obj = ManifoldFirstOrderObjective(;
    cost = (M,p) -> dot(p, A*p),
    gradient = (M,p) -> A*p - dot(p, A * p) * p
)
retraction_method = ProjectionRetraction()
vector_transport_method = ProjectionTransport()

function get_stepsize_strat() 
    Manopt.WolfePowellLinesearchStepsize(
        M; sufficient_curvature = 0.1,
        sufficient_decrease = 0.2
    )
end

Random.seed!(1)
p0 = rand(M)

conjugate_gradient_descent!(M, obj, copy(p0);
    debug = (mp, cgs, n_iter) -> println("iter $n_iter"),
    stepsize = get_stepsize_strat());
```
As already explained in #490, RCG sometimes does not correct non-descent directions. Here, after iteration 118, this leads to the stepsize strategy going into an infinite loop.
```
...
iter 112
iter 113
iter 114
iter 115
iter 116
iter 117
iter 118

```

kellertuer:
This resolves #495.
@jonas-pueschel is this what you had in mind? Note that I even adapted the message idea from Armijo, so one can even check afterwards whether these have happened and display a message.

# 📋
* [x] check how to do code coverage - maybe by setting them to zero so they fire directly.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,7 +6,7 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.5.27] November 11, 2025
 
 ### Added
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Manopt"
 uuid = "0fc0a36d-df90-57f3-8f93-d78a9fc72bb5"
 authors = ["Ronny Bergmann <manopt@ronnybergmann.net>"]
-version = "0.5.26"
+version = "0.5.27"
 
 [deps]
 ColorSchemes = "35d6a980-a343-548e-a6ea-1d62b119f2f4"


src/plans/stepsize.jl
```diff
@@ -1849,21 +1849,25 @@ $(_var(:Keyword, :X; add = "as type of memory allocated for the candidates direc
...
-        R <: Real, TRM <: AbstractRetractionMethod, VTM <: AbstractVectorTransportMethod, P, T,
+        R, TRM <: AbstractRetractionMethod, VTM <: AbstractVectorTransportMethod, P, T, I,
     } <: Linesearch
     sufficient_decrease::R
     sufficient_curvature::R
     candidate_direction::T
     candidate_point::P
     last_stepsize::R
     max_stepsize::R
+    message::String
```
mateuszbaran: Please don't store statuses as strings, that's a significant cost for small problems.

kellertuer: That is a good point. Quite a while back we reworked `get_reason` to not allocate a string. Sure that is something nice.

For now this follows the same route as `ArmijoLinesearchStepsize` – and actually all that call `line search_backtrack` (at least BarzileiBorwein and AWN), because especially with the modularisation of the lines each, I did not yet come up with any solution to avoiding this.

So for now this (just) is consistent with the others, for the very simple reason that I never had an idea how to improve this. When writing this I even did not have this in mind (and just mimicked Armijo), but I have spent quite some time a while back and never had an idea how to fix this.

So I agree this is not optimal, but for this PR the choice might be; Having something that is not optimal or not having the safeguards.

src/plans/stepsize.jl
```diff
@@ -1849,21 +1849,25 @@ $(_var(:Keyword, :X; add = "as type of memory allocated for the candidates direc
 * `max_stepsize=`[`max_stepsize`](@ref)`(M, p)`: largest stepsize allowed here.
 $(_var(:Keyword, :retraction_method))
 * `stop_when_stepsize_less=0.0`: smallest stepsize when to stop (the last one before is taken)
+* `stop_increasing_at_step=100`: for the initial increase test (s_plus), stop after these many steps
+* `stop_decreasing_at_step=1000`: for the initial decrease test (s_minus), stop after these many steps
 $(_var(:Keyword, :vector_transport_method))
 """
 mutable struct WolfePowellLinesearchStepsize{
-        R <: Real, TRM <: AbstractRetractionMethod, VTM <: AbstractVectorTransportMethod, P, T,
+        R, TRM <: AbstractRetractionMethod, VTM <: AbstractVectorTransportMethod, P, T, I,
```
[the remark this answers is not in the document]
mateuszbaran: I've never encountered anything real-like that isn't `<:Real`. Sometimes even non-real-like things subtype `Real` for convenience. I like having some upper bounds to make it easier to see what is supposed to be what.

kellertuer:
So, here is my approach to a rework. It took me 3hours to come up with this, and there is still quite a few tests failing now. But before I fix them: Does this allocate less enough?

I think I am the worst person to ask about allocations, I am not good at this. But I also do not like spamming a strict with a million int fields. So before I try to fix the tests, some feedback on whether this is reasonable or I should abandon this PR would be nice. This is a total detour from what the actual PR is about, but well. I am also used to that by now. My code is so bad, that no PR ends up doing what it intends to do but also rewrites a lot of other parts. It is what it is.

src/plans/stepsize.jl
kellertuer: a single reason might to be enough. I did now a dictionary see below, but it seems to become very tricky in like: At the point it happens I have all information to generate a good warning/error string.
Half a century later (when someone asks for a reason/error/message) I do not. So I could allocate all that information and store it somewhere and blow up the code tremendously – so that both allocations and code complexity goes up.

Or I allocate once there.

I am not sure. I am probably (yet again as for so often with programming) not clever enough here.

kellertuer:
Well before my 3 hour research we were approaching and infinite number of Int fields in all step sizes.

If the current one is not good, then I give up. That was the best I could do for something that is a total detour of this PR already.

Then I either leave this open until this is fixed in some way by someone else of slowly recommend to close. I am very tired of trying to fix code with a few lines that turns into a month of reworking a large code base – which is what I currently exclusively do.

src/plans/stepsize.jl
mateuszbaran: If a line search works well, no one will ever look at those messages so generating them is a wasted effort. I can help with coding this right. I wasn't aware that a few more fields are an issue for you, so I will try to come up with something cleaner.

kellertuer: Currently (without 3 additional ints) a simple thing like Armijo, which I would consider having 3 Euclidean parameters + retraction, already has 14 fields. And the old string warning was in one single nice place.

Now the string is at least also in one place, though one, where it has far less knowledge to report on. But sure it is a place where the string is only generated if actually asked.

For me a structure with more than 10 fields is slowly getting chaotic, at least if that is 3-4x the actual parameters of a thing.

PR #532, opened 2025-11-11
Proposal to remove the JuMP Extension

kellertuer:
_Edit:_ during this discussion the title and intention of the PR changed.

I gave the idea of non-array types another try, this is a very first sketch, me diving into an archaeological excatvation in paths through undocumented JuMP internals.

@blegat would you maybe help me in two things?

1. How would a non-array variable be added to JuMP? I did not find any example nor was I able to specify that in `@variable` myself. An example call would be neat to further play Tomb Raider through JuMP code with functions I would need
2. I renamed the second variable in 
https://github.com/JuliaManifolds/Manopt.jl/blob/2fdb6440df8285ea9c21254640468bc4b98fe28e/ext/ManoptJuMPExt.jl#L613-L619
from `func` to `point`, but otherwise it is often documented as `info`. What is this, what does it contain? Maybe magically some variable info where I can get the point type again from?
I currently have an `@info` in the code to get more of this, but it looks very very very very long and to me pure gibberish:

```
[ Info: ScalarVariable{Float64, Float64, Float64, Float64}[ScalarVariable{Float64, Float64, Float64, Float64}(VariableInfo{Float64, Float64, Float64, Float64}(false, NaN, false, NaN, false, NaN, true, 0.2672612419124244, false, false)), ScalarVariable{Float64, Float64, Float64, Float64}(VariableInfo{Float64, Float64, Float64, Float64}(false, NaN, false, NaN, false, NaN, true, 0.5345224838248488, false, false)), ScalarVariable{Float64, Float64, Float64, Float64}(VariableInfo{Float64, Float64, Float64, Float64}(false, NaN, false, NaN, false, NaN, true, 0.8017837257372732, false, false))]
```

One further thing that is totally unclear to me is whether you ever store tangent vectors (gradients for example). The challenge is then, that this might (in Manifolds) even be of different “shape” than points. Example
* `SVDMPoint` (for fixed rank matrices, the example I would love to be able to provide) – stores an SVD, so U, S, V, where S is a vector (in data structure, mathematically more a diagonal matrix, but stored as a vector)
* corresponding tangents are `UMVTangentvector`s, similar structures for U and V but M is now indeed a n-by-n matrix. 

Well, hopefully JuMP never cares and this part just works ;)


odow:
What's the simplest example you want to be able to build? No JuMP. Pure Manopt.jl

kellertuer:
To not only have the pessimistic side: reasons why it might fit JuMP

* several manifolds can be seen as constrains (the sphere is a unit norm constraint) - though not all, while Stiefel is unit norm vectors, Grassmann is a quotient manifold of that, and these do usually not easily fit into a “constraint thinking” - Grassmann is, however, already possible in the current Array-variant
* Some types points are just wrappers (like for Hyperbolic), all others we currently have can at least be “vectorized” (squeezed into an 1D array), though that is an per-point-type manual thing to do then.
* Even fixed rank with its representation in 3 factors (by the way for numerical stability and ease of implementation) is a constrained thingy: n-by-m matrices of rank p. Turning it into a manifold makes the Riemannian problem unconstrained, so that is maybe nice to have (at the cost of using Manifolds.jl / differential geometry tools)

But again, if it does not fit, I am totally fine deleting the extension as well. I am not here to try to sneak this package into something where it does not belong.

kellertuer:
Ok. Enjoy JumpDev 2025!
Auckland, that sounds very nice.

kellertuer:
I gave this a careful thought again, and with all the arguments stated and discussion we had, I think it is best to remove the JuMP extension for now.

I thought for a while it was cool for visibility, since you really have a large user base, but as a mere mortal it feels impossible to me to maintain such an extension. Furthermore
* manifolds as a constraints are a bit of a stretch modelling wise – though true in several cases.
* matrices are already a stretch, since all JuMP variables are vectors
* typed points and especially also distinguishing points and tangent vectors seems to be a huge effort and is not really supported nor something that seems to be wishful to have in JuMP

All these in mind – I am open to have such a “bridge” between Manopt and JuMP somewhere, but better in an own package somewhere in the “JuMP universe”, I am happy to help with manifold-related things. This way around, the parts that are internal and not (yet) so well documented are “closer” to the package. I would carefully claim that both Manifolds.jl and Manopt.jl are reasonably documented – but of course I am also available to help if someone wants to give this connection a new try.

I will keep this PR open maybe until somewhen in January 2026 and then merge it.

Thanks for all the discussions and efforts.

kellertuer:
I diverted the PR to be collected on a dev-0.6 branch, since we plan to collect a few forthcoming breaking changes for that release. Will merge this when the tests have completed.

odow (approved):
Probably for the best. @blegat and I may revisit if we get around to adding true vector-valued nonlinear operators.

(Un)related: this PR is a great example of why I'm very hesitant to add new extensions in JuMP-related packages. Linking here so I have a back link if I ever need to find this in future: https://github.com/jump-dev/MathOptInterface.jl/pull/2972 and https://github.com/jump-dev/MathOptInterface.jl/pull/2967

kellertuer:
As I mentioned, I am happy to help from the Manopt-side, if you give it a new try. I personally felt too often lost within JuMP. And that is not meant as negative as it sounds. I think the docs for users of JuMP are great and you do a very good job on the development and commenting (even now here on such short note!), just that from the development side, I often was diving deep into the source code of JuMP itself.

To the “vector valued nonlinear operators” – I fear we would especially need variables (and corresponding costs) beyond vectors – into matrices and even arbitrary structs (wrappers of representations or even structured data like for fixed rank an SVD or so).

But maybe some day we continue this adventure then :)

blegat:
Sounds good, I am planning to experiment with vector-valued nonlinear operators in https://github.com/blegat/ArrayDiff.jl/
Now, I need to choose a good name, do you have any policy on the names used by packages outside of JuliaManifolds ? We request that packages outside of jump-dev don't use JuMP in their names for instance.
One possibility is JuMPonManifolds which has a bouncy feel

kellertuer:
I always jump around on manifolds. It indeed sounds a bit bouncy. If so you should name all packages like that (or even use `TranpolinX.jl` and such).

Joke aside (had a long day already since I am currently in JST time zone). Great that you reach out about a name.

I am fine with either `JuMPManoptExt.jl` or `ManoptJuMPExt.jl` (inspired by e.g. `OptimizationManopt.jl` within SciML, which they even chose without asking), and we do not have so much of a Scheme. If it is within the JuMP-org, maybe putting JuMP first is reasonable? On the other hand, the capital letters grouping in the middle also looks a bit strange.

Back to the start … since if you actually do **jump** on a trampoline, the surface does form a manifold and it looks even a bit hyperbolic...
So indeed also just `Trampolin.jl` is fine this me, though maybe a bit harder to discover for users.

_edit:_ Trampolin.jl is indeed a name that is available in the General Registry and though that started as a joke when I started writing it, it is somehow nice. relates to JuMP (is a bit bouncy, yes) and curved.

mateuszbaran:
`BouncyCastle.jl` would also be a fun name :slightly_smiling_face: . But watch out for sharp corners we're going to add soon with #554 :wink: .

More seriously, regarding `Ext` names, I think it's reserved for actual package extensions? If a separate package is required then something like JuMPManopt or JuMPManifolds sounds good. I think it's fine to use words "Manopt"/"Manifolds" in packages outside of JuliaManifolds, though I'd prefer if the first word in the name wasn't something already existing in JuliaManifolds.

PR #552, opened 2025-11-27
Print debug also in initial iteration

JoshuaLampert:
Addresses the discussion started at https://github.com/JuliaManifolds/Manopt.jl/pull/544#discussion_r2539412189. I went through the different debug options and enabled printing at the initial value for those I evaluated it would make sense to print the debug statement at the first iteration. If anything is missing or you don't think it makes sense to print it already at the initial value, I am open to adapt that. The only one I would like to keep is the feasibility one.

### DIFF
--- a/src/plans/debug.jl
+++ b/src/plans/debug.jl
@@ -715,7 +715,7 @@ mutable struct DebugIterate <: DebugAction
             long::Bool = false,
             prefix = long ? "current iterate:" : "p:",
             format = "$prefix %s",
-            at_init::Bool = true,
+            at_init::Bool = false,
         )
         return new(io, format, at_init)
     end
--- a/test/plans/test_debug.jl
+++ b/test/plans/test_debug.jl
@@ -106,7 +106,7 @@ Manopt.get_parameter(d::TestDebugParameterState, ::Val{:value}) = d.value
         @test String(take!(io)) == "Last: 1.000000"
         # Iterate
         DebugIterate(; io = io)(mp, st, 0)
-        @test String(take!(io)) == "p: [3.0, 2.0]"
+        @test String(take!(io)) == ""
         DebugIterate(; io = io)(mp, st, 1)
         @test String(take!(io)) == "p: $(st.p)"
         # Iteration
--- a/test/solvers/test_gradient_descent.jl
+++ b/test/solvers/test_gradient_descent.jl
@@ -36,7 +36,7 @@ using ManifoldDiff: grad_distance
         )
         p = get_solver_result(s)[]
         res_debug = String(take!(my_io))
-        @test res_debug === "p: [-1.5707963267948966] f(x): 1.357071\n"
+        @test res_debug === " f(x): 1.357071\n"
         p2 = gradient_descent(
             M,
             f,


kellertuer:
Thanks!

A small problem might be consistency. I do not mean to prior prints – to me it is fine to “break” old code in this sense, since it just breaks what is seen in the print.

Until now only the cost was printed at the beginning, now you chose a slightly larger set. The two you mentioned, I think could be printed. Furthermore:

* What about the initial point?
* ...or all others in
https://github.com/JuliaManifolds/Manopt.jl/blob/7373e5b92a6d10416ddebefdc607d5ab5139fb4c/src/plans/debug.jl#L1401-L1410

(i.e. excluding the warnings and times)?

I do not have an answer yet, but we could either
* check to choose a new good default
* ...additionally allow to set that? The iterate might (for most manifolds) be too long)

One simple idea, which however expands on complexity of using these symbols would be tuples of symbol and bool (or even a pair?) `(:Cost, true)` or `(:Cost, init => true)` (we could even remodel to pass pair as kwargs...

just a few ideas. If you feel this is too complicated, we stick to just choosing a new default what to print at init.

PS: Cool that you directly adopted mentioning the PR in the changelog :)

JoshuaLampert:
I'm definitely in favor of consistency. I also like the possibility of customization. Since many of the `Debug*` already allow keywords, we could add a boolean keyword `init` to them with sensible default, which can be set by explicitly creating the struct via the constructor. I think I would keep the symbol construction as it is, i.e., it always creates the default. Otherwise it would be inconsistent with how other keyword arguments are handled.
Let's go through the list:
in debug.jl:
* `DebugCost`: Currently printed at `k = 0`. I would add `init = true` as kwarg.
* `DebugChange`: Currently not printed at `k = 0`. I think here is doesn't make sense because for the change we already need a previous value. I would keep it as it is.
* `DebugDivider`: Currently printed at `k = 0`. I would add `init = true` as kwarg.
* `DebugEntry`: Currently printed at `k = 0`. I would add `init = true` as kwarg.
* `DebugFeasibility`: Currently (in master) *not* printed at `k = 0`. I would add `init = true` as kwarg.
* `DebugIfEntry`: Currently printed at `k = 0`. I would add `init = true` as kwarg.
* `DebugEntryChange`: Similar to `DebugChange`. I would keep it as is.
* `DebugGradientChange`. Same again. I would keep it as is.
* `DebugIterate`: Currently not printed at `k = 0`. I would add an `init = false` as kwarg.
* `DebugIteration`: Already handles `k = 0` appropriately. Would keep it as is.
* `DebugMessages`: Would keep it as is.
* `DebugStoppingCriterion`: Currently not printed at `k = 0`. I would guess there will never be a stopping reason before the first iteration started. So I would keep it as is.
* `DebugWhenActive`: I think we can keep it as it is.
* `DebugTime`: At `k = 0`, this initializes the time. So it doesn't make sense to print it there, i.e., I would keep it.
* `DebugWarnIfCostIncreases`: This will never print at `k = 0` because the initial `old_cost` is `Inf`. So keep as is.
* `DebugWarnIfCostNotFinite`: Currently printed at `k = 0`. I would keep it.
* `DebugWarnIfFieldNotFinite`: Same as `DebugWarnIfCostNotFinite`.
* `DebugWarnIfGradientNormTooLarge`: Same as `DebugWarnIfCostNotFinite`.
* `DebugWarnIfStepsizeCollapsed`: Similar to `DebugWarnIfCostNotFinite`.

So the only one for which I would _change_ the default it `DebugFeasibility`. Do you have another opinion for one or more of the above? We could then also discuss other `Debug*` entries located in other files.

JoshuaLampert:
I added `at_init` as discussed to all the debug options in debug.jl. So let's discuss the [other options in other files](https://github.com/search?q=repo%3AJuliaManifolds%2FManopt.jl%20%3C%3A%20DebugAction&type=code):
bundle_plan.jl:
- `DebugWarnIfLagrangeMultiplierIncreases`: Would keep as is because it needs a previous value.

proximal_plan.jl:
- `DebugProximalParameter`: Currently not printed at `k = 0`. I would add `init = true` as kwarg.

primal_dual_plan.jl:
- `DebugDualResidual`: Can probably stay as it needs storage.
- `DebugPrimalResidual`: Can probably stay as it needs storage.
- `DebugPrimalDualResidual`: Can probably stay as it needs storage.
- `DebugDualChange`: Can probably stay as it needs storage.

first_order_plan.jl:
- `DebugGradient`:  Currently not printed at `k = 0`. I would add `init = true` as kwarg.
- `DebugGradientNorm`: Currently not printed at `k = 0`. I would add `init = true` as kwarg.
- `DebugStepsize`: Currently not printed at `k = 0`. I think we do not have `get_last_stepsize` at `k = 0`? So probably keep it as is?

proximal_bundle_method.jl:
- `DebugWarnIfLagrangeMultiplierIncreases`: Can probably stay as it needs an old value.


So I would change `DebugProximalParameter`, `DebugGradient`, and `DebugGradientNorm`.

JoshuaLampert:
Tests pass. So this is ready for a review from my point of view, @kellertuer. Thanks for your input, @kellertuer.

src/plans/debug.jl
```diff
@@ -399,26 +403,27 @@ how to print the entry.
...
-    function DebugEntry(f::Symbol; prefix = "$f:", format = "$prefix %s", io::IO = stdout)
-        return new(io, format, f)
+    at_init::Bool
+    function DebugEntry(f::Symbol; prefix = "$f:", format = "$prefix %s", io::IO = stdout, at_init::Bool = true)
+        return new(io, format, f, at_init)
     end
 end
 function (d::DebugEntry)(::AbstractManoptProblem, st::AbstractManoptSolverState, k)
-    (k >= 0) && Printf.format(d.io, Printf.Format(d.format), getfield(st, d.field))
+    (k >= !d.at_init) && Printf.format(d.io, Printf.Format(d.format), getfield(st, d.field))
```
kellertuer: My only small semantic remark would be, that “Integer larger or equal a boolean” is not so easy to read/parse.

Could we maybe do something like `(k >= (d.at_init ? 0 : 1))` yes for most interpretations that is the same.

JoshuaLampert: Yes, I thought about that. We can certainly use your suggestion. I agree, it's a bit easier to read.

JoshuaLampert:
I just realized that the failing tests came from `DebugIterate`. However, we agreed on setting `at_init = false` as default for `DebugIterate`, but for some reason I set it to `true` by default. So I changed that to `false`, which hopefully also fixes the test.

PR #558, opened 2025-12-10
Fix early stopping in `StopWhenRepeated`

JoshuaLampert:
I just had a case, where the interior stopping criterion of a `StopWhenRepeated` stopping criterion was satisfied at initialization, but not anymore in the first few iterations, i.e., it was not yet repeated `n` times, but the optimization was stopped nonetheless, which was confusing to me. This PR fixes this by not returning early in the initialization.


kellertuer (approved):
Now it looks better. Thanks for spotting that edge case.

Do you need that soon? Then I could release a new version in the evening.

PR #577, opened 2026-02-12
Add point capability to `_produce_type` to improve GPU compatibility

mateuszbaran:
This is a draft of the enhancement discussed in #574 .

cc @zazabap


kellertuer:
> intentionally verify that the solver works with an immutable p?

Yes, that test is intentionally, since `Euclidean(1)` would be vectors of length 1, but `Euclidean()` is our convention for reals/immutable. Internally we could `Ref` that – think we even do that already.

kellertuer (commented):
A small comment

mateuszbaran:
This is now ready up to that one comment. For now I think testing by checking that Float32 is kept should be a reasonable solution. For more comprehensive tests there is https://github.com/JuliaGPU/GPUArrays.jl/tree/master/lib/JLArrays which circumvents the lack of GPU on Github CI runners to a larger degree.

kellertuer (approved):
lgtm.

PR #583, opened 2026-03-03
Fix RTR with Float32

mateuszbaran:
Fixes #582


kellertuer (approved):
thanks for quickly fixing it.
The `convert` lines look a bit verbose, but maybe writing that a bit shorter / condensed is also something we can check on a later PR :)

PR #596, opened 2026-04-15
Better support for custom point types in conjugate_gradient_descent

### LINKED ISSUE #533: Armijo linesearch does not allocate the right candidate point type
If you try to run a code on typed points like

```
using Manopt, Manifolds
using ManifoldDiff: grad_distance
M = Hyperbolic(2)
data = PoincareBallPoint.([ [0.1, 0.2], [0.3,0.25], [0.35,0.4] ])
n = length(data)
f(M, p) = sum(1 / (2 * n) * distance.(Ref(M), Ref(p), data) .^ 2)
grad_f(M, p) = sum(1 / n * grad_distance.(Ref(M), data, Ref(p)));
m1 = gradient_descent(M, f, grad_f, data[1])
```

This errors since the candidate point within Armijo line search is an array (from the default representation on Hyperbolic) due to this line

https://github.com/JuliaManifolds/Manopt.jl/blob/962ae4912de6def552368a09c39c7190ce87744b/src/plans/stepsize/stepsize.jl#L106

allocating a default random point. A solution could be to pass the initial point to step sizes as well (optionally, so they rand otherwise to be nonbreaking) and use that points type to allocate the candidate.

Might affect other step sizes as well that are linesearch based (i.e. need a candidate).

mateuszbaran:
Fixes https://github.com/JuliaManifolds/Manopt.jl/issues/533#issuecomment-4251018134 .

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,13 +6,17 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.5.35] April 3, 2026
+## [0.5.35] April 16, 2026
 
 ### Changed
 
 * Improved formatting of the references in the Readme.md (#586)
 * Bump compat for RecursiveArrayTools.jl to include version 4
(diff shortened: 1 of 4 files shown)


kellertuer (approved):
lgtm, thanks for the fix.

Issue #595, opened 2026-04-13
Proposal: stable result interface or result-builder hook for downstream domain packages


Harumaru58:
Hi Ronny,
I hope you had a peaceful easter break! 
While working on stabilizing our package, I have a suggestion to make since you previously encouraged. :)

Here goes. Our package exposes custom public result types such as `CPDResult`, `BTDResult`, and `ApproxResult`. These wrap Manopt solver outputs into decomposition-friendly results with fields like components, factors, reconstruction helpers, and solver metadata.

At the moment, from upstream, downstream packages need to call `solve(...)`, inspect the returned result object, pull out fields like point / cost / grad norm / iterations / converged / solver info, and then convert that into their own result type. This works, but it means downstream packages end up depending on the concrete shape of Manopt result values rather than on a stable interface.

**Would you be open to providing a stable extension point for custom result construction?**

What we need is not tensor-specific result types in Manopt, but a stable way for downstream packages to build their own typed results without depending on internal result layout.

A few possible shapes that would help:
- stable accessors like `solution_point(result)`, `solution_cost(result)`, `solution_grad_norm(result)`, `solution_iterations(result)`, `solution_converged(result)`, `solution_solver(result)`
- a `result_builder` hook in `solve(...)`
- or a small stable summary result type with a guaranteed schema for those common fields

Any of those would make it much easier for downstream domain packages to wrap Manopt results safely and consistently.

Thanks!

kellertuer:
Hi,

thanks for the ideas!
I did indeed not have any pause for easter, since I am still in Tokyo (travelling back this week) – and here there is no Easter. But when I am back home, I make sure to take a break.

 Some of the things you want to have are already there:

## Accessing “Things from the state” like the minimiser `p*`

Since for the high-level interfaces (so not calling `solve!` but using `gradient_descent!`) you can “activate” to get more information instead of just the minimiser (setting `return_state = true`) there is also the stable and documented `get_solver_result(state)` – which (a) for a state accesses the solver result (thought that is oftenn really just `state.p`) – and (b) for everything else than the state it is the identity. This way it is stable independent of whether you had `return_state = true` or not. See https://manoptjl.org/stable/plans/state/

Further things to access (also documented on the `state` page)

* `get_iterate` sounds nearly synonymous with the `get_solver_result` (and is called by the result thingy), but acts a bit more low level
* `get_gradient` to access a last evaluated gradient if the solver has this (often this is `state.X`)
[...]

PR #599, opened 2026-04-24
introduce a stopped_at function for solver states.

kellertuer:
As proposed in #595 this introduces a function to access the iteration when a solver stopped.

### DIFF
--- a/test/plans/test_gradient_plan.jl
+++ b/test/plans/test_gradient_plan.jl
@@ -16,6 +16,7 @@ using ManifoldsBase, Manopt, Test
         stopping_criterion = StopAfterIteration(20),
         stepsize = Manopt.ConstantStepsize(M),
     )
+    @test stopped_at(gst) == -1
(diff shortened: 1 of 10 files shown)


test/plans/test_gradient_plan.jl
```diff
@@ -16,6 +16,7 @@ using ManifoldsBase, Manopt, Test
         stopping_criterion = StopAfterIteration(20),
         stepsize = Manopt.ConstantStepsize(M),
     )
+    @test stopped_at(gst) == -1
```
mateuszbaran, line 19: Maybe let's also add a second test after the solver is stopped?

mateuszbaran (approved):
LGTM up to one more test that I think would be nice to have.

test/plans/test_gradient_plan.jl
kellertuer: Ok, I can check where that fits well.

PR #600, opened 2026-04-25
Add customizable subsolver direction to ALM

mateuszbaran:
I mostly want to be able to set the memory to something less than 30.

### DIFF
--- a/src/solvers/augmented_Lagrangian_method.jl
+++ b/src/solvers/augmented_Lagrangian_method.jl
@@ -325,6 +325,8 @@ $(_kwargs(:sub_problem; default = "`[`DefaultManoptProblem`](@ref)`(M, sub_objec
 $(_kwargs(:sub_state; default = "`[`QuasiNewtonState`](@ref)` ")), where more precisely
   as quasi newton method, the [`QuasiNewtonLimitedMemoryDirectionUpdate`](@ref) with [`InverseBFGS`](@ref) is used.
 * `sub_stopping_criterion::StoppingCriterion=`[`StopAfterIteration`](@ref)`(300)`$(_sc(:Any))[`StopWhenGradientNormLess`](@ref)`(ϵ)`$(_sc(:Any))[`StopWhenStepsizeLess`](@ref)`(1e-8)`,
+* `sub_direction_update::AbstractQuasiNewtonDirectionUpdate=`[`QuasiNewtonLimitedMemoryDirectionUpdate`](@ref)`(M, copy(M, p), InverseBFGS(), min(manifold_dimension(M), 30))`:
+  the direction update rule for the sub solver, where by default a limited memory quasi-Newton method with inverse BFGS update is used.
 
 
 For the `range`s of the constraints' gradient, other power manifold tangent space representations,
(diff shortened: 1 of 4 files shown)


src/solvers/augmented_Lagrangian_method.jl
```diff
@@ -325,6 +325,8 @@ $(_kwargs(:sub_problem; default = "`[`DefaultManoptProblem`](@ref)`(M, sub_objec
 $(_kwargs(:sub_state; default = "`[`QuasiNewtonState`](@ref)` ")), where more precisely
   as quasi newton method, the [`QuasiNewtonLimitedMemoryDirectionUpdate`](@ref) with [`InverseBFGS`](@ref) is used.
 * `sub_stopping_criterion::StoppingCriterion=`[`StopAfterIteration`](@ref)`(300)`$(_sc(:Any))[`StopWhenGradientNormLess`](@ref)`(ϵ)`$(_sc(:Any))[`StopWhenStepsizeLess`](@ref)`(1e-8)`,
+* `sub_direction_update::AbstractQuasiNewtonDirectionUpdate=`[`QuasiNewtonLimitedMemoryDirectionUpdate`](@ref)`(M, copy(M, p), InverseBFGS(), min(manifold_dimension(M), 30))`:
+  the direction update rule for the sub solver, where by default a limited memory quasi-Newton method with inverse BFGS update is used.
```
kellertuer:
```suggestion
  the direction update rule for the default sub solver, a quasi-Newton method with limited memory inverse BFGS. This keyword is used within the `sub_state`, so it has no effect, if you set the sub solver manually.
```

This is imprecise. If you set the `sub_state=` yourself, the new keyword is ignored.

mateuszbaran: Thanks, that a good point.

kellertuer:
I think instead of manually “passing down” more and more keywords (and introducing them with a `sub_`, I would prefer if you just use the `sub_kwargs` keyword. Would that be ok as well?

mateuszbaran:
That would make sense but `sub_kwargs` already exists for ALM but it specifically excludes `direction_update`. Having `direction_update` in `sub_kwargs` would be a breaking change. I can also add `direction_update` to default `sub_kwargs` for Manopt 0.6 only.

kellertuer:
That I do not understand why does it exclude the direction update?

Overall I am just unsure whether it is better to use sub_kwargs (keep the ALM keywords at a finite amount) or introduce more and more keywords to a solver that is a bit crowded on those already anyways.

src/solvers/augmented_Lagrangian_method.jl
```diff
@@ -460,9 +465,7 @@ function augmented_Lagrangian_method!(
                 M;
                 p = copy(M, p),
                 X = zero_vector(M, p),
-                direction_update = QuasiNewtonLimitedMemoryDirectionUpdate(
-                    M, copy(M, p), InverseBFGS(), min(manifold_dimension(M), 30)
-                ),
+                direction_update = sub_direction_update,
                 stopping_criterion = sub_stopping_criterion,
                 stepsize = default_stepsize(M, QuasiNewtonState),
                 sub_kwargs...,
```
kellertuer, line 471: The sub_kwargs are passed in here _after_ `direction_update`, so setting that with the sub_kwargs should be fine.

mateuszbaran: Ah, I didn't know that's a thing in Julia -- then sub_kwargs is indeed enough. Thanks for pointing it out!

kellertuer: last keyword wins is what works here, see the last paragraph of https://docs.julialang.org/en/v1/manual/functions/#Keyword-Arguments : explicitly having a kw twice is not allowed, but implicitly the last one can overwrite earlier defaults.

Thanks for closing this, since the number of kwargs stays smaller this way :)

PR #606, opened 2026-05-19
Update prox-grad backtracking condition

hajg-ijk:
The `:convex` backtracking strategy for `proximal_gradient_method` now entails a slightly different condition whenever the upper bound on the sectional curvature of the manifold, input via the `k_max` kwarg, is positive. This comes with a "tolerance" type parameter, `δ`, which must be positive.


kellertuer:
Is this breaking existing behaviour? Then it should probably be part of the forthcoming 0.6 release?

Besides that one of your stopping criteria is now not code cov'ed on 2 lines, `StopWhenGradientMappingNormLess` never reaches the point where it reports true.

hajg-ijk:
> Is this breaking existing behaviour? Then it should probably be part of the forthcoming 0.6 release?
> 
> Besides that one of your stopping criteria is now not code cov'ed on 2 lines, `StopWhenGradientMappingNormLess` never reaches the point where it reports true.

I don't think this is breaking. But I am a bit puzzled by those 2 missing lines. Will try to cover them.

kellertuer (approved):

PR #607, opened 2026-05-20
Removal of the asymptote helper functions -> ManifoldAsymptote.jl

kellertuer:
For the forthcoming breaking change it is a good chance to also remove the color package dependencies by moving the asymptote exports to their own small package at 

https://github.com/JuliaManifolds/ManifoldAsymptote.jl

However, this should only be merged once that package is registered to not leave users without any asy rendering.


kellertuer:
This should be good to be merged into dev0.6 by now, though the new package is only available after the waiting period (in 3 days), all functionality has been moved there, is now even tested (was not tested here before) and even includes a bugfix.

mateuszbaran (approved):

PR #609, opened 2026-06-02
Don't print debug actions when `k=-1`

JoshuaLampert:
This fixes a small bug, when using some debug actions (`DebugTime`, `DebugEntryChange`, `DebugWarnIfCostNotFinite`, `DebugWarnIfFieldNotFinite`, and `DebugWarnIfGradientNormTooLarge`) together with `DebugEvery` (with `always_update = true`), where the actions ignore the `DebugEvery` and print their output at _each_ iteration instead of only every `every` iterations. This is because they don't add a guard for `k = -1`, which is passed by `DebugEvery` for the iterations that should be skipped. For the warnings it might be debatable whether they should intentionally skip the `DebugEvery` or not, but since `DebugWarnIfCostIncreases` and `DebugWarnIfStepsizeCollapsed` already had the `k < 0` included, I also added it to the other three warning debug actions for consistency. I can also remove it consistently for all five warning debug actions if preferred.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* a small bug where debug statements were printed even though they should not be due to `DebugEvery` and unified warnings to print independent of `DebugEntry`. (#609)
+* a small bug where debug statements were printed even though they should not be due to `DebugEvery` and unified warnings to print independent of `DebugEvery`. (#609)
 
 ## [0.5.38] May 19, 2026
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,12 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.39] June 2, 2026
+
+### Fixed
+
+* a small bug where debug statements were printed even though they should not be due to `DebugEvery`. (#609)
+
 ## [0.5.38] May 19, 2026
 
 ### Changed
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,11 +6,11 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.5.39] June 2, 2026
+## [0.5.39] June 3, 2026
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


JoshuaLampert:
Failing tests with Julia nightly seem unrelated and an issue of HSL.jl.

kellertuer:
For the “normal” debugs I agree, the idea would be to never print with negative k, the implementation was not consistent there – thanks for spotting and making it consistent!

For the Warnings, you are right that was also not consistent. but I am just not sure which mode is the better one: If one activates / adds a warning I think always warning would be better?
Otherwise a user might say “I ask the algorithm to warn me and it did not.”

So we could also remove the `k < 0` checks from the two that currently have it – instead. I think this has until now really just not been fixed / documented, which warning mode is the better one.

kellertuer:
...and yes, I have not yet understood the nightly errors, but they are there for a while already and not your fault. I should maybe check whom to ask on slack somewhen.

JoshuaLampert:
> For the Warnings, you are right that was also not consistent. but I am just not sure which mode is the better one: If one activates / adds a warning I think always warning would be better?
Otherwise a user might say “I ask the algorithm to warn me and it did not.”

Yes, I mostly agree with that. But you could also say "I ask the algorithm to only give debug output every so and so iterations via `DebugEvery` and it did not" :)
But I'll remove the `k < 0` from all warnings.

kellertuer:
Sure, as I wrote, both sides have their good arguments.

This looks good now. Could you specify the two aspects we unified the actions to in the changes as well?

Then this should be good to go and I can register it after merge.

JoshuaLampert:
Done in [35dc2c5](https://github.com/JuliaManifolds/Manopt.jl/pull/609/commits/35dc2c557b6c3a07c6c11445b7a62809aa0dc362).

Changelog.md
```diff
@@ -6,6 +6,12 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.39] June 3, 2026
+
+### Fixed
+
+* a small bug where debug statements were printed even though they should not be due to `DebugEvery` and unified warnings to print independent of `DebugEntry`. (#609)
```
kellertuer:
```suggestion
* a small bug where debug statements were printed even though they should not be due to `DebugEvery` and unified warnings to print independent of `DebugEvery`. (#609)
```

you also ment _every_ not _entry_ for the second ?

JoshuaLampert: Whoops, yes, sure. Thanks for catching that.

kellertuer (approved):
Nice. As soon as the tests pass, I can merge this and register the new version for you.

Thanks for spotting this and unifying the behaviour.

PR #615, opened 2026-06-14
change title of embedding objective tutorial

araujoms:
I think the current title can only be understood by people who know enough about the subject that do not need the tutorial. I was specifically looking for this information, and couldn't find the tutorial, just searched for it in the wild. Only after @kellertuer linked it to me that I found it.


kellertuer:
A challenge in that title is, that it really only works if the manifold is embedded in some Euclidean space. But I can think about a better title.

kellertuer:
...and yes I mainly wrote that tutorial after I had (a) implemented both the gradient and Hessian conversion (and before that understood them from Nicolas Boumals book) – and explained it 2-3times somewhere on Slack, because this came up a few times.

So yeah the title can surely be improved.

tutorials/EmbeddingObjectives.qmd
```diff
@@ -1,5 +1,5 @@
 ---
-title: "How to define the cost in the embedding"
+title: "Obtain the Riemannian gradient and Hessian"
```
kellertuer:
```suggestion
title: "Obtain the Riemannian gradient and Hessian from their Euclidean analogues in the embedding"
```

For the menu point I can not yet bring that down to something short and precise, but here is an idea at least for the title. What I am missing is, that one does need the gradient / Hessian in the embedding.

“Obtain” alone sounds a bit like one either can generically magically compute it – or the tutorial is a complete repetition of a text book on both.

araujoms: I think it's fine for titles to not be completely precise, they should just be good guides to the content.

kellertuer:
Ah now only the Changelog is erroring. It would be great if you could add an entry to the Changelog for a next unreleased version (same as you did for Manopt on your last PR – and again referring to this PR in brackets for completeness :) )

kellertuer:
Perfect, thanks! I might “steal” the 5.40 from you again and merge it into the other PR on the 0.6 then in then end, because that is what we are currently trying to finish :)

PR #617, opened 2026-06-23
Introduce a robust Levenberg-Marquardt

kellertuer:
Following https://arxiv.org/abs/2606.23560, this PR introduces a robust Levenberg-Marquardt algorithm for Riemannian manifolds. (see also #332). This should be the last PR to be added to #605.


docs/src/plans/objectives/constrained.md
```diff
@@ -0,0 +1,49 @@
...
+```@meta
+CurrentModule = Manopt
+```
+
+```@docs
+ConstrainedManifoldObjective
+ManifoldConstrainedSetObjective
+```
+
+It might be beneficial to use the adapted problem to specify different ranges for the gradients of the constraints
```
mateuszbaran: This comment is very confusing to me, probably due to lack of context.

docs/src/solvers/LevenbergMarquardt.md
```diff
@@ -15,6 +15,49 @@ LevenbergMarquardt!
...
+Manopt.add_linear_operator_coord!
+```
+
+### Internal structures
+
+The following structure helps to avoid allocating unnecessary zero vectors.
+
+```@docs
+Manopt.ZeroTangentVector
+```
```
mateuszbaran, line 59: Maybe let's move it straight to ManifoldsBase?

docs/src/notation.md
```diff
@@ -5,6 +5,10 @@ with the following additional parts.
 
 | Symbol | Description | Also used | Comment |
 |:--:|:--------------- |:--:|:-- |
-| ``\operatorname{arg\,min}`` | argument of a function ``f`` where a local or global minimum is attained | |
-| ``k`` | the current iterate | ``ì`` | the goal is to unify this to `k` |
+| ``\mathbf{a}`` | vectors in Euclidean space | ``\mathbf{b}, \mathbf{c}`` | |
+| ``\mathbf{A}`` | a matrix | ``\mathbf{B}, \mathbf{C}`` | |
+| ``\mathcal{A}, \mathcal{L}`` | linear operators, usually on tangent spaces | | |
+| ``\mathcal{C}`` | a (geodesically) convex set | | |
+| ``\operatorname*{arg\,min}_v`` | argument/variable ``v`` of a function ``f`` where a local or global minimum is attained | |
```
mateuszbaran: $f$ is not mentioned in the symbol.

docs/src/notation.md
```diff
@@ -5,6 +5,10 @@ with the following additional parts.
...
+| ``k`` | the current iterate | ``i`` | the goal is to unify this to `k` |
```
mateuszbaran, line 13:
```suggestion
| ``k`` | the current iteration | ``i`` | the goal is to unify this to `k` |
```
Some authors refer to the current solution as "iterate".

ext/ManoptManifoldsExt/manifold_functions.jl
```diff
@@ -60,6 +67,15 @@ function max_stepsize(M::FixedRankMatrices, p)
...
 max_stepsize(M::FixedRankMatrices) = manifold_dimension(M)
 
+"""
+    max_stepsize(::SymmetricPositiveDefinite, p)
+
+Return the maximum stepsize on the [`SymmetricPositiveDefinite`](@extref) manifold.
+The injectivity radius of `SymmetricPositiveDefinite` is infinite, but we return the
+logarithm of the maximum floating-point number to avoid numerical issues.
+"""
+max_stepsize(::SymmetricPositiveDefinite, p) = log(floatmax(eltype(p)))
```
mateuszbaran, line 77: This might be worth mentioning in the changelog.

Changelog.md
```diff
@@ -9,17 +9,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 * Stopping criteria can now be initialized using `initialize_stepsize!`, similar to solvers. (#554)
 
 ### Changed
 
+* In the [Riemannian Levenberg Marquardt algorithm](https://manoptjl.org/stable/solvers/LevenbergMarquardt/)t the `η` parameter has been renamed to `candidate_acceptance_threshold`, `β` to `damping_increase_factor` and `β_reduction` to `damping_reduction_factor`. (#617)
+* the constructor for the [Levenberg-Marquardt state](https://manoptjl.org/stable/solvers/LevenbergMarquardt/#Manopt.LevenbergMarquardtState) has been unified with the remaining states, to take the `sub_problem` and `sub_state` arguments as second and third positional arguments, respectively. (#617)
+* the keyword `initial_jacobian_f` within `LevenbergMarquardt` is unified in naming to the residual values vector and called `initial_jacobian_matrices`. If you call `LevenbergMarquardt` with a single vector component, also a single matrix is allowed. (#617)
+* an internal field of the solver state of Levenberg-Marqwuardt was called `jacobian_f` the same as the functions whose result it meant to cache if applicable. To distinguish both, the field is now called `jacobian_matrices`. (#617)
```
mateuszbaran, line 30:
```suggestion
* an internal field of the solver state of Levenberg-Marquardt was called `jacobian_f` the same as the functions whose result it meant to cache if applicable. To distinguish both, the field is now called `jacobian_matrices`. (#617)
```

src/plans/conjugate_residual_plan.jl
```diff
@@ -205,6 +266,9 @@ A state for the [`conjugate_residual`](@ref) solver.
 * `α::R`: a step length
 * `β::R`: the conjugate coefficient
 $(_fields(:stopping_criterion; name = "stop"))
+* `warm_start`: whether to warm start or not when reusing this state, i.e.
+  * `true` (default): means we reuse the values in `X` on initialization and set the remaining terms accordingly. This involved one call to the objectives linear system and right hand side.
+  * `false`: Initialize `X` to the zero vector and hence `d=r=-b(p)`, but we avoid evaluating the linear operator.
```
mateuszbaran, line 271: Maybe it would be a good idea to mention warm start in the changelog?

src/plans/nonlinear_least_squares/nls_general_plan.jl
```diff
@@ -0,0 +1,1293 @@
...
+in the square root already if ``ρ(s)'' < 0`` for stability reason, which means setting ``α = 0``.
+In the case ``s = 0`` we also set the operator scaling ``α / s = 0``.
+
+This function offers two `mode`s
+- `:Default` keeps negative ``ρ''(s) < 0`` but makes sure the square root is well-defined.
+- `:Strict` set ``α = 0`` when ``ρ''(s) < 0`` or when ``s = 0``
+"""
+function get_LevenbergMarquardt_scaling(
+        ρ_prime::Real, ρ_double_prime::Real, FkSq::Real,
+        threshold::Real = 1.0e-5, mode::Symbol = :Default
```
mateuszbaran: Since we haven't found any case where `:Default` is better than `:Strict`, I think it makes more sense to have `:Strict` as the default.

src/plans/robustifiers.jl
```diff
@@ -0,0 +1,442 @@
...
+"""
+    (a, b, c) = get_robustifier_values(ρ::AbstractRobustifierFunction, x::Real)
+
+Evaluate the robustifier ``ρ`` and its first two derivatives at `x`.
+
+# Returns
+A tuple `(a, b, c)` with
+* `a = ρ(x)`
+* `b = ρ'(x)`
+* `c = ρ''(x)` (might be `missing` if not defined)
```
mateuszbaran: Do we have a robustifier with missing second derivative? That doesn't sound particularly useful.

mateuszbaran (commented):
A few small-ish comments 🙂

src/plans/robustifiers.jl
kellertuer: I am not sure why I included that, sure we can remove it.

docs/src/plans/objectives/constrained.md
kellertuer: I also do not understand what this meant to refer to ;)
I will just remove it

docs/src/solvers/LevenbergMarquardt.md
kellertuer: Sure, we could do that.

docs/src/notation.md
kellertuer: Right, I added `f(v)`.

docs/src/notation.md
kellertuer: I think all do, this is a typo.

ext/ManoptManifoldsExt/manifold_functions.jl
kellertuer: Sure, it can be considered breaking even.

src/plans/nonlinear_least_squares/nls_general_plan.jl
kellertuer: That sounds reasonable, yes.

src/plans/conjugate_residual_plan.jl
kellertuer: Sure I'll add a line.

src/plans/nonlinear_least_squares/nls_general_plan.jl
kellertuer: I also changed that default up in the `Levenberg-Marquardt(...)` definition

docs/src/solvers/LevenbergMarquardt.md
kellertuer: Hm, I fear it is a bit preliminary to do that.

* `is_vector` would have to be thoroughly defined
* `X+Y` does not yet work if either of them is a zero tangent vector

and probably several other aspects … it should maybe be also a subtype of the abstract tangent vector type (which is a const – with a parameter?)

I spent 30 minutes looking into that and would conclude this is more work than I expect and it would postpone this PR for quite a while. So Lets maybe open an issue in ManifoldsBase and collect what such a type would need to have first?

src/plans/nonlinear_least_squares/nls_general_plan.jl
kellertuer: HM, it makes about 10 solver runs behave very differently, so I think there is quite a difference. Is this necessary to be changed? Then I have to dive into these errors – which are really different results on solver runs...

_edit:_ and if strict becomes the default having a second one called “Default” is also misleading. I'll revert this vor now.

docs/src/solvers/LevenbergMarquardt.md
mateuszbaran: OK, then let's solve this later.

src/plans/nonlinear_least_squares/nls_general_plan.jl
mateuszbaran: Does any of those failures actually indicate that the results are worse? `:Default` could be renamed to `:Full` or `:Complete` for example.

docs/src/solvers/LevenbergMarquardt.md
kellertuer: Perfect – I still think, similar to the identity element we came up with for Lie groups, this is a very good idea – and the (maybe a bit minimalistic) start here was a good initial point, but I would like to take it a bit slower on ManifoldsBase then, to make it “as smooth” as the Lie group identity.

src/plans/nonlinear_least_squares/nls_general_plan.jl
kellertuer: I could not figure out completely. Is it maybe enough to set `:Strict` as default in the solver interface then (and maybe the state - but nowhere else)?

I am not good in naming, so I am not 100% sure what to call a default mode where we do not “cut off” alpha.  `:Normal`?

mateuszbaran: `:Normal` is also fine. I think we need to do review all occurrences of `:Strict` and `:Default` in Manopt.jl

kellertuer: I tried to do that once, but can try again ;) 
If strict should be default, default should not be called that, so let's take normal.

mateuszbaran: I can try looking into test failures next time.

kellertuer: I will give it another try and check all `mode`, `:Default`, and `:Strict` occurrences and adapt them.

kellertuer: Finished in https://github.com/JuliaManifolds/Manopt.jl/pull/617/commits/e7fefcd0e0b44fbd80979df6ad6715eb253a1e02.

mateuszbaran (approved):
I think this is good enough now 🙂

PR #625, opened 2026-07-10
Use ZeroVector from ManifoldsBase

mateuszbaran:
Completes https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/279 .


kellertuer (approved):
lgtm.

kellertuer:
I think this can also wait a bit and we can register a new version when we have a feature or two additionally. Thanks for adapting it :)

PR #626, opened 2026-07-10
Add callbacks to all solver states and interfaces, and start a developer guide section

### LINKED ISSUE #611: Callback in all  states
While the main design decision in Manopt are decorators, it might be nice to have Callbacks as well.
With the modularity of Manopt, it is maybe a bit tricky to do these. Here is an idea, that could even be used in debug/record then as well.

An algorithm declares “Callback anchors” within the `solver_step!` function – maybe using a macro like

```
@callback_anchor :AnchorName
```

by default this generates a no-op call to the “callback-dispatch” or in other words “registers” the anchor.
By default this dispatch would just be a no-op (maybe something like `callback(Val(:AnchorName), problem, state, iteration)`
An `CallbackState` could then be a decorator with a dictionary of callbacks for anchors.

This would allow anchors for example like

* `:BeforeSubsolver`
* `:AfterSubSolver`

or any other reasonable phase within one step of the solver. The default anchors could already be the ones debug/record themselves have (before/after iteration, before/after algorithm)

The signature of a callback function would hence always be `cbf(problem, state, iteration)`, though one could probably also allow “simple” ones as `cbfs()`.

On the high-level interfaces, this would take the same route as debug/record and one could maybe specify them as pairs

```
callbacks = [ :AnchorName => cbf1, :AnotherAnchorName => cfb2]
```

kellertuer:
This PR introduces callbacks into the solver interface by adding a field to all solvers.
The documentation is refactored to some extend to split the former`plan` into a Developer guid (`base/`) and the library section of common elements that solvers share (`commons`) – which is also reflected in the code organisation.

## Solvers to refactor

* [x] `adaptive_regularization_with_cubics` (`AdaptiveRegularizationState`)
* [x] `alternating_gradient_descent` (`AlternatingGradientDescentState`)
* [x] `augmented_Lagrangian_method` (`AugmentedLagrangianMethodState`)
* [x] `convex_bundle_method` (`ConvexBundleMethodState`)
* [x] `ChambollePock` (`ChambollePockState`)
* [x] `cma_es` (`CMAESState`) - state is now also exported
* [x] `conjugate_gradient_descent` (`ConjugateGradientDescentState`)
* [x] `conjugate_residual` (`ConjugateResidualState`)
* `CoordinatesNormalSystemState` not necessary
* [x] `cyclic_proximal_point` (`CyclicProximalPointState`)
* [x] `difference_of_convex_algorithm` (`DifferenceOfConvexState`)
* [x] `difference_of_convex_proximal_point` (`DifferenceOfConvexProximalState`)
* [x] `DouglasRachford` (`DouglasRachfordState`)
* [x] `exact_penalty_method` (`ExactPenaltyMethodState`)
* [x] `Frank_Wolfe_method` (`FrankWolfeState`)
* [x] `gradient_descent` (`GradientDescentState`)
* [x] `gradient_sampling` (`GradientSamplingState`) state is now also exported 
* [x] `interior_point_Newton` (`InteriorPointNewtonState`)
* [x] `LanczosState`
* [x] `LevenbergMarquardt` (`LevenbergMarquardtState`)
* `LevenbergMarquardtBoxSubsolver` not necessary
* [x] `mesh_adaptive_direct_search` (`MeshAdaptiveDirectSearchState`)
* [x] `NelderMead` (`NelderMeadState`)
* [x] `particle_swarm` (`ParticleSwarmState`)
* [x] `primal_dual_semismooth_Newton` (`PrimalDualSemismoothNewtonState`)
* [x] `projected_gradient_method` (`ProjectedGradientMethodState`)
* [x] `proximal_bundle_method` (`ProximalBundleMethodState`)
* [x] `proximal_gradient_method` (`ProximalGradientMethodState`)
* [x] `proximal_point` (`ProximalPointState`)
* [x] `quasi_Newton` (`QuasiNewtonState`)
* `StepsizeState` not necessary, mainly serves as an interims storage.
* [x] `stochastic_gradient_descent` (`StochasticGradientDescentState`)
* [x] `subgradient_method` (`SubGradientMethodState`)
* [x] `truncated_conjugate_gradient_descent` (`TruncatedConjugateGradientState`)
* [x] `trust_regions` (`TrustRegionsState`)
* [x] `vectorbundle_newton` (`VectorBundleNewtonState`)

In the end
* [x] remove the warning in `get_callbacks` once all states have that implemented, cf. https://github.com/JuliaManifolds/Manopt.jl/blob/15e7e547ba50497e4f5bd1fd884b89516b6952d8/src/base/state/callback.jl#L48-L52 – then this PR is even nonbreaking, since this accessor has a “default empty return”


kellertuer:
@mateuszbaran The function in 

https://github.com/JuliaManifolds/Manopt.jl/blob/d269eac6d9b8830eacd49d365c17d706f85146be/src/solvers/quasi_Newton.jl#L868-L885

seems to have lost code cov. Any idea why? I think for now I mainly used `get_cost(M, objective, p)` and maybe there exists the generic one that extracts `M` and the `objective` from a problem. But one that also uses a state – I did not notice this before. Is there a special reason to have this?
(maybe it is also just Codecov being drunk again?)

_edit:_ Maybe really just a hiccup.

mateuszbaran:
I'd personally not worry about coverage until actual CI tests are fully passing. This function is just a minor optimization to avoid re-computing objective after a line search.

kellertuer:
Sure. 

Yeah I was mainly surprised because until now that signature form was not used and it looks hence a bit strange to me, but I see where you get the idea from and what it is probably for.

kellertuer:
If this is not so urgent, then I can continue a bit refactoring the docs; I can also do that on a next PR though.

Besides looking at the changes once more, we should also properly deprecate the `callback=` keyword in the decoration of a state (and test the warning instead of showing one in the tests) – and deprecate the `DebugCallback` action.

mateuszbaran:
I've added a small tutorial for callbacks, fixed coverage regression, and fixed one bug I've encountered while fixing the regression (HZ status summary not working right). `provided_callbacks` maybe should be exported now?

kellertuer:
I think we can postpone further rework on the docs a bit and merge this.

Reworking the docs will also require a bit of moving function definitions around – I want to get rid of the plan/ folder basically, but that should maybe not delay this PR.

tutorials/HowToCallback.qmd
```diff
@@ -0,0 +1,173 @@
+
+---
+title: "How to use callbacks"
+author: Ronny Bergmann
+---
+
+This tutorial illustrates how to use custom callbacks.
+`Manopt.jl` 0.6.2 introduces a generic way to work with callbacks within solvers.
+While simple [debug](HowToDebug.md) and [recording](HowToRecord.md) are available,
+callbacks offer a more flexible war to interact with a solver run.
```
mateuszbaran:
```suggestion
callbacks offer a more flexible way to interact with a solver run.
```
Maybe let's be a bit less belligerent 😉

mateuszbaran (approved):
Sure, I think this looks good now (up to that one character 🙂 )

tutorials/HowToCallback.qmd
kellertuer: ops. Yes.

kellertuer:
Nice. I am just leaving right now, but I could bump version and such tomorrow morning and then merge and register.

PR #630, opened 2026-07-26
Refactor in-place dispatch route – improve developer docs

### LINKED ISSUE #610: Work towards “in-place first” internally
For the high-level interfaces it is nice that
```
gradient_descent(M, f, grad_f)
```

works for the allocating gradient `grad_f(M,p)` by default.
This aims to be easier to use. 

Internally this is a bit chaotic since
* one always has to distinguish allocating and in-place everywhere.
* an objective can only be one of a kind. For example for the Hessian, both gradient and Hessian have to be the same kind of function (alloc vs nonalloc)

Since the first step is nice for first-time users, keeping that default would be nice; but internally always preferring in-place would be nice.

One possibiliy is maybe to “wrap allocating early” in a function wrapper

```
struct AllocationFunctionWrapper{T,F}
function::F
end
  # Gradient example:
  function (f:: AllocationFunctionWrapper)(M, X, p)
  # evaluate allocation function and copy over
  return copyto!(M, X, p, f.function(M,p)
end
```

where `T` could be a type to determine different things to do (as opposed to copying over the gradient), though most often it is probably just a tangent vector or a point.
Then internally all Objectivbes could always assume having in-place functions; their constructors could maybe be tolerant still.

This is an issue more for the long-time, since it would require a humungous rework.

### LINKED ISSUE #631: `hasconverged(::StopWhenAll)` reports `false` for nested criteria
Bumped into this when looking a bit more at `AlgorithmsInterface.jl`:

The "are all criteria active" `hasconverged` uses a `length` (non-recursive) and compares it to `get_active_stopping_criteria` (recursive), so this can mismatch when combining `&` and `|`.
Current tests only have `s1 & s2` or `s1 | s2`, not `s1 & s2 | s3`.

MWE:

```julia
s1 = StopWhenGradientNormLess(1.0e-4)   # indicates_convergence == true
s2 = StopAfterIteration(10)             # indicates_convergence == false
cap = StopAfterIteration(20)
sc = (s1 | s2) & cap                    # StopWhenAll(StopWhenAny(s1, s2), cap)

# every leaf active: the group has genuinely stopped, and s1 is a convergence certificate
s1.at_iteration = 1
s2.at_iteration = 11
cap.at_iteration = 21
has_converged(sc)   # false -- expected true
#  active leaves = [s1, s2, cap] -> 3, length(sc.criteria) -> 2, gate fails

# now deactivate s2, which is the criterion that cannot certify anything
s2.at_iteration = -1
has_converged(sc)   # true
#  active leaves = [s1, cap] -> 2 == 2, gate passes
```

---

Disclaimer: found with assistance of AI

kellertuer:
This PR resolves #610. Since it is a larger refactor, I wanted to open this PR a bit earlier so it is visible and people are aware, the code base is changing quite a bit.

# Goal of this PR / Refactor
* objectives will loose their `EvaluationType` - it was not reflected in any of the fields per se and a bit clumsy anyways (and maybe just a very early wrong design decision back then)
* functions inside the objective will “earn” this instead. They will be the former in-place by default
* a wrapper for the former allocating case allows to more flexibly have allocating functions in the objective
* the former `plan/` folder is split into
  * a `base/` folder defining and describing interfaces
  * a `commons/` folder collecting implementations several solvers have in common, like stopping criteria
  * some parts from the `plan/` also move to their respective solvers, if there is no incentive to have these in the `commons` area

While parameters of several structs change, especially all objectives “loose” their first parameter, the hope is that in the end all constructors stay the same and hence this PR is non-breaking.

# Roadmap

* [x] refactor all objectives by removing the evaluation type, refactor all corresponding accessor functions, temporarily remove all `evaluation = ` keywords thereon
* [x] refactor all functions from the `plan/` to `base/` and `commons/` – as a structured way to work through all points from the previous item
* [x] Refactor the docs into a developer guide and a commons area (instead of the plan) once Manopt complies again
* [x] reintroduce the `evalution = ` to the objectives such that they wrap their function components accordingly. This would mean “wrapping” all functions to (internally) be allocating.
* [x] (maybe) additionally find a way to have a more fine granular interface to set single functions to be allocating. Maybe also just the new `AllocatingManifoldFunction` server well enough so wrap single allocating functions
* [x] carefully check that all tests work again, hopefully non-breaking.
* [x] finish the developer docs, especially a few areas, where design ideas should be mentioned for sections in this new area of the documentation.
* [x] check that the new “commons/” area is readable, since it does collect sometimes slightly larger sets of commonly provided functions or types, like step sizes or stopping criteria.

# TLDR
The information “in-place or allocating” gets “more modular” within the objective, code complexity is reduced – and we get developer docs as a side aspect.


kellertuer:
The code part of this rework is finished – maybe up to renaming a few test files and maybe reorganising a few code files as well. But this could already be reviewed and criticised.

The docs rework is still missing (it only pushes them since I locally turned errors to warnings to look at renderings). This might take a bit still, since in the new base/ (Developer tools) a few things should be explained like we do in ManifoldsBase.

kellertuer:
The docs now render fine again, that is without warning/errors about missing elements, same for the new CodeBlocks plugin and its warnings.

After all this refactoring, it is probably still necessary to go through the rendered pages for once and check that the resulting new organisation is find and there is not too many “loose ends”.

A few of the new sections also still need short introductions, especially in the new “Developer Guide” section with respect to design ideas.

kellertuer:
@mateuszbaran if you could take a short look at whether the new refactor fits what you had in mind to only store the inplace variants – that would be great.

I hope the new `InplaceManifoldFunction` (cf. https://manoptjl.org/previews/PR630/commons/functions/#Manopt.InplaceManifoldFunction) wrapper is also fine.

src/solvers/LevenbergMarquardt.jl
```diff
@@ -299,9 +495,9 @@ function step_solver!(
     M = get_manifold(dmp)
     nlso = get_objective(dmp, true)
     FpSq = get_cost(M, nlso, lms.p)
-    set_parameter!(lms.sub_problem, Val(:Objective), Val(:Penalty), lms.damping_term * FpSq)
+    set_parameter!(lms.sub_problem, :Objective, :Penalty, lms.damping_term * FpSq)
     # update base point of the tangent space the subproblem works on
-    set_parameter!(lms.sub_problem, Val(:Manifold), Val(:Basepoint), lms.p)
+    set_parameter!(lms.sub_problem, :Manifold, :Basepoint, lms.p)
```
mateuszbaran: `Val` is preferable in most cases because it should ensure static dispatch.

src/commons/functions.jl
```diff
@@ -0,0 +1,516 @@
...
+        return new{F}(f, result, point_index)
+    end
+    # do not “wrap twice”
+    function InplaceManifoldFunction(imf::InplaceManifoldFunction, ::Symbol = :Point; point_index::Int = 1)
+        return imf
+    end
+end
+function (imf::InplaceManifoldFunction)(M, v, args...)
+    (imf.result === :Point) && return copyto!(M, v, imf.f(M, args...))
+    (imf.result === :Points) && return copyto!.(Ref(M), v, imf.f(M, args...))
```
mateuszbaran:
```suggestion
    (imf.result === :Points) && return map((pa, pb) -> copyto!(M, pa, pb), v, imf.f(M, args...))
```
Let's make compiler's life a bit easier here. `TangentVectors` should be handled similarly.

src/commons/functions.jl
```diff
@@ -0,0 +1,516 @@
...
+    (imf.result === :TangentVector) && return copyto!(M, v, args[imf.point_index], imf.f(M, args...))
+    (imf.result === :TangentVectors) && return copyto!.(Ref(M), v, Ref(args[imf.point_index]), imf.f(M, args...))
```
mateuszbaran: Point should actually be optional in tangent vector `copyto!`, I think we should even completely remove the variant with point.

src/commons/functions.jl
```diff
@@ -0,0 +1,516 @@
...
+  * default that works for most functions: 1
+  * for the proximal map `(M, λ, p)` this is index 2
+
+  # Constructor
+
+    InplaceManifoldFunction(f, result = :Point)
+"""
+struct InplaceManifoldFunction{F} <: AbstractDecoratedManifoldFunction{F}
+    f::F
+    result::Symbol
```
mateuszbaran: I'd suggest making `result` a type parameter of `InplaceManifoldFunction`, and probably also `MutableManifoldFunction` for better performance.

mateuszbaran:
It looks roughly fine, I just have a few minor suggestions that shouldn't impact the overall design.

src/commons/functions.jl
kellertuer: But that would be a large breaking change over in ManifoldsBase?
Also not that for some manifolds it is probably not optional? FixedRank probably?

src/commons/functions.jl
kellertuer: Kind of like an enum?

kellertuer:
At least making the result a type is quite a change still, since I would have to go through all places where the functions are constructed. But yeah, again, I am not good at efficient code, there I need help. Afterwards refactoring is then an effect of that where I am then not even good at efficient coding.

src/commons/functions.jl
mateuszbaran: Allowing both is not breaking and IMO disallowing the variant with the point would be a reasonable low-priority thing to do along the next breaking release. And `FixedRankMatrices` `copyto!` even works without the manifold.

src/commons/functions.jl
mateuszbaran: Like changing `struct InplaceManifoldFunction{F}` to `struct InplaceManifoldFunction{result,F}` -- a bit like real/complex number thing for some manifolds.

src/commons/functions.jl
kellertuer: Ah maybe I thought too math-u there, because the definition of the tangent vector depends a lot on the base point for FR. I think such a dependency was also the reason we have the point-dependency.

mateuszbaran:
The constructor can stay as is I think, constant propagation should work there, it's just the internal representation that needs to be updated.

src/commons/functions.jl
kellertuer: but then I also have to change `result` either to a `Val` of symbol or even some extra new type with subtypes to distinguish? Sorry, I do not understand what is meant here (i.e. what actually needs to be done) nor what the benefit will be

src/commons/functions.jl
mateuszbaran: AFAIK there are exactly zero cases where `copyto!` actually depends on the point. IIRC we only put the point there just in case but that case never materialized.

src/commons/functions.jl
mateuszbaran: No, you can store a Symbol there just normally. I think I can handle that change myself later today.

kellertuer: OK, thanks for the help :)

src/commons/functions.jl
kellertuer: I do not have good overview, but maybe something to have in mind, sure.

The original argument why the point is there is that mathematically we always had both M and p present when talking about X, kind of like an implicit tangent space thing.

src/commons/functions.jl
mateuszbaran: Done :).

src/commons/functions.jl
mateuszbaran: I've also done this myself.

src/commons/functions.jl
mateuszbaran: Usually we do have the point but sometimes it's really inconvenient to access `p` from the place where we need to make a copy, and it's useless effort in the end.

src/commons/functions.jl
kellertuer: Oh I somehow even missed this one. Sure the map is better.

src/solvers/LevenbergMarquardt.jl
kellertuer: Hm ,why do I not get a “accept this change” here? 

Yeah every now and then I struggle a bit when to change from symbol to Val and when to maybe first pass down. But maybe passing down can also just happen for both always anyways.

src/commons/functions.jl
kellertuer: Sure; done carefully, we can surely allow a copy version without the p.
Carefully in the sense of not breaking too much.

kellertuer: ...it is still more like an issue for ManifoldsBase.

src/solvers/LevenbergMarquardt.jl
mateuszbaran: Maybe because I haven't proposed anything in a Github suggestion :wink: . I just think the old variant was fine.

kellertuer: Ah I see, I removed those and you want them back (had to switch to the changes tab to see this); yeah I think the main problem is really the “dispatch patch path”
and that is a bit complicated to handle both “pass down symbols” and “pass down Vals” in parallel. So maybe I changed that because it was the easy fix compared to rewriting the whole get/set parameter tree.

For now – the change to Val will happen one step later. It passed to “Problem+Symbol (+Rest)” in the first two, that matches to “Problem+Val (+Rest)” then dispatch on Manifold – there the same with Basepoint will happen.

It is always step-by.-step, so in the new version “entering with 2 Vals directly” might be a problem? Not sure, have to carefully look at that again probably. I had mainly in mind the approach form above.

Sure that can be refactored to “pass down” (here to the manifold) whatever is after the second parameter – but I am not sure that is implemented that flexibly for now.

kellertuer: ...if I changed the code, the old code errored after some refactor.

For now I am happy, this refactor turned out to only have modified finitely many lines (it was close!)

mateuszbaran: Val needs to be directly in the caller or it's very likely worse than useless. You can't switch to it in the middle.

mateuszbaran: I mean, you technically can switch to it but it defeats the whole purpose.

kellertuer: hm. Then the whole “call tree” of `set_parammeter` and `get_parameter` is both bullshit and inefficient. Was before and still is.

Because the scheme described above is what is currently done. If you come along with a symbol it is wrapped in a Val to dispatch (always only the second argument) and that dispatch passes down or sets a value.

Is that then wrong?

But anyways. I would probably need a few hours somewhere to review this again. There must have been a reason I changed this. But this whole PR was so many hours of work, that I do not remember why I changed every single line.

mateuszbaran: As long as `Val`s can be passed and are not un-wrapped and re-wrapped, it's good enough for me.

kellertuer: No, that is not done even at the current situation of the code.
It can be that I removed it, because
(a) only symbols are passed down, Vals error – then we need a new dispatch for “unmatched Vals to be passed down”
(b) it looked nicer without the Vals in like readable code – and sure that is a small tradeoff, if was are more efficient we should switch back.

But then I see what to do and can check that in the evening today.

kellertuer: Short check: Yes, test fail when changing this. So it might require a substantial refactor somewhere in the parameter-function definitions. And that might stem from exactly the scheme I wrote about above.

On the positive side: Good to know what is efficient here (though I hope to someday maybe be able to do that myself)
On the negative side: Endless PR stays endless -.-

kellertuer: It seems passing Vals – the parameters are not actually set. So it is a larger problem it seems.

kellertuer: No worries, it is just about 250 or so occurrences of `set_parameter!` to check 😄
(joke aside, I am a bit sad that this seems infinitely buggy all here)

But from a short look I noticed that until now
* **all** other solvers do these updates without `Val`
* **all** callbacks do this without `Val` – ah there it might be fine since we use the symbol as key in the dict.

So yes there is a bug here to fix, but should the other subslver updates also switch to using Val directly?

kellertuer: ...ah I see one disadvantage in “wrap early” (having Vals early) – ambiguities on Vals.
But probably “keep Symbols” has similar problems in fighting stuff.

kellertuer: Long live regexp! I think I got it. Just that not only 2026 in general seems the year-of-nothing-works – now I can no longer push because my public key is not accepted? GitHub is doing strange things! But I close this and push when GitHub lets me push again.
I would be ready for 2027 by now, by the way.

kellertuer: Here is the commit that unifies `set_parameter!` to “use Vals earlier – and internally always” https://github.com/JuliaManifolds/Manopt.jl/pull/630/commits/c375007479d1420dca50f1cc35edf2088eca241b

kellertuer:
...and now codecov is drunk. Here it reports 3 lines are missing, looking at the PR on their page, there is nothing to even report on – looking on the branch instead, there is 11 lines, none of which are in the file reported here.

So I have no clue what to fix. Time to go home, after I already spent 20 minutes before I could **push** to GitHub again (yeah it just suddenly broke – and somewhen suddenly worked again).

_edit:_ found the culprit, and fixed one passdown again. Now the parameters should be unified to first passing to Val, then down into the Manopt-depth (e.g. problem->objective->cost) – and internally all places use Val now – symbols are still allowed and wrapped early (they are also never unwrapped – never were before either)

src/commons/functions.jl
kellertuer: I opened https://github.com/JuliaManifolds/ManifoldsBase.jl/issues/285

src/solvers/LevenbergMarquardt.jl
mateuszbaran: Thanks for taking care of this, I think `Val` is generally a better choice for solvers. Regarding callbacks, they have a different design and I think using just Symbols is fine there.

Hopefully next year things will work better. Or everyone will just use AI to generate poor code they think they need instead of using existing libraries, that's sadly also a possibility.

kellertuer: Yes I noticed that for callbacks then as well that there the symbol is a key :)

When I get replaced by AI, that is fine with me as long as I still get a certain salary at least.

And AI coded geometry, those I see already here and there. And utterly poorly written, sure.

mateuszbaran:
I think the new design is good now. Let me know if you want me to check something specific, there are a lot of changes so it's hard to look at everything.

There still seem to be a few remaining TODOs in docs so the PR doesn't look entirely finished.

kellertuer:
Nice!

For now I am fine. Yeah the TODO blocks are areas where we now have new Developer Guide sections and some of the aspects of Manopt.jl and its design ideas still need to be explained.

Concerning what to read – what do you mean, you do not want to read _all_ 17k added lines ;) ?

I think The Inplace function and the mutating function wrappers are the main new things. The rest is both reorganising the code and adding it everywhere.

kellertuer:
I went carefully through the docs and read a lot – as well as asking an AI to give some hints to unify presentation of the docs.

Besides the 6 small snippets on the new tiles on the start page, that I want to rephrase a bit still, this PR should be finished by now. 🥳

kellertuer:
Uff. What a PR. But I think I have now verified all new parts once more and finalised the 6 new tiles on the start page. I would probably merge and register tomorrow then.

docs/src/index.md
```diff
@@ -1,4 +1,56 @@
...
+        wrap: true
+    title: Generic Implementations
+    details: Algorithms are implemented on a general Riemannian manifold using the interface from ManifoldsBase.jl.
+    link: https://juliamanifolds.github.io/ManifoldsBase.jl/stable/
+  - icon: 🧩
+    title: Composable
+    details: All components are designed in a modular fashion, for example problems, objectives, and solver states, but also stopping criteria and step sizes. They can be reused and combined easily.
+  - icon: ⚡️
+    title: Efficient
+    details: All methods use in-place evaluations for both the objective and the manifold. There is also an advanced, but easy to use caching system further promoting efficiency.
```
mateuszbaran: I think you could also mention here compatibility with GPU and https://github.com/JuliaManifolds/ManifoldsGPU.jl which can give massive speedup for some operations.

docs/src/index.md
```diff
@@ -1,4 +1,56 @@
...
+        light: /logo-manifoldsbase.png
+        dark: /logo-manifoldsbase-dark.png
+        alt: ManifoldsBase.jl
+        wrap: true
+    title: Generic Implementations
+    details: Algorithms are implemented on a general Riemannian manifold using the interface from ManifoldsBase.jl.
+    link: https://juliamanifolds.github.io/ManifoldsBase.jl/stable/
+  - icon: 🧩
+    title: Composable
+    details: All components are designed in a modular fashion, for example problems, objectives, and solver states, but also stopping criteria and step sizes. They can be reused and combined easily.
```
mateuszbaran:
```suggestion
    details: All components are designed in a modular fashion, for example problems, objectives, and solver states, but also stopping criteria and step sizes. They can be reused and combined easily. Points and tangent vectors can be represented using arbitrary Julia types.
```

docs/src/index.md
```diff
@@ -1,4 +1,56 @@
...
+  - icon: 🛠️
+    title: Customizable
+    details: All solvers bring recording, debug and callback capabilities, the objectives can keep track of internal function calls.
+  - icon:
+        light: /logo-manifolds.png
+        dark: /logo-manifolds-dark.png
+        alt: Manifolds.jl
+        wrap: true
+    title: Use with Manifolds.jl
+    details: If you have a certain manifold in mind, checkout whether it is already provided in Manifolds.jl or LieGroups.jl
```
mateuszbaran:
```suggestion
    details: If you have a certain manifold in mind, check out whether it is already provided in Manifolds.jl or LieGroups.jl.
```

mateuszbaran:
I've added a few minor suggestions for the landing page but overall I like it :slightly_smiling_face: .

docs/src/index.md
kellertuer: Sure, I could even turn the tile into a link to that. Will try :)
Though that support is a bit preliminary.

kellertuer: ...but I added that as well then.

mateuszbaran (approved):

PR #635, opened 2026-08-22
Always evaluate all stopping criteria for StopWhenAll/StopWhenAny

### LINKED ISSUE #632: `StopWhenArr`/`StopWhenAy` skips side-effecting children
`all` and `any` are short-circuiting, such that if subcriteria are stateful, they can be gated by earlier predicates returning true or false.

For example, `sc = StopAfterIteration(100) & StopWhenChangeLess(M, 1e-9)` would have the first condition fire for iterations `1:99`, never updating the stored state in the second condition, and then for iteration 100 the second condition would compare the current state with the first state, instead of the 99th.

Note that this is true for both `any`, `all` and `_fast_any`.

---

Disclaimer, was surfaced during AI-assisted session

kellertuer:
This fixes the side effects encountered in #632.

One could maybe make the former fast evaluation optional – but for now we do not really have a (side effect free) expensive stopping criterion that would make that worth doing. I would maybe introduce that (again) when it becomes necessary / requested.
One could then store a binary mask which ones have to be updated – do them first and on the other ones stop early (if the first group already provides the binary result for example – of when the first “kicks”)

But for now – the solution here aims for simplicity.

Transparency remark: Parts of this PR were proposed by an AI – but still assessed and even mainly written by myself.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `StopWhenAll` and `StopWhenAny` evaluated their criteria with short-circuiting, so criteria later in the list were not called in every iteration. Stateful criteria like
   `StopWhenChangeLess` or `StopWhenRepeated` were hence not updated and the reset at
-  initialization did not reach them either. Not `requires_update` is used to determine, which ones need to be called even after one could use a short cut already – only the ones without state are “short-circuited”. (#632,#635)
+  initialization did not reach them either. Now `requires_update` is used to determine which ones need to be called even after one could use a shortcut already – only the ones without state are “short-circuited”. (#632,#635)
 
 ## [0.6.4] August 21, 2026
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,12 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.6.5] August 22, 2026
 
+### Added
+
+* a new function `has_state` that per `<:StoppingCriterion` type can indicate, whether a state is present that needs updating in every iteration.
+* a new shortcut for the `StopWhenCriterionWithIterationCondition` that only evaluates every `n`th iteration using `stopping_criterion ≞ n` (`\measeq` <TAB> on REPL) as constructor. While in theory also `%` would have been possible, there was a discussion not to use mnore-common symbols here, cf. (#509) – and the `m` could be seen as “modulo”.
+
 ### Fixed
 
-* `StopWhenAll` and `StopWhenAny` evaluated their criteria with short-circuiting, so criteria
-  later in the list were not called in every iteration. Stateful criteria like
+* `StopWhenAll` and `StopWhenAny` evaluated their criteria with short-circuiting, so criteria later in the list were not called in every iteration. Stateful criteria like
   `StopWhenChangeLess` or `StopWhenRepeated` were hence not updated and the reset at
-  initialization did not reach them either. All criteria are now evaluated in every call. (#632,#635)
+  initialization did not reach them either. Not `has_state` is used to determine, which ones need to be called even after one could use a short cut already – only the ones without state are “short-circuited”. (#632,#635)
 
 ## [0.6.4] August 21, 2026
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,15 @@ The file was started with Version `0.4`.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.5] August 22, 2026
+
+### Fixed
+
+* `StopWhenAll` and `StopWhenAny` evaluated their criteria with short-circuiting, so criteria
+  later in the list were not called in every iteration. Stateful criteria like
+  `StopWhenChangeLess` or `StopWhenRepeated` were hence not updated and the reset at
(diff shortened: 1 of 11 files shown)


kellertuer:
Mateusz, I asked a short check from you, because I think you wrote the `_fast_any` back then; maybe you had a reason for the short-cut evaluations?

mateuszbaran:
I think it's fine to remove `_fast_any`. I'd expect Julia to optimize the new code well.

mateuszbaran:
Though I personally like the short-circuiting semantics so maybe it would make sense to have that as a flag in StopWhenAny/StopWhenAll?

kellertuer:
I was thinking about a flag, but if I would probably directly go for a binary indicator per element in the group as I wrote.

The danger with a flag is that a user activating that might not see the side effects – as Lukas pointed out (and the tests now also check) for example for `StopWhenChangeLess`. So such a flag would need a bit of docs and a note to a user to be careful.

And if I reintroduce it, with or without the old `_fast_any`? (I always forget why we had that in the first place)

mateuszbaran:
I see, the flag should then probably be on the criterion level (that is a new function `is_eligible_for_short_circuiting(::Type{StopWhenChangeLess}) = false` etc.). Then we have a slightly elaborate logic of skipping some subcriteria.

`_fast_any` might not be necessary anymore, I introduced it in the CMA-ES PR because Julia was quite slow at dealing with its elaborate stopping criteria.

kellertuer:
Yeah, on the SC level sounds good, but that would be quite some work to go through all SC and check and then thunk of a good logic in all/any how to handle them and when one can shortcut. I would have to think about that a long long time I fear. It sounds tricky to me to get that right.

mateuszbaran:
I'd just ask AI to go over all stopping criteria and define it based on whether a criterion uses values from previous iterations or not. It doesn't sound complicated.

kellertuer:
That part not, sure.

Handling that in maybe-shortcut-if-possible is something I currently can not wrap my had around easily.

kellertuer:
Oh and when we have such a new function (name still needs improvement, maybe just `requires_call` or so) we could also use it in a SC I thought we already had, namely something similar to `DebugEvery` – a criterion only checked on certain intervals.

But sure, we can do that – just the logic within the now new function becomes a bit more involved, checking when which SC still needs to be called. Have to think about this for a while probably.

mateuszbaran:
I've added the eligibility mechanism; AI proposed a nice method for handling the logic but did a few errors with marking specific criteria as eligible (hopefully I corrected them all).

kellertuer:
Ah, that looks reasonable, I have to look at it tomorrow with a fresh coffee, but at first glance just the function name could be a bit improved hopefully, maybe just `has_state(sc)` – in AlgorithmsInterface an SC is even split int SC and SC_state – to indicate it has an internal state (that requires updating).

But nice idea. An we could use it for a meta stopping criterion that only evals every Nth iteration as well
(until now we only have a repeated one and ones that “start late” or “end early” I think)

kellertuer:
I changed the name to `has_state` and noticed that my idea of the modulo one Is nearly included already, one can use the iteration based condition ones – and since we use funny utf8 for the shortcuts there, I introduced yet another funny pne.

One thing that could still be done is that in principle, `has_state(StopWhenAny)` _could_ return false, when all its inner criteria are false allowing to “shortcut” a whole “branch” in a nested setting.

Just on a type and not an instance I was too lazy to find out how to best to that (iterate through the types of a tuple that is).

mateuszbaran:
`has_state` is a bit misleading; technically if a criterion stores some values for printing stopping reason later, it's also state but it doesn't matter for short-circuiting. Maybe `has_logical_state` to emphasize what kind of state is meant there?

> One thing that could still be done is that in principle, `has_state(StopWhenAny)` _could_ return false, when all its inner criteria are false allowing to “shortcut” a whole “branch” in a nested setting.

That's right, I can add that logic there.

kellertuer:
Uff. I was happy I found a name I was ok with 😄 
Your function name was a bit too narrow, mine is now maybe too broad.
Hm, `has_logical_state` does not do it for me really, the thresholds or at_iteration could also be logical.

It is more like it has a “state that needs to be kept up to date” (as compared to the “plain” state of parameters/at_iteration things). I simply considered in my reasoning the parameters/at_iteration more like the parameters of the SC and the rest the actual state (in like dynamic part, similar to the solver state being dynamic).

kellertuer:
What about a name like `requires_update`?

In general I like the current boolean form (negative of yours) a bit better, because it is the positive one (there has to be done something) not the negative one (nothing to be done can be skipped) – but then I feel “requires update” is the closest we can get.

mateuszbaran:
I like `requires_update` better than `has_state`. Even though it's fairly broad, it's also difficult to accidentally misunderstand.

kellertuer:
Then let's take that, it will stay an internal function anyways so we can rename it when we get a better idea. Can you do the renaming (it is now just a search and replace) when you do the other ideas as well? 

Besides that, we should be finished with this I think.

mateuszbaran:
Sure, I will do the renaming in a moment.

mateuszbaran:
Done. I've also done a few other minor updates like fixing type stability and marking `StopAfter` as not requiring updates (my mistake from yesterday).

kellertuer:
Thanks your changes look good to me. Do you want to approve the PR, then I can merge it later (already also working on next weeks lecture as well here).

mateuszbaran:
Thanks, I will do one last pass and then approve.

kellertuer:
Perfect. And thanks for the great collaboration on this one again. I think together we found a really good solution in both code and naming.

mateuszbaran (approved):

PR #638, opened 2026-08-25
Fix warning and use cache in CI

JoshuaLampert:
Looking, e.g., at https://github.com/JuliaManifolds/Manopt.jl/actions/runs/32821271727/job/97719720572?pr=637, there is a warning and a notice from GitHub Actions. The warning is about the argument `file` in codecov/codecov-action, which has been renamed to `files` and the notice is about a missing julia-actions/cache step (this should speed up CI because it can reuse the Pkg cache from previous CI runs).
Does this also need a changelog entry? It's not really relevant for a user.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,13 +18,14 @@ jobs:
       - uses: julia-actions/setup-julia@v3
         with:
           version: ${{ matrix.julia-version }}
+      - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
       - uses: julia-actions/julia-processcoverage@v1
       - uses: codecov/codecov-action@v7
         with:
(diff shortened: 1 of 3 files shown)


JoshuaLampert:
I also changed julia-actions/setup-julia@latest to julia-actions/setup-julia@v3 to hopefully fix the warning, e.g., in https://github.com/JuliaManifolds/Manopt.jl/actions/runs/32821271782/job/97719720569?pr=637. It complains about Node.js 20 being used, but v3 of julia-actions/setup-julia should use Node.js 24. So I assume latest points to an older version. Using julia-actions/setup-julia@v3 is also consistent with what CI uses.

kellertuer:
Oh, thanks for even fixing such things! No, that does not need a Changelog entry – the dependabot also does not do them. We can just merge it.

The changelog-CI is so we always remember to at least consider whether we need it :)

PR #639, opened 2026-08-25
Fix an inconsistency with `AbstractConstrainedFunction`

kellertuer:
Thanks @JoshuaLampert for noticing. Now it should be nicely unified.

Note that for the one function that actually uses `μ`, `λ`, _and_ `ρ` still need a special setter, since in that case the type of the duals is different from the penalty (which is just a number).
Besides that, the abstract type now does bear both `CO` and `T` as a parameter (instead of a mix of both as before). That way, the types can still be nicely used and checked in the `set_parameter!` function.

I will merge this (and register a new version) after lunch then.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,8 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `augmented_Lagrangian_method` set the penalty parameter `ρ` of the sub problem's cost to the
   constant `1/3` instead of the current `alms.ρ`, while its gradient did receive `alms.ρ`. (#637)
-* fix `set_parameter!` for `:μ` and `:λ` for  `AugmentedLagrangianCost`. (#637)
-* fix `AbstractConstrainedFunction` to have two parameters, one for the constrained objective it wraps and one for the type of dual parameters.
+* fix `set_parameter!` for `:μ` and `:λ` for `AugmentedLagrangianCost`. (#637)
+* fix `AbstractConstrainedFunction` to have two parameters, one for the constrained objective it wraps and one for the type of dual parameters (#639).
 
 ## [0.6.5] August 22, 2026
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,9 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `augmented_Lagrangian_method` set the penalty parameter `ρ` of the sub problem's cost to the
   constant `1/3` instead of the current `alms.ρ`, while its gradient did receive `alms.ρ`. (#637)
-* `set_parameter!` for `:μ` and `:λ` tied the dual variable's type to the type parameter of
-  `AbstractConstrainedFunction`. It hence never applied to `AugmentedLagrangianCost` and, falling
-  back to a no-op, silently left its `μ` and `λ` at their initial values. (#637)
+* fix `set_parameter!` for `:μ` and `:λ` for  `AugmentedLagrangianCost`. (#637)
+* fix `AbstractConstrainedFunction` to have two parameters, one for the constrained objective it wraps and one for the type of dual parameters.
 
 ## [0.6.5] August 22, 2026
 
--- a/src/base/function/constrained.jl
+++ b/src/base/function/constrained.jl
@@ -14,7 +14,7 @@ function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:μ}, μ:
     return acf
 end
 get_parameter(acf::AbstractConstrainedFunction, ::Val{:μ}) = acf.μ
-function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:λ}, λ) where {CO, T}
+function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:λ}, λ::T) where {CO, T}
     acf.λ = λ
     return acf
 end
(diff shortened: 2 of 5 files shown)


src/base/function/constrained.jl
```diff
@@ -1,23 +1,34 @@
...
+abstract type AbstractConstrainedFunction{CO, T} <: AbstractManifoldFunction end
 
-function set_parameter!(acf::AbstractConstrainedFunction, ::Val{:μ}, μ)
+function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:μ}, μ::T) where {CO, T}
     acf.μ = μ
     return acf
 end
 get_parameter(acf::AbstractConstrainedFunction, ::Val{:μ}) = acf.μ
-function set_parameter!(acf::AbstractConstrainedFunction, ::Val{:λ}, λ)
+function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:λ}, λ) where {CO, T}
```
JoshuaLampert:
```suggestion
function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:λ}, λ::T) where {CO, T}
```
To be consistent with the `μ` case a few lines above?

src/base/function/constrained.jl
```diff
@@ -1,23 +1,34 @@
...
     acf.λ = λ
     return acf
 end
 get_parameter(acf::AbstractConstrainedFunction, ::Val{:λ}) = acf.λ
+function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:ρ}, ρ::T) where {CO, T}
+    acf.ρ = ρ
+    return acf
+end
+get_parameter(acf::AbstractConstrainedFunction, ::Val{:ρ}) = acf.ρ
+function set_parameter!(epc::AbstractConstrainedFunction{CO, T}, ::Val{:u}, u) where {CO, T}
```
JoshuaLampert:
```suggestion
function set_parameter!(epc::AbstractConstrainedFunction{CO, T}, ::Val{:u}, u::T) where {CO, T}
```
Same here?

Changelog.md
```diff
@@ -12,9 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * `augmented_Lagrangian_method` set the penalty parameter `ρ` of the sub problem's cost to the
   constant `1/3` instead of the current `alms.ρ`, while its gradient did receive `alms.ρ`. (#637)
-* `set_parameter!` for `:μ` and `:λ` tied the dual variable's type to the type parameter of
-  `AbstractConstrainedFunction`. It hence never applied to `AugmentedLagrangianCost` and, falling
-  back to a no-op, silently left its `μ` and `λ` at their initial values. (#637)
+* fix `set_parameter!` for `:μ` and `:λ` for  `AugmentedLagrangianCost`. (#637)
```
JoshuaLampert:
```suggestion
* fix `set_parameter!` for `:μ` and `:λ` for `AugmentedLagrangianCost`. (#637)
```

Changelog.md
```diff
@@ -12,9 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+* fix `AbstractConstrainedFunction` to have two parameters, one for the constrained objective it wraps and one for the type of dual parameters.
```
JoshuaLampert:
```suggestion
* fix `AbstractConstrainedFunction` to have two parameters, one for the constrained objective it wraps and one for the type of dual parameters (#639).
```

src/base/function/constrained.jl
```diff
@@ -1,23 +1,34 @@
...
+    epc.u = u
+    return epc
```
JoshuaLampert:
```suggestion
function set_parameter!(acf::AbstractConstrainedFunction{CO, T}, ::Val{:u}, u) where {CO, T}
    acf.u = u
    return acf
```
Probably the variable name should now be `acf` instead of `epc`?

src/commons/sub_functions.jl
```diff
@@ -325,30 +317,22 @@ This struct is also a functor in both formats
...
+    ρ::T
+    u::T
 end
 function ExactPenaltyGrad(
-        co::ConstrainedManifoldObjective, ρ::R, u::R; smoothing = LinearQuadraticHuber()
-    ) where {R}
-    return ExactPenaltyGrad{typeof(smoothing), typeof(co), R}(co, ρ, u)
+        co::ConstrainedManifoldObjective, ρ::R, u::R; smoothing::S = LinearQuadraticHuber()
+    ) where {R, S <: SmoothingTechnique}
+    return ExactPenaltyGrad{S, typeof(co), R}(co, ρ, u)
```
JoshuaLampert:
```suggestion
        co::ConstrainedManifoldObjective, ρ::T, u::T; smoothing::S = LinearQuadraticHuber()
    ) where {T, S <: SmoothingTechnique}
    return ExactPenaltyGrad{S, typeof(co), T}(co, ρ, u)
```
Should these rather be `T`s to avoid confusion since `T` is used in the definition of the struct?

JoshuaLampert (commented):
Thanks! I left a few small suggestions below.

src/base/function/constrained.jl
kellertuer: Nirgs. I had hoped it was finally consistent. Maybe under the current circumstances (a bit much too do in general) I should stop trying to code consistently :(

src/base/function/constrained.jl
kellertuer: sure, Fixed on a separate commit, since GitHub got confused here with 2 changes on one line ;)

src/base/function/constrained.jl
```diff
@@ -1,23 +1,34 @@
...
-    AbstractConstrainedFunction{T} <: AbstractManifoldFunction
+    AbstractConstrainedFunction{CO, T} <: AbstractManifoldFunction
 
-A common supertype for functors that model constraint functions.
+A common supertype for functions that model functions of the subproblem for a
+constrained objective, which they therefore wrap.
 
 This supertype provides access to the fields `λ` and `μ`, the dual variables of
-constraints of type `T`.
+constraints of type `T` while the objectives type `CO` is also a parameter.
```
JoshuaLampert: One other thing I noticed: Is this sentence still true? Now `ExactPenaltyCost` and `ExactPenaltyGrad` are subtypes of `AbstractConstrainedFunction`, but do not have fields `λ` and `μ` (but `ρ` and `u` instead).

kellertuer: Sure. I can also rephrase that...

JoshuaLampert (approved):

PR #640, opened 2026-08-25
Introduce Benchmarks

### LINKED ISSUE #634: (A small section of) Benchmarks
I think mainly to avoid regression of performance in the future and maybe to compare a little bit to other approaches some kind of benchmark – that maybe can be activated on CIs as well, would be nice.

In a `benchmark/` folder using the same workspace
in some way trying to be reproducible or reporting a certain analysis of a few default solvers on small examples (so it does not take ages on CI).

I am not yet sure which packages to best use therein to (a) accomplish such a benchmark of (b) others to compare to.

kellertuer:
This is a first start for benchmarks to close #634.


kellertuer:
Transparency Note: I collected the ideas for this benchmark structure from several packages that an AI proposed to check and said AI also proposed an initial module structure here, which I then refined.

kellertuer:
@JoshuaLampert what do you think of this structure?

@mateuszbaran I saw you started something like this in Manifolds already, thought maybe a bit less modular. What do you think about such benchmarks?

In general, I am not yet sure the new action is working, since it does not seem to get triggered and I can not see why not.
Locally I can run the benchmarks, even individual ones (that was my goal, similar to our tests) – but I could not yet run comporisons (I do not understand how that works actually).

JoshuaLampert:
The structure looks good to me, but I didn't go through everything. That the benchmarks are not running in this PR is expected. This is a consequence of `pull_request_target` being [used in the action](https://github.com/JuliaManifolds/Manopt.jl/pull/640/changes#diff-22eef62c9d7b36d02f5fd5aaada92702e22d6795c464239cb7b8fe0f26ea1e1cR3), which is correct here, see https://github.com/MilesCranmer/AirspeedVelocity.jl/issues/97 for more context. The benchmark results should appear as a comment starting with the next PR.

JoshuaLampert:
Oh, but for the reasons explained in the issue I posted above, I would recommend setting `bench-on: ${{ github.event.pull_request.head.sha }}` in the action.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,32 @@
...
+  pull-requests: write    # needed to post comments
+
+jobs:
+  bench:
+    runs-on: ubuntu-latest
+    if: contains( github.event.pull_request.labels.*.name, 'benchmark') || github.ref == 'refs/heads/master' || contains(github.ref, 'refs/tags/')
+    steps:
+      - uses: MilesCranmer/AirspeedVelocity.jl@action-v1
+        with:
+          julia-version: '1'
```
JoshuaLampert:
```suggestion
          julia-version: '1'
          bench-on: ${{ github.event.pull_request.head.sha }}
```
Like this. Otherwise new benchmarks are not picked up in the PR that adds them.

kellertuer:
So your change already fixes what you wrote or do I have to add that somewhere else as well? 

For me – the benchmark code itself slowly makes sense, but CI – even in general – is usually a large jungle of confusion and guesswork, guess work, and long try-and-error.

JoshuaLampert:
> So your change already fixes what you wrote or do I have to add that somewhere else as well?

My change should fix it. But it remains that this PR cannot run the benchmarks yet (for permission reasons I think). To see it in action, we need to wait for the next PR.

kellertuer:
Ah yes, that might be that new actions on a PR are not yet run. 

Then this PR is nearly finished, since it is meant a s start providing the structure and just a first few small benchmarks.

mateuszbaran:
> @mateuszbaran I saw you started something like this in Manifolds already, thought maybe a bit less modular. What do you think about such benchmarks?

I think it's easier to interpret more focused benchmarks (here: a single iteration of a solver, initialization, termination condition evaluation) instead of full runs which combine multiple aspects. The general structure seems fine though.

For full runs there are other methods like https://github.com/JuliaSmoothOptimizers/SolverBenchmark.jl that gather multiple figures of merit.

Regarding the Manifolds.jl benchmarks, they just didn't get enough attention; it's not enough to set it up, someone has to then keep checking them and (ideally) make attempts at improving the time.

kellertuer:
I was not even aware the had benchmarks in Manifolds.jl and only noticed that today;
Maybe because they are nearly never run.
One could maybe increase visibility by having some recent benchmark displays as part of the docs or so.

For the single-step versus solver checks. That sounds reasonable. For now I am happy to first have an initial setup – and for the small problems tackled now as well as solvers like gradient descent a single iteration would really be a bit too short I fear?

But sure, a good next step could be to have slightly larger problems and solvers that only benchmark a single step on these.

mateuszbaran:
> One could maybe increase visibility by having some recent benchmark displays as part of the docs or so.

Maybe; I just have too many more interesting things to do :sweat_smile: .

> For the single-step versus solver checks. That sounds reasonable. For now I am happy to first have an initial setup – and for the small problems tackled now as well as solvers like gradient descent a single iteration would really be a bit too short I fear?

A single iteration of a very low dimensional problem might indeed be at the noise threshold, especially on CI, but even a 100-dimensional problem should be large enough to be consistently measurable. And you save on compute time to run the benchmark by running a single iteration.

kellertuer:
I think for these super small problems here, the whole solver runs are fine.

And sure, single iteration steps are a good next step.

PR #641, opened 2026-08-27
Introduce BarzileiBorwein as its own Stepsize

### LINKED ISSUE #537: Rework Barzilai-Borwein rule to be able to use it as an initial guess for other line searches
It could be better than 1.0 in other cases.

kellertuer:
this closes #537.

The main modelling of the idea to solve this are done – but a few details are still missing:

* [x] introduce `BarzileiBorweinStepsize` as its own separate stepsize (not only in nonmonotone line searches
* [x] isolate its docs
* [x] introduce a `BarzileiBorwein` factory
* [x] move docs of the formulae to the factory
* [x] add `show`/ `status_summary` for the new stepsize
* [x] refactor `NonMonotoneLinesearch` to internally use the new  `BarzileiBorweinStepsize` internally as a field
* [x] make sure the `NonMonotoneLinesearch` also works on the circle again. Currently it seems to accidentally produce a `Float64` point on which one can of course not work in-place in the linesearch
* [x] update `show`/ `status_summary` for the `NonMonotoneLinesearch`
* [x] introduce a new `AbstractInitialLinesearchGuess` – maybe as generic as that it allows to use another line search as initial guess – then BB is just a special case here
* [x] Check docs that all renders fine
* [x] Check test coverage
* [x] Fix that Benchmark is really only running if the tag is present.

### DIFF
--- a/src/commons/stepsizes.jl
+++ b/src/commons/stepsizes.jl
@@ -745,7 +745,8 @@ mutable struct BarzilaiBorweinStepsize{
         if max_stepsize <= min_stepsize
             throw(
                 DomainError(
-                    max_stepsize, "The upper bound for the step size lower bound.",
+                    max_stepsize,
+                    "The upper bound for the Barzilai–Borwein step size has to be larger than the lower bound $(min_stepsize).",
                 ),
             )
         end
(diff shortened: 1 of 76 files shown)


kellertuer:
update after a bit of twiddling: The step sizes (actually all of them) now carefully consider whether upon construction any of the incoming vectors or points to be stored need to be wrapped just to be safe. NM did that very implicitly before, now all stepsizes to this explicitly.

kellertuer:
Transparency remark:

The last maybe 15 commits where based on feedback by an AI on what one can fix and unify in the code base

* iteration number is now always k
* a lot of backticks and typos in docs were found and fixed
* a few places missing/nothing was mixed up but nowhere caused an error until now
* the benchmark CI had a bug and ran even on this PR. For that the fix was proposed and changed by the AI
* for all others I carefully reviewed each and every line before doing the commits manually.

kellertuer:
@mateuszbaran is this like you had the feature in mind?
Sorry it is blended a bit with typos I found. Will continue with that maybe better on a next PR ;)

Main things to look at are the new stepsize https://manoptjl.org/previews/PR641/commons/stepsizes/#Manopt.BarzilaiBorwein-Tuple (previously only used within Nonmonotone line search) and the new (internal function for an) initial guess https://manoptjl.org/previews/PR641/commons/stepsizes/#Manopt.StepsizeInitialGuess so that the BB stepsize can be used as an initial guess here.

src/commons/stepsizes.jl
```diff
@@ -654,6 +672,242 @@ function AdaptiveWNGradient(args...; kwargs...)
...
+    * retraction method:         $(_MANOPT_INDENT)$(bbs.retraction_method)
+    * vector transport method:   $(_MANOPT_INDENT)$(bbs.vector_transport_method)
+    """
+end
+
+"""
+    BarzilaiBorwein(; kwargs...)
+    BarzilaiBorwein(M::AbstractManifold; kwargs...)
+
+Consider the current iterate and gradient ``p_k`` and ``X_k`` as well as the last
```
mateuszbaran, line 854: `s`/`s_k` and `y`/`y_k` are used a bit too interchangably in this docstring.

kellertuer: Ah yes sorry, I might have started with _k and then moved to without ks for y and s. Do you have a preference that to unify them to?
In the struct I used both fields without k for now

mateuszbaran: Maybe it should also mention what is done when there is no last iterate.

mateuszbaran: I usually prefer using `_k` until there is too much of it, and then I prefer not using it 😄 .

kellertuer: Sure will mention the default in the start.

And yes, first with _k until it was too much, then without is how this current state came along. I will check

src/commons/stepsizes.jl
```diff
@@ -654,6 +672,242 @@ function AdaptiveWNGradient(args...; kwargs...)
...
+    end
+    #compute the new Barzilai-Borwein step size
+    s1 = real(inner(M, p, bb.s, bb.y))
+    s2 = real(inner(M, p, bb.y, bb.y))
+    s2 = s2 == 0 ? 1.0 : s2
+    s3 = real(inner(M, p, bb.s, bb.s))
+    #indirect strategy
+    return if bb.strategy == :inverse
+        if s1 > 0
+            stepsize = min(bb.max_stepsize, max(bb.min_stepsize, s1 / s2))
```
mateuszbaran:
```suggestion
            stepsize = clamp(s1 / s2, bb.min_stepsize, bb.max_stepsize)
```
`clamp` has the same logic but I think it's a bit more readable.

src/commons/stepsizes.jl
kellertuer: HM, GitHub decided to only load half the PR for now. I fixed and unified it, the _k is added in all appropriate places, I mentioned the initial case twice (s_k=y_k=0 – stepsize becomes max)

mateuszbaran:
The code extraction looks more or less like what I had in mind, so I think it should be fine.

mateuszbaran (approved):

PR #643, opened 2026-08-29
A Code review

kellertuer:
I got a nice grant by the Julia core developers / community (thanks!)

An AI should never be used to generate code, but I let it analyse the code and produce reproducing code that narrows down errors and bugs in the code or the docs.
This PR works through such a generated list, the changes to the code are an (exhausting for me at least) collaboration with the AI to narrow down what to fix, the changelog and commits are hand crafted by the human me.

__Status:__ 
* [x] The AI claims to have found about 100 issues, most either docs ore sometimes a bit uncareful keyword defaults. All are fixed.
* [x]  A second run on another model currently found 50 further remarks that were not yet checked 
* [x] update the contributions.md and explain which GitHub Tags _mean_ something in review and which ones also _act_ something (benchmark, skip documentation, no changelog necessary)

__Nice new things along the way__:
* a `no changelog necessary` tag to skip that check
* a `skip documentation` tag to skip documentation building
* ...both were created with help of AI and added to the dependabot, so it will no longer have failing CIs in the future
* better benchmark tag handling
* several further typos and imbalanced brackets fixed in the docs.

A final rerun of the analysis will be performed when the token gods are gracious, but I hope a second run will not discover much.


kellertuer:
Sory that this is a monster-PR; I felt like checking all the things reported by the AI in one longer task here was easier than opening several PRs.

@mateuszbaran can you skim this PR whether you feel that is something we want to do on LieGroups and Manifolds as well?

The main things found
* typos (about 50% maybe?)
* outdated docs compared to actual signatures
* not “passed on keywords”
* a few cases where certain combinations of keywords provided yielded an error, because the safeguard on the input was missing exactly such a case
* maybe 5 actual bugs – in IPNewton in 2 places the duals Z and Y (for inequality and eq constraints) were accidentally switched

So a lit of nitty-gritty edges cases basically. Also quite a few I had to revise and improve the proposed fixes, sometimes because they were plain wrong, mostly because they were to complicated or the opposite of what is usually done in Manopt – the complicated thing was the dominant case here.

mateuszbaran:
> @mateuszbaran can you skim this PR whether you feel that is something we want to do on LieGroups and Manifolds as well?

Sure, why not, the only drawback is that someone has to actually review the issues reported by AI but it makes the libraries more polished in the end.

> Sory that this is a monster-PR; I felt like checking all the things reported by the AI in one longer task here was easier than opening several PRs.

https://github.com/JuliaManifolds/Manifolds.jl/pull/917 also turned a bit larger than expected but I wanted to do everything in one go :smile: . It's feature-complete already.

kellertuer:
I could split that per manifold or per function in Manifolds.jl for example. That is why I am asking.

Sure this one is monstrous, but I did not find a good spot to “split” in the end. And the main things (typos and uncareful check of input) is also often very similar in what is changed.

mateuszbaran:
Sure, splitting per manifold sounds good.

BTW, I've recently seen that Runic can format Julia code in .md and .qmd files (see https://github.com/fredrikekre/runic-action , the `extensions` part. Maybe it would be good to enable it in JuliaManifolds?

kellertuer:
Yes that would be good to extend runic to.

kellertuer:
This is finally nearly done. I would like to rerun one of the code analysis's – but ran out of token while trying to recheck some of the claims (most of which are by now made up nonsense).

kellertuer:
I am considering to maybe fire up the code analysis a third time, just out of curiosity; but today it used up the quota for the next 5 hours in 7.5 minutes and I am not sure it got anything done it can continue.
So I might also just leave it at the current state.

I do hope I was careful enough – I rejected several proposals that were plain wrong in code.  As soon as it was “just text” I was maybe not super careful, but I avoided too long and sloppy text changes as well.

One item I might still check a bit is the `HybridCoefficientRule`. There the AI noted that the “final update” CG does on restart at

https://github.com/JuliaManifolds/Manopt.jl/blob/16f773a5ec15e18fc09d08816e324d86c0ae52b2/src/solvers/conjugate_gradient_descent.jl#L204-L210

only updates the storage of the “outer” Hybrid rule, not the inner stored ones. BealeReset solves this with only having 3 own memories for all its subroles, which might also not be best. One could maybe solve this by implementing a function for `update_storage!(cgs.coefficient, amp, cgs)` so one can dispatch on the “rule inside the coefficient storage” again. By default it would call the method used above.
On this point the AI was very unhelpful. It went through
* it is a severe bug
* it is not a bug
* it is a minor bug
* it is not a bug
* it is a severe bug
* it is something minor and we can just do as Beale (no this is more general here...)
* it is a minor anecdote left to the reader

If would resolve anyways after a first step after the restart (and with old==new that might just be a nullstep) – so it is probably not urgent to fix this.

mateuszbaran:
I guess `DirectionUpdateRuleStorage` should not be nested anyway? If there is more to do during a restart then maybe we should have a separate function for that.

kellertuer:
in the Hybrid one it is slightly nested, since every single “inner” one brings its own storage – and the lower bound is the one using the storage from the HybridRule directly. 

So the idea would then be that `update_storage` on a `DirectionUpdateRuleStorage` would just update the .storage – as CG does manually for now. and the `DirectionUpdateRuleStorage{<:Hybrid}` would additionally update all its children.

mateuszbaran:
I see, then it's really quite convoluted.

kellertuer:
I think it is ok and can be fixed in a nice way, I just need a bit of time to carefully check

kellertuer:
Ah a third run did finish. Somehow it mainly found a few more typos – the “severe” errors are by now very very edge edge cases, will still carefully check which ones to take and then probably also tackle the Hybrid one since I have a good idea I hope.

kellertuer:
I think this looks fine now (hope the code cov is back with the last commit)

While I do not want to submit the reports to the repository – they are super-slop, hard and not fun to read I worked through the following, a lot of them where refactored and rewritten to either be correct or match Manopts style a bit more

* [report.md](https://github.com/user-attachments/files/31962958/report.md)
* [report2.md](https://github.com/user-attachments/files/31962961/report2.md)
* [report3.md](https://github.com/user-attachments/files/31962963/report3.md)

I think I had a short try without such a report on the side and that was mainly one chat of pure chaos. So in retrospect let the LLM edit the report and when an item is okay apply it was a workflow that worked – at least to some extend.

I might run a fourth check somewhen but maybe there were enough for this PR.

I will read the changelog tomorrow one last time and then probably merge and register.

_edit:_ Read and sorted the Changelog. Will merge tomorrow morning then.

mateuszbaran:
Oh, this is a very long and detailed report to review. It's good you managed to work through that.

kellertuer:
One? I got that three times ;)
When the first was finished I asked another model and got the second, then asked the first model again and got the third.

Yes it was a bit of work, will maybe better so that on parts of the code base next time.

PR #644, opened 2026-09-03
A bug in adaptive regularization algorithm working on manifolds in complex numbers

amitjamadagni:
Hi, 

We have been integrating different solvers from Manopt.jl to optimize tensor network representations using ITensors.jl (a sample of this integration has been published in one of our recent works [1]). I have further tested these optimization routines with different solvers and in the process noticed that Adaptive Regularization with Cubics results in an error,  a sample to reproduce the error is at [2]. From my understanding this is possibly due to a comparison involving complex numbers (the values are real as expected but the variables in the algorithm get assigned as complex numbers with a negligible complex part). I have tried to fix these occurrences in this PR and haves tested it with the above sample and also on some other examples that we are currently working on (which we hope to put out on arXiv soon).

Thanks for the nice package!

[1] https://doi.org/10.1063/5.0317339
[2] https://gist.github.com/amitjamadagni/bb84ca867df1a19e6f5c846efbab8ad4

### DIFF
--- a/src/solvers/Lanczos.jl
+++ b/src/solvers/Lanczos.jl
@@ -193,8 +193,8 @@ function step_solver!(dmp::AbstractManoptProblem{<:TangentSpace}, ls::LanczosSta
         ls.Hp_residual .= ls.Hp_residual - α * ls.Lanczos_vectors[k]
         # Update tridiagonal matrix
         ls.tridig_matrix[k, k] = α
-        ls.tridig_matrix[k - 1, k] = real(β)
-        ls.tridig_matrix[k, k - 1] = real(β)
+        ls.tridig_matrix[k - 1, k] = β
+        ls.tridig_matrix[k, k - 1] = β
         min_cubic_Newton!(dmp, ls, k)
     end
     copyto!(M, ls.S, p, sum(ls.Lanczos_vectors[k] * ls.coefficients[k] for k in 1:k))
--- a/src/solvers/Lanczos.jl
+++ b/src/solvers/Lanczos.jl
@@ -152,7 +152,7 @@ function step_solver!(dmp::AbstractManoptProblem{<:TangentSpace}, ls::LanczosSta
             copyto!(M, ls.Lanczos_vectors[1], p, ls.X / nX)
         end
         get_objective_hessian!(M, ls.Hp, arcmo, p, ls.Lanczos_vectors[1])
-        α = inner(M, p, ls.Lanczos_vectors[1], ls.Hp)
+        α = real(inner(M, p, ls.Lanczos_vectors[1], ls.Hp))
         # This is also the first coefficient in the tridiagonal matrix
         ls.tridig_matrix[1, 1] = α
         ls.Hp_residual .= ls.Hp - α * ls.Lanczos_vectors[1]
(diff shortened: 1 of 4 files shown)


src/solvers/adaptive_regularization_with_cubics.jl
```diff
@@ -540,8 +540,8 @@ function step_solver!(dmp::AbstractManoptProblem, arcs::AdaptiveRegularizationSt
     ρ_vec = arcs.X + 0.5 * get_hessian(M, mho, arcs.p, arcs.s)
     ρ_den = -inner(M, arcs.p, arcs.s, ρ_vec)
```
kellertuer: I think just an `real` here should be enough? inner is not guaranteed to be real.

kellertuer:
Great that you are using Manopt, I am always happy to read about that!

Thanks for reporting it and directly seeing the fix :) I think I spotted how one `real` is enough. We decided in the interface somewhen that `inner` is not necessarily required to return real, cf. https://juliamanifolds.github.io/ManifoldsBase.jl/stable/functions/#ManifoldsBase.inner-Tuple{AbstractManifold,%20Any,%20Any,%20Any}, but yeah maybe Manopt code is sometimes a bit older that that decision/precision – so we might have missed that in a few spots

I personally do not work often enough on complex manifolds, so that ight also have been a reason.

I also checked whether the neighbouring PR (#643) “saw” this, it indeed did not! So this fix is very welcome. Can you check whether my proposal works fine as well?

kellertuer:
If you like you could add a changelog entry as well – otherwise I can add it when this is on master and merged into the neighbouring one – that one has a longer list of changes already ;)

kellertuer:
If you like, you could add your paper [1] also to https://github.com/JuliaManifolds/Manopt.jl/blob/master/docs/src/about.md#work-using-manoptjl

That is still a bit work in progress, but I think it is nice to collect a few papers that actively use Manopt :)

amitjamadagni:
Thanks! Yes the suggested change as in making the inner real is sufficient. I have checked this with the example as above and it works! 

I have added a line to the Changelog.md and also the paper to the docs. I hope there are in the right place! Thanks once again for the quick response!

src/solvers/Lanczos.jl
```diff
@@ -189,12 +189,12 @@ function step_solver!(dmp::AbstractManoptProblem{<:TangentSpace}, ls::LanczosSta
...
         ls.Hp_residual .= ls.Hp - β * ls.Lanczos_vectors[k - 1]
-        α = inner(M, p, ls.Hp_residual, ls.Lanczos_vectors[k])
+        α = real(inner(M, p, ls.Hp_residual, ls.Lanczos_vectors[k]))
         ls.Hp_residual .= ls.Hp_residual - α * ls.Lanczos_vectors[k]
         # Update tridiagonal matrix
         ls.tridig_matrix[k, k] = α
-        ls.tridig_matrix[k - 1, k] = β
-        ls.tridig_matrix[k, k - 1] = β
+        ls.tridig_matrix[k - 1, k] = real(β)
+        ls.tridig_matrix[k, k - 1] = real(β)
```
kellertuer:
```suggestion
        ls.tridig_matrix[k - 1, k] = β
        ls.tridig_matrix[k, k - 1] = β
```

cf. line 162; β is a norm. a norm is always real (see https://github.com/JuliaManifolds/ManifoldsBase.jl/blob/3900fcb115272d0a6781db529d1d54fe6a406936/src/ManifoldsBase.jl#L990 – though yeah the docs do not fully say that ;))

amitjamadagni: Yes, I guess I must have more closely read that code block as β is being compared right at the start! I was reading that the entries of the tridiagonal matrix must be real and hence put everything to real. Indeed here it is unncessary!

kellertuer: No worries, that is also what code review is for. I do not expect you (or anyone else) to know the whole code base (neither the detail on inner nor the one on norm) :)

kellertuer (approved):

kellertuer:
Thanks again. On my current working branch I turned your citation into an actual one, see

https://github.com/JuliaManifolds/Manopt.jl/pull/643/changes/91673e71a7c8ee13372672cba4b87a87161f0338

but kept the rest. When one overs over that in the docs one still sees the title :)

